PR #6, opened 2018-11-27
[Add] Load option for user config

alusor:


isaachinman:
Hi @alusor, thanks for the PR.

First - can you please remove the `package-lock.json` files? This project uses `yarn`.

Second - this is a good idea, but in general we need to load all namespaces for SSR to work correctly. Did you have a specific use case? Can you explain your thoughts? Perhaps we want to load all namespaces on the server, but pass this user-defined `load` option to the clientside instance only?

alusor:
Sure i will remove it.
Second - Yes i think we need to load all namespaces, the thing is that in your source, the load have 'languageOnly' value, in my case i have 'es' and 'es-CO' so i need to set the value to 'all' to correctly render 'es-CO'

isaachinman:
Oh sorry, I misunderstood. This is the `load` option from `i18next` directly that essentially enables or disables sub-language locales. The two main options we'd want to allow here would be `languageOnly` and `all`.

Once this PR is cleaned up a bit I'll be happy to accept it.

src/config/create-config.js
```diff
@@ -10,19 +10,23 @@ export default (userConfig) => {
 
   combinedConfig.allLanguages = combinedConfig.otherLanguages
     .concat([combinedConfig.defaultLanguage])
+  combinedConfig.load = combinedConfig.load || 'languageOnly'
```
isaachinman: Default values should go inside `default-config.js`, not via the pipe operator like this.

src/config/create-config.js
```diff
@@ -10,19 +10,23 @@ export default (userConfig) => {
...
+
 
   if (isNode && !process.browser) {
     const fs = eval("require('fs')") // eslint-disable-line
     const path = require('path')
 
     const getAllNamespaces = p => fs.readdirSync(p).map(file => file.replace('.json', ''))
     const {
-      allLanguages, defaultLanguage, localePath, localeStructure,
+      allLanguages, defaultLanguage, localePath, localeStructure, load, fallbackLng
```
isaachinman: Please keep this alphabetical.

isaachinman:
Ah, and this new option will need to be added to the README.

isaachinman:
@alusor Do you know if you'll have time to tidy this up? Otherwise no worries, I will take care of it.

alusor:
done, i was do the necessary changes

kachkaev:
There is still package-lock.json in the example folder :)

isaachinman:
This PR is actually unnecessary. @alusor Thank you for your time regardless.

You can pass `load` into the class constructor:

```jsx
new NextI18Next({ load: 'all' })
```

PR #8, opened 2018-11-29
Feature: Custom language detectors

lucasfeliciano:
This is my proposal to support custom language detectors.

Would be nice if the interface for custom detectors wasn't different from each other

Currently we have

Server's interface:
```js
 lookup(req, res, options)
```

Client's interface:
```js
lookup(options)
```

Which make it unpractical to write isomorphic detectors, changing to the interface below would make things much simpler since we wouldn't need to write two detectors
```js
lookup(options, req, res)
```

### DIFF
--- a/src/create-i18next-client.js
+++ b/src/create-i18next-client.js
@@ -1,21 +1,32 @@
 import isNode from 'detect-node'
 import i18next from 'i18next'
 import i18nextXHRBackend from 'i18next-xhr-backend'
-import i18nextBrowserLanguageDetector from 'i18next-browser-languagedetector'
+import I18nextBrowserLanguageDetector from 'i18next-browser-languagedetector'
 
 const i18n = i18next.default ? i18next.default : i18next
 i18n.nsFromReactTree = []
 
-if (isNode) {
-  const i18nextNodeBackend = eval("require('i18next-node-fs-backend')") // eslint-disable-line
-  const i18nextMiddleware = require('i18next-express-middleware')
-  i18n.use(i18nextNodeBackend).use(i18nextMiddleware.LanguageDetector)
-} else {
-  i18n.use(i18nextXHRBackend).use(i18nextBrowserLanguageDetector)
-}
-
 export default (config) => {
   if (!i18n.isInitialized) {
+    if (isNode) {
+      const i18nextNodeBackend = eval("require('i18next-node-fs-backend')") // eslint-disable-line
+      const I18nextMiddleware = require('i18next-express-middleware')
+
+      const serverLanguageDetector = new I18nextMiddleware.LanguageDetector()
+
+      config.detection.serverDetectors.forEach((detector) => {
+        serverLanguageDetector.addDetector(detector)
+      })
+      i18n.use(i18nextNodeBackend).use(serverLanguageDetector)
+    } else {
+      const browserLanguageDetector = new I18nextBrowserLanguageDetector()
+
+      config.detection.browserDetectors.forEach((detector) => {
+        browserLanguageDetector.addDetector(detector)
+      })
+      i18n.use(i18nextXHRBackend).use(browserLanguageDetector)
+    }
+
     i18n.init(config)
   }
   return i18n
--- a/src/config/default-config.js
+++ b/src/config/default-config.js
@@ -21,6 +21,8 @@ export default {
     format: (value, format) => (format === 'uppercase' ? value.toUpperCase() : value),
   },
   detection: {
+    serverDetectors: [],
+    browserDetectors: [],
     order: ['cookie', 'header', 'querystring'],
     caches: ['cookie'],
   },


src/create-i18next-client.js
```diff
@@ -1,21 +1,32 @@
...
-  i18n.use(i18nextXHRBackend).use(i18nextBrowserLanguageDetector)
-}
-
 export default (config) => {
   if (!i18n.isInitialized) {
+    if (isNode) {
+      const i18nextNodeBackend = eval("require('i18next-node-fs-backend')") // eslint-disable-line
+      const I18nextMiddleware = require('i18next-express-middleware')
+
+      const serverLanguageDetector = new I18nextMiddleware.LanguageDetector()
```
isaachinman: What are the consequences of initialising a detector in (default) case where a user has not provided any custom detectors?

src/create-i18next-client.js
```diff
@@ -1,21 +1,32 @@
...
+
+      config.detection.serverDetectors.forEach((detector) => {
+        serverLanguageDetector.addDetector(detector)
+      })
+      i18n.use(i18nextNodeBackend).use(serverLanguageDetector)
+    } else {
+      const browserLanguageDetector = new I18nextBrowserLanguageDetector()
```
isaachinman: Same question here.

src/create-i18next-client.js
lucasfeliciano: It will behave like behave now, using the constructor gives you methods to add new custom detectors but the default ones are still initialized.

src/create-i18next-client.js
lucasfeliciano: 👆

lucasfeliciano:
> In general this approach seems fine to me. However, do you mind explaining your specific use case?

My use case is that I need to detect language based on `domain`.

`.com -> en`,  `.es -> es` `fr.domain.ch -> fr-CH`.

In order to do that I need to create a custom detector:

[Frontend](https://github.com/i18next/i18next-browser-languageDetector#adding-own-detection-functionality)

[Backend](https://github.com/i18next/i18next-express-middleware#adding-own-detection-functionality)

>Also, forgive my ignorance as I've never used these custom detectors - why do you need multiple/an array here? How is a single (req, res, options) function not sufficient?

Because if you want to you can add multiple custom detectors and order how they will be applied with the `order` [option](https://github.com/i18next/i18next-express-middleware#detector-options)

example: `  order: ['myCustomDetector1', 'querystring', 'cookie', 'myCustomDetector2', 'header'],`

> Changing language detection is going to run into major problems with the localeSubpaths functionality. I'm not sure how we'd reconcile that, as the /${lng}/ formatting is by nature opinionated. This is my main concern.

It won't be a problem since we are just adding new detections and they are optional because you need to pass the order option. The other detections modes are not touched.

> Also, any changes to API/options need to be accompanied by documentation.

Yes, since I'm introducing two new options:     `serverDetectors: [], browserDetectors: []`

I'm just giving an option to nexti18next initialize `i18n` with custom detectors which can be done if you don't use `nexti18next`
 
Maybe @jamuhl have nice insights for better API since this is more related to i18next behaviour than `nextjs`

isaachinman:
What I'm saying is that if a user has provided custom detectors, we most likely need to disable all `localeSubpaths` functionality, and export a warning instead of the `Link` component, etc.

If your solution involves a custom pattern, and not the `/${lng}/` pattern, you will have to handle navigation and redirects manually.

Always happy to hear @jamuhl's opinion.

jamuhl:
@isaachinman would be awesome if we could have both i guess -> working with subpaths or without. An option could be just return the next.js Link component if subpath functionality is turned off.

But would also understand if subpath is the default and only way to set language (additional options are always more complexity)

isaachinman:
It simply wouldn't work. Let's take @lucasfeliciano's use case:

If `localeSubpaths` was enabled, and they passed their custom detector, the `Link` component would end up redirecting to paths like:

    de.myapp.com/de/

By going down the "custom detection" path, users are committing to handling their own navigation. I would like to make this abundantly clear by not exporting a `Link` component at all, as the entire point of that component is to support `localeSubpaths`.

If that sounds acceptable, we can proceed with:

- [ ] Forcibly disable `localeSubpaths` if `clientDetectors` or `ssrDetectors` is non-empty.
- [ ] Export a warning instead of `Link` component if `clientDetectors` or `ssrDetectors` is non-empty.
- [ ] Add new options to documentation.

As a side note, where did the `ssrDetectors` naming come from? I'd rather it be `clientDetectors` and `serverDetectors`.

isaachinman:
@lucasfeliciano Let me know if you want to make these changes, otherwise I'll get it done.

lucasfeliciano:
> By going down the "custom detection" path, users are committing to handling their own navigation. I would like to make this abundantly clear by not exporting a Link component at all, as the entire point of that component is to support localeSubpaths.

So we also need take in consideration the `subdomain` [detector](https://github.com/i18next/i18next-browser-languageDetector/blob/master/src/browserLookups/subdomain.js) on [i18next-browser-languagedetector](https://github.com/i18next/i18next-browser-languageDetector/blob/master/src/index.js#L48), which is not custom.

I can do the changes during the weekend today I'm probably busy all day with ongoing work here at @TicketSwap

isaachinman:
Let's handle one thing at a time.

The subdomain support is a good point, and I've created #16 to track the progress of that.

I'll wait for the other changes to get these custom detectors merged. Thanks!

isaachinman:
Yes, I believe this issue would be made redundant by the potential plugin fix being discussed in #14.

If @jamuhl believes that fix is possible, I will close #14 and this PR, and open a new issue to track the progress within the actual i18next repo.

jamuhl:
for me ok...

lucasfeliciano:
@jamuhl Do you want to make a battle plan for this?
I can implement if we have a solution that makes sense for you.
What I don't want to do is try to push your project or this one to a direction that doesn't match the vision of the library.

:)

jamuhl:
@lucasfeliciano will see if i can add support for calling `use` after init after publishing v13 - but yet not sure if we can support this...as it might have to big unwanted side effects:  https://github.com/isaachinman/next-i18next/issues/14#issuecomment-443737496

isaachinman:
Should be fixed with #28. Please let me know if I'm missing something.

PR #12, opened 2018-11-30
bugfix: remove deconstruction for pageProps

untsop:
`getInitialProps` in Page will stop working if you deconstruction `pageProps`.


isaachinman:
Good catch, thanks!

PR #24, opened 2018-12-04
Feature: Enable default language subpath

anandaprabawa:
This is my proposal to enable default language subpath.

Currently the default language just showing the root path `/`, and we might want to show the language `/en/` in the address bar

To enable this feature by passing an option into the NextI18Next constructor:
```jsx
new NextI18Next({ localeSubpaths: true, defaultLocaleSubpath: true })
```

If `localeSubpaths` is false, the defaultLocaleSubpath won't work.

### DIFF
--- a/src/utils/lng-path-detector.js
+++ b/src/utils/lng-path-detector.js
@@ -22,8 +22,8 @@ export default (req, res, next) => {
       req.i18n.changeLanguage(defaultLanguage)
     }
     /*
-      If a user has a default language prefix
-      in their URL, strip it. Except it has defaultLocaleSubpath to true
+      If a user has a default language prefix in their URL
+      and defaultLocaleSubpath set to false, strip it.
     */
     if (
       language === defaultLanguage
--- a/src/utils/lng-path-detector.js
+++ b/src/utils/lng-path-detector.js
@@ -1,7 +1,7 @@
 export default (req, res, next) => {
   if (req.i18n) {
     const language = req.i18n.languages[0]
-    const { allLanguages, defaultLanguage } = req.i18n.options
+    const { allLanguages, defaultLanguage, defaultLocaleSubpath } = req.i18n.options
     /*
       If a user has hit a subpath which does not
       match their language, give preference to
@@ -23,9 +23,13 @@ export default (req, res, next) => {
     }
     /*
       If a user has a default language prefix
-      in their URL, strip it.
+      in their URL, strip it. Except it has defaultLocaleSubpath to true
     */
-    if (language === defaultLanguage && req.url.startsWith(`/${defaultLanguage}/`)) {
+    if (
+      language === defaultLanguage
+      && req.url.startsWith(`/${defaultLanguage}/`)
+      && !defaultLocaleSubpath
+    ) {
       res.redirect(301, req.url.replace(`/${defaultLanguage}/`, '/'))
     }
   }
--- a/src/utils/lng-path-detector.js
+++ b/src/utils/lng-path-detector.js
@@ -1,7 +1,9 @@
 export default (req, res, next) => {
   if (req.i18n) {
     const language = req.i18n.languages[0]
-    const { allLanguages, defaultLanguage, defaultLocaleSubpath } = req.i18n.options
+    const {
+      allLanguages, defaultLanguage, localeSubpaths, defaultLocaleSubpath,
+    } = req.i18n.options
     /*
       If a user has hit a subpath which does not
       match their language, give preference to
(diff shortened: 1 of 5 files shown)


isaachinman:
It's a nice proposal, but do you feel like contributing a PR?

anandaprabawa:
> It's a nice proposal, but do you feel like contributing a PR?

Yes, of course

src/utils/lng-path-detector.js
```diff
@@ -23,9 +23,13 @@ export default (req, res, next) => {
     }
     /*
       If a user has a default language prefix
-      in their URL, strip it.
+      in their URL, strip it. Except it has defaultLocaleSubpath to true
```
isaachinman: Let's instead write:

> If a user has a default language prefix in their URL and `defaultLocaleSubpath` set to false, strip it.

anandaprabawa: I've updated this and the readme.

isaachinman:
Ah, please ignore that. I was mistaken. Another consideration, though: in the new options paradigm, `localeSubpaths` and `defaultLocaleSubpath` are interdependent. That is to say, the `defaultLocaleSubpath` option requires that the `localeSubpaths` option also be true.

Because of that, I am strongly considering changing the option from a bool to a string, where `localeSubpaths` can either be `non-defaults` or `all`. This would of course be a breaking change, so perhaps it's best to line that up with a v1.0 release.

Thanks again.

isaachinman:
@anandaprabawa If, given this new option, I hit `localhost:3000` with a default language of `en`, the server actually returns a 200 with our Next doc. The redirect only takes place on the client. This is not a production-ready solution - we would need to return a 301 to `localhost:3000/en/`.

There are already 301s in place in some of the source code, but I had not considered this use case. I think we're going to need some new logic.

isaachinman:
Any update here?

anandaprabawa:
> Any update here?

Not yet, can we close this PR?

isaachinman:
Why do you want to close it? We're 90% of the way there.

anandaprabawa:
@isaachinman i've updated the changes to return 301 when hit `localhost:3000`, does it work as you expected?

src/utils/lng-path-detector.js
```diff
@@ -21,11 +23,23 @@ export default (req, res, next) => {
...
       req.i18n.changeLanguage(defaultLanguage)
     }
+
     /*
-      If a user has a default language prefix
-      in their URL, strip it.
+      If a user enable the localeSubpaths and defaultLocaleSubpath,
+      request to the root path will redirect to default language
+     */
+    if (localeSubpaths && defaultLocaleSubpath && req.url === '/') {
```
isaachinman: You need to check here if the current language is the default language, no?

isaachinman:
This PR will conflict with #48 - one of you will have to do some conflict resolution.

src/utils/lng-path-detector.js
anandaprabawa: Isn't that the default language always the root path `/` ? why we need to check it?

isaachinman: Yes actually, this should be fine. However, you should not be using `req.url` as that will include query parameters.

isaachinman:
Anyone interested in picking this up? Otherwise it will be closed as it's quite old now.

capellini:
@isaachinman I can pick it up, but if possible, given how much the code base has changed, could we do so after #90 is merged?

isaachinman:
@capellini Sure.

lucasfeliciano:
I can also help it out.
Let me know if you easily can tackle this @capellini, otherwise I'm up to jump into this too.

capellini:
@lucasfeliciano Acutally, if you don't mind taking over this one, I'd like to focus on getting unit tests to 100% coverage and getting more e2e tests up and running.

lucasfeliciano:
You can also assign me to this one @isaachinman.
I would appreciate if you assign me to this issue; otherwise, I lose focus. 🙈

isaachinman:
@lucasfeliciano This one I will gladly assign to you, but I think we're better off closing this PR and opening an issue to track the progress, as this work is very old at this point.

lucasfeliciano:
Alright!

Let me know about the priority on this issue because I'm going on a city trip on Thursday and I'm on Sunday.
I can tackle this when I come back if it is not a problem

isaachinman:
@lucasfeliciano Priority on this is fairly low to be honest. A couple people have requested this, but have disappeared or not responded when I tried to continue the discussion. There are probably other areas where your help would be more useful.

dayze:
We will be very interested in this feature, I can try to look at it this week.

chemuto:
Hi @dayze . Did you have a chance to look into this? I would also be interested in using this feature. Cheers

dayze:
Hey @chemuto, I'll start on it monday.

PR #30, opened 2018-12-05
Update README.md

igorbuts:
Mention about updating scripts after adding `server.js` file.


isaachinman:
I don't think I'd like to include anything as opinionated as this. The instructions are mere guidelines, not strict rules. You can call it `not-a-server.js` if you want.

If anything, we can include a simple link to [this section](https://github.com/zeit/next.js#custom-server-and-routing) of the NextJs docs.

isaachinman:
Also just to note, based on the way you structured your link, any changes to `package.json` would result in a miss-link.

isaachinman:
78961ac

PR #43, opened 2018-12-12
pass additional parameters of getInitialProps all the way down

fgreinus:
Hey guys,

when using this package I stumbled upon the problem, that the next-apollo integration provides the apollo client object as second parameter to the getInitialProps call. But as the HOC for the NextApp does not pass these values along that information is lost.

Use case is, that I need the apollo client in the getInitialProps of my app to fetch the authentication state of the user.

Maybe this is also helpful for others around here.

If this is anti-pattern or something you shouldn't do - please explain me why and what are possible options to work around that. I'm still pretty new to the whole nextjs ecosystem.

Thanks!

### DIFF
--- a/src/hocs/app-with-translation.js
+++ b/src/hocs/app-with-translation.js
@@ -28,19 +28,19 @@ export default function (WrappedComponent) {
       }
     }
 
-    static async getInitialProps({ Component, ctx }) {
+    static async getInitialProps({ Component, ctx }, ...additionalContext) {
 
       let pageProps = {}
       let regularProps = {}
 
       if (Component.getInitialProps) {
-        pageProps = await Component.getInitialProps(ctx)
+        pageProps = await Component.getInitialProps(ctx, ...additionalContext)
       }
 
       // Run getInitialProps on wrapped _app
       if (WrappedComponent.getInitialProps) {
         const { pageProps: wrappedPageProps, ...rest } = await WrappedComponent
-          .getInitialProps({ Component, ctx })
+          .getInitialProps({ Component, ctx }, ...additionalContext)
         pageProps = {
           ...pageProps,
           ...wrappedPageProps,


isaachinman:
First of all, we are not talking about "parameters" or "arguments", but rather a single argument which is an object.

Second, I wasn't aware that people were using this package with other `next-x` packages, cool!

Finally, this approach seems totally sensible. That said, I'd rather rename `additionalContext` to `rest`. If someone has a problem with that naming, we can call it `otherKeys` or something.

isaachinman:
Apologies, I misread the PR. This _is_ in regards to further arguments on `getInitialProps`. I have not seen this done before, and I wonder why `next-apollo` is doing this instead of passing it through the object.

I'll need to look into this a bit further.

isaachinman:
Okay, after looking around it seems like NextJs is [committed to handling all their internals via the first-arg object](https://github.com/zeit/next.js#fetching-data-and-component-lifecycle).

Using [rest params](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters) here should be fine, and certainly isn't an anti-pattern, but could break if the NextJs API changes.

isaachinman:
@fgreinus Did you test this approach and ensure it works with `next-apollo`?

fgreinus:
Thanks for your input! Actually I was able to get it working using the with-apollo-auth example from the next.js repo.
This is because in there the apollo client is manually set in the context object (ctx) and then passed along.
See: https://github.com/zeit/next.js/blob/canary/examples/with-apollo-auth/lib/withApollo.js

This methods has no need for any change on next-i18next and seems to be working fine for me now. Your questions/thoughs made me try that approach, so thank you!

I'll close this here then.

Have a nice day and thanks for you work 👍 .

kachkaev:
@fgreinus I'm also using `i18next` and `apollo-client` in a few of projects and I recently played with integrating them with `next-i18next`. The composition of HOCs seem to work and you can check-out my approach in this open-source code base here: https://gitlab.com/kachkaev/website-frontend.

isaachinman:
@fgreinus Ah yeah, very nice. It definitely makes more sense (to me) to pass things through the object. Good to hear that it worked.

fgreinus:
> @fgreinus I'm also using `i18next` and `apollo-client` in a few of projects and I recently played with integrating them with `next-i18next`. The composition of HOCs seem to work and you can check-out my approach in this open-source code base here: https://gitlab.com/kachkaev/website-frontend.

Thanks! That's pretty similar to what i'm currently doing, except that i'm passing the apollo client instance into the context object that's passed to other getInitialProps calls, because i need it in _app.jsx's getInitialProps method.

Issue #18, opened 2018-12-01
Add tests


isaachinman:
This project is in desperate need of a test suite to prevent basic regression.

It'd be great to have unit tests on logic like [lng-path-corrector](https://github.com/isaachinman/next-i18next/blob/master/src/utils/lng-path-corrector.js), component tests on the HOCs and `Link`, and some integration tests on the [example](https://github.com/isaachinman/next-i18next/tree/master/example) app.

This would be a great way for anyone wanting to get involved to start contributing.

PR #45, opened 2018-12-12
Allowing 'as' prop in Link and fixing 'href' prop

capellini:
Previously, if a user already passed in an 'as' prop to the i18n Link
component, it would overwrite the 'as' prop to the next/link Link component.
This checks to see if an 'as' prop has been passed in to the i18n link
component and, if so, modifies it, rather than overwriting it.

Also, if the 'href' prop happened to already have query parameters, the 'lng'
parameter would be incorrectly added (e.g. if 'href=/foo?bar', then it would
be modified to 'href=/foo?bar?lng=en').  This commit also checks to see if
query parameters already exist and if, so, appends the lng parameter properly.

Lastly, this commit adds some initial test infrastructure using enzyme and
jest with code coverage and adds tests for the Link component.


capellini:
> However, although the matter of testing is something we desperately need (I asked for help awhile ago in #18), I'd like if you could separate these two issues into two distinct PRs:

Yes, of course.  I should have thought to do that myself.  I'll take out the testing stuff from this PR and add a separate PR for that.

isaachinman:
Okay thanks, that's great. Can you just tag me here again when you're comfortable with this PR being reviewed?

capellini:
@isaachinman -- ready for review

src/components/Link.js
```diff
@@ -27,27 +27,31 @@ export default function () {
...
     }
   }
 
   Link.propTypes = {
+    as: PropTypes.string,
     children: PropTypes.node.isRequired,
     href: PropTypes.string.isRequired,
   }
 
+  Link.defaultProps = { as: null }
```
isaachinman: Please add line breaks and space this as a normal object.

src/components/Link.js
```diff
@@ -27,27 +27,31 @@ export default function () {
...
         return (
-          <NextLink href={`${href}?lng=${lng}`} as={`/${lng}${href}`}>
-            {children}
-          </NextLink>
+          <NextLink href={hrefWithLang} as={`/${lng}${as || href}`}>{children}</NextLink>
         )
       }
-      return <NextLink href={href}>{children}</NextLink>
+
+      return <NextLink href={href} as={as}>{children}</NextLink>
```
isaachinman: Is it safe to pass `null` into `as`?

src/components/Link.js
```diff
@@ -27,27 +27,31 @@ export default function () {
...
     render() {
       const { defaultLanguage, localeSubpaths } = config
-      const { children, href } = this.props
+      const { as, children, href } = this.props
       let lng = null
       if (Array.isArray(i18n.languages) && i18n.languages.length > 0) {
         [lng] = i18n.languages
       }
       if (localeSubpaths && lng && lng !== defaultLanguage) {
+        const hrefWithLang = href.includes('?') ? `${href}&lng=${lng}` : `${href}?lng=${lng}`
```
isaachinman: Not sure if this is the safest way to do query-checking. There would be no harm in pulling in a third-party lib to do query mutations - I would actually prefer that.

Regardless, can you please use `if/else` statements instead of a ternary? It's especially confusing in this case, due to the `'?'`.

src/components/Link.js
kachkaev: How about adding [Prettier](https://github.com/prettier/prettier) to the project to avoid such issues in future? I'm using it everywhere and it saves a lot of times, especially in big teams. I can help with setting that up if there is interest.

isaachinman: I'll add it in the future as we have more contributors. Not a huge priority right now as `eslint` will catch the important things.

capellini: I originally had it that way, but CodeClimate failed because the file would have been over 25 lines long.  Any chance that we can increase that limit on CodeClimate?

src/components/Link.js
capellini: I had left off the `defaultProps` entirely, which would have left the 'as' prop as `undefined`, but that triggered the `react/require-default-props` rule.  Perhaps I can disable that rule for this file, as I think it's valid to leave 'as' without a default prop.  What do you think?

src/components/Link.js
capellini: :+1: 

The `if/else` may also push us over our CodeClimate lines of code limit.  Let me see what I can do.

src/components/Link.js
isaachinman: We're not going to make decisions like this based on CodeClimate. Don't worry if such a thing causes it to "fail".

src/components/Link.js
isaachinman: It was a genuine question. If it's alright to pass `null` into `as`, I have no problems here. I suspect it is.

Otherwise, just put `undefined` as the default prop. I don't want to disable the `react/require-default-props` rule - it's important.

capellini:
@isaachinman - requested changes made...I "tested" the new `url-parse` functionality by using the node REPL.  Is there a better way to do this until I can submit the testing PR?

isaachinman:
@capellini Perhaps play around with the `example` codebase and try running `yarn run-example:prod` from the root.

I've never used the lib you brought in, only [query-string](https://www.npmjs.com/package/query-string).

I've updated the CodeClimate settings - do you mind doing a force push to trigger a build?

capellini:
> Perhaps play around with the `example` codebase and try running `yarn run-example:prod` from the root.

I played around a bit and it seems to work fine.

> I've never used the lib you brought in, only query-string.

LMK if you would prefer that I use query-string and I can change the code...no problem.

PR #49, opened 2018-12-13
Add context to react-tree-walker call

capellini:
By default react-tree-walker passes in an undefined context, which
breaks packages like Material UI and JSSProvider.  This fix passes
in an empty object for context, which fixes components that use the
legacy context API.

See

### DIFF
--- a/src/hocs/app-with-translation.js
+++ b/src/hocs/app-with-translation.js
@@ -82,7 +82,7 @@ export default function (WrappedComponent) {
           if (instance && instance.props && instance.props.ns) {
             nsFromTree = [...new Set(nsFromTree.concat(instance.props.ns))]
           }
-        })
+        }, {})
       }
 
       // Step 3: Perform data fetching, depending on environment


capellini:
Note that I'm still having trouble using the new React Context API, but it seems others are having the same problem, see ctrlplusb/react-tree-walker#40.  It looks like the fix was merged 12 days ago, but there as yet to be a new release cut.  This change just fixes Components that use the legacy context API.

isaachinman:
@capellini Thanks for noticing this. I did indeed play around with this previously, as it's related to a dev-only bug which I've hot-fixed with dfa93a8.

Any chance you mind taking a look at that, and seeing if you can spot the problem? I'm 90% sure it's context-related.

capellini:
> Any chance you mind taking a look at that, and seeing if you can spot the problem? I'm 90% sure it's context-related.

I took a look at dfa93a8 but it didn't fix the Material UI context problem that I was having.  I'm still experiencing problems with areas where I'm using the new Context API, but that's a `react-tree-walker` known issue, which has a merged PR, but a new release hasn't been cut yet.

But I can say that the problem is most definitely context-related.

PR #50, opened 2018-12-13
Fix url.href in url-parse usage

capellini:
In the current usage, if a URL without a protocol and host is passed
into URL, 'http://localhost' will be added to the beginning of the
href (e.g. '/foo' will become 'http://localhost/foo').  This change
ensures that a URL without a protocol and host is preserved and
left without the protocol and host (e.g. '/foo' remains '/foo').

### DIFF
--- a/__tests__/components/Link.test.js
+++ b/__tests__/components/Link.test.js
@@ -114,14 +114,14 @@ describe('Link component', () => {
     // without 'as' prop -- no query parameters
     let component = createLinkComponent()
 
-    expect(component.prop('href')).toEqual('http://localhost/foo/bar?lng=de')
+    expect(component.prop('href')).toEqual('/foo/bar?lng=de')
     expect(component.prop('as')).toEqual('/de/foo/bar')
 
     // without 'as' prop -- query parameters
     props.href = '/foo/bar?baz'
     component = createLinkComponent()
 
-    expect(component.prop('href')).toEqual('http://localhost/foo/bar?baz=&lng=de')
+    expect(component.prop('href')).toEqual('/foo/bar?baz&lng=de')
     expect(component.prop('as')).toEqual('/de/foo/bar?baz')
 
     props.href = '/foo/bar'
@@ -130,7 +130,7 @@ describe('Link component', () => {
     props.as = '/foo?bar'
     component = createLinkComponent()
 
-    expect(component.prop('href')).toEqual('http://localhost/foo/bar?lng=de')
+    expect(component.prop('href')).toEqual('/foo/bar?lng=de')
     expect(component.prop('as')).toEqual('/de/foo?bar')
   })
 })
--- a/package.json
+++ b/package.json
@@ -69,6 +69,6 @@
     "prop-types": "^15.6.2",
     "react": "^16.6.3",
     "react-i18next": "^8.4.0",
-    "url-parse": "^1.4.4"
+    "urijs": "^1.19.1"
   }
 }
--- a/src/components/Link.js
+++ b/src/components/Link.js
@@ -19,7 +19,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import NextLink from 'next/link'
-import Url from 'url-parse'
+import URI from 'urijs'
 
 export default function () {
 
@@ -34,11 +34,15 @@ export default function () {
         [lng] = i18n.languages
       }
       if (localeSubpaths && lng && lng !== defaultLanguage) {
-        const url = new Url(href, true)
-        url.set('query', { ...url.query, lng })
+        const hrefWithLang = URI(href)
+          .addQuery('lng', lng)
+          .normalizeQuery()
+          .toString()
 
         return (
-          <NextLink href={url.href} as={`/${lng}${as || href}`}>{children}</NextLink>
+          <NextLink href={hrefWithLang} as={`/${lng}${as || href}`}>
+            {children}
+          </NextLink>
         )
       }
 
--- a/yarn.lock
+++ b/yarn.lock
@@ -6308,11 +6308,6 @@ querystring@0.2.0, querystring@^0.2.0:
   resolved "https://registry.yarnpkg.com/querystring/-/querystring-0.2.0.tgz#b209849203bb25df820da756e747005878521620"
   integrity sha1-sgmEkgO7Jd+CDadW50cAWHhSFiA=
 
-querystringify@^2.0.0:
-  version "2.1.0"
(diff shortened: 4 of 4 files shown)


isaachinman:
Hm, it might be time to use a different package for this. We shouldn't have to manually check if a query exists.

Can you look at [qs](https://www.npmjs.com/package/qs) or [query-string](https://www.npmjs.com/package/query-string)?

capellini:
> Can you look at [qs](https://www.npmjs.com/package/qs) or [query-string](https://www.npmjs.com/package/query-string)?

:+1:  Absolutely.  I'll take a look and get back to you.

isaachinman:
Thanks. I think they're less focused on mutating query strings, but should work fine with some parsing/stringifying.

isaachinman:
@capellini It looks like you're using _both_ `url-parse` and `qs` now? That's not what I had in mind - let's just use a single package.

capellini:
> It looks like you're using _both_ `url-parse` and `qs` now? That's not what I had in mind - let's just use a single package.

I'm still using `url-parse` so that we can extract the query string from the `href`. Otherwise, we would need a different method, since `qs.parse` (and the `query-string` package) don't know how to do this:

```js
qs.parse('/foo.com/bar?baz=bing', {})  // == { '/foo.com/bar?baz': 'bing' }
```

We don't want to just `href.split('?')` right?  Is there a better way/another package that can do both for us?  I can research this, but was wondering if you knew of something.

isaachinman:
@capellini I'm fine with proprietary code solving the problem if it seems robust and ends up being a lot smaller than pulling in a third party lib. The important part is that it probably needs to be vanilla JS as it's universal and can't rely on browser APIs.

There are [a lot](https://stackoverflow.com/questions/901115/how-can-i-get-query-string-values-in-javascript) of different approaches.

There should be a relatively safe way of splitting the query string from `href` manually, and then using `qs` for the rest. I'll leave it up to you unless you would like some help.

capellini:
@isaachinman - `url-parse` dependency removed.  I took the simplest route in breaking up the URL into path, query, and hash components.  Let me know what you think.

isaachinman:
There's no need to be doing slicing etc manually - we have `URL` from `url` in Node which is (I think) a 1:1 for the browser API. Check out [this example](https://codesandbox.io/s/p94x13q510). I'd rather use built-ins - that's why they're there. Let's not add a new `util` file until it's used in at least 2-3 places.

Sorry a simple thing has turned into such a back and forth, I appreciate your patience. If you want me to take over, just say so.

isaachinman:
Ah also, we've dropped `react-tree-walker`, so that will need to get resolved here.

capellini:
> I'd rather use built-ins - that's why they're there. 

Sorry about that. I thought that we were trying to avoid browser APIs.

> Let's not add a new `util` file until it's used in at least 2-3 places.

Agreed.  My thought was that we could use this util function to also fix [the bug mentioned here](https://github.com/isaachinman/next-i18next/pull/48#discussion_r242017370).  But I might be getting ahead of myself.

> Sorry a simple thing has turned into such a back and forth, I appreciate your patience. If you want me to take over, just say so.

No worries at all.  Each developer has their own style and this is your repository, so I have no problems changing things until you're comfortable.

> Ah also, we've dropped `react-tree-walker`, so that will need to get resolved here.

Ok.  Will do.

capellini:
@isaachinman -- it appears that `new URL(...)` requires that the URL be an absolute URL, but I think that we should allow for relative URLs, as well.

We can pass in a base URL as the second argument, but I'm not sure where we can pull that from.  Do you know if next.js stores this somewhere?

Another option is that we can just pass in a placeholder if the URL is a relative URL and remove it later, like so:

```js
href = origHref
if (isRelativeUrl(origHref)) {
  href = `http://example.com${origHref}`
}
const url = new URL(href)
...
let newHref = url.href
if (isRelativeUrl(origHref)) {
  newHref = newHref.replace('http://example.com', '')
}
...
```

Any other ideas?

isaachinman:
Wow, who knew that URL manipulation was going to be such a sticky subject in JavaScript of all languages.

I'm very busy at the moment, but I will take some time to dig into this (and related npm packages) sometime this week.

capellini:
> Wow, who knew that URL manipulation was going to be such a sticky subject in JavaScript of all languages.

Yes, I was a bit surprised myself.

> I'm very busy at the moment, but I will take some time to dig into this (and related npm packages) sometime this week.

Great.  Thanks.

isaachinman:
@capellini I just found [urijs](https://www.npmjs.com/package/urijs). Looks great. What do you think? Works with both absolute and relative.

capellini:
> @capellini I just found [urijs](https://www.npmjs.com/package/urijs). Looks great. What do you think? Works with both absolute and relative.

@isaachinman It seems to be popular at ~800k weekly downloads, though I am a little concerned at the last release 10 months ago, the 41 open issues, that the activity on their issue tracker doesn't seem to have many recently-closed issues.  But I think for our purposes that this should work fine, as we have a relatively-simple use case.  I'll work on incorporating it now.

capellini:
@isaachinman We should be good to merge this, but I suggest merging the test PR first.  Then I can rebase this and modify the test for Link, as appropriate, and be more confident that it's working as expected.

isaachinman:
Wow, that's a lot cleaner of a solution!

Feel free to rebase when you have a chance and we can get this merged as well.

capellini:
> Feel free to rebase when you have a chance and we can get this merged as well.

@isaachinman I think we're all set to merge on this one.

kachkaev:
It might be worth considering something else as a replacement for urijs as it has a rather significant impact on the bindle size. For me it's 42KB parsed / 13 KB gzipped. For comparison, next-i18next with its i18next dependencies is only 16KB gzipped:

screenshot 2018-12-23 at 17 04 15


WDYT of [`url`](https://www.npmjs.com/package/url) package folks? It has both `parseI()` and `format()` methods and is 2.76KB gzipped.

isaachinman:
PRs always welcome. If you think you can reduce the bundle size with a 1:1 solution that is sensible and maintainable, it will be merged.

isaachinman:
@kachkaev I'm not that familiar with the built-ins in this domain, as is obvious by the conversation above. However, something like [this](https://codesandbox.io/s/p94x13q510) should work in both environments, although you'd have to conditionally define `URL` and `URLSearchParams`.

I'd prefer either:

1. A utility that makes this incredibly easy to do, easy to reason about, zero dev effort (our current implementation).
2. Achieved fully with built-ins in a non-crazy-making way.

PR #66, opened 2018-12-21
Properly set namespace on client

### LINKED ISSUE #64: Properly set namespace on client
If a user sets their own `defaultNS` in their custom configuration, these changes are not properly propagated to `ns` if we are client-side.  See https://github.com/isaachinman/next-i18next/pull/48#pullrequestreview-185426858.

Pull request forthcoming.

capellini:
If a user specifies another defaultNS and we are client-side, then
config.ns will never be reassigned. This change properly sets the
ns key on the config object.

Resolves #64.

### DIFF
--- a/__tests__/config/create-config-non-node.test.js
+++ b/__tests__/config/create-config-non-node.test.js
@@ -1,6 +1,6 @@
 /* eslint-env jest */
 
-import { setUpTest, tearDownTest } from './test-helpers'
+import { userConfig, setUpTest, tearDownTest } from './test-helpers'
 
 import createConfig from '../../src/config/create-config'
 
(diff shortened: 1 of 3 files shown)


isaachinman:
This is great, thanks. In the future we might want to actually set `config.ns` to the zero index namespace defined via `requiredNamespaces`, _else_ `defaultNS`. Currently, `config.ns` is always the `defaultNS` (and this PR bug fixes a use case for that), but that means that the default namespace [always get sent, no matter the request](https://github.com/isaachinman/next-i18next/blob/master/src/hocs/app-with-translation.js#L81). This is inherently an opinionated matter, and could change if people don't like it.

Issue #70, opened 2018-12-27
CURL fails unless a fallbackLng is specified


anthonywebb:
If you run this example and try to curl or make a postman request to the server (any route) it will return an error 

```
TypeError: Cannot read property '0' of null
    at i18nextMiddleware (/Users/foo/projects/bar/node_modules/i18next-express-middleware/lib/index.js:73:59)
```

This is because no fallbackLng is specified when you init.

https://github.com/isaachinman/next-i18next/blob/32f9bdeb93c8e86da178ae13fee2b92c7f58a942/example/i18n.js#L3

isaachinman:
Hi @anthonywebb. The `fallbackLng` is specifically [not set](https://github.com/isaachinman/next-i18next/blob/master/src/config/default-config.js#L11) in development. This is because you can get false positives if you're testing in the language that is already your fallback language - it's better to see things broken in dev than have your server silently falling back and appearing to function correctly.

Why are you performing non-browser requests in dev mode? If you need to do this, why not just specify your `fallbackLng`? I'm not sure this is really a bug, or if anything needs to be done.

anthonywebb:
Fair enough, my server also acts as an API proxy for JSON requests to get data from our api server.  So I do that proxy testing in postman or curl, not sure I like hitting a url on my server and barfing a 500 for a route that loads fine in a browser. But the project is yours so feel free to accept/reject the PR.

Thanks again for a great project.

isaachinman:
I think in this case I'd rather users see translation failure than get a false positive, as I mentioned. I do believe your use case is relatively rare, and is of course fixable by adding your `fallbackLng` in. Worth noting this error only occurs in dev - your prod environment should be fine.

I think I will leave this out of the example for now - rather keep things as simple as possible instead of creating confusion to fix an edge case.

If more people run into this and want things changed, I will be happy to.

PR #71, opened 2018-12-27
Resolve issue when hitting routes via CURL/POSTMAN

anthonywebb:


isaachinman:
Closing for now due to reasons I enumerated on in #70. If more people run into this problem and would like the example config changed, we can definitely do that.

PR #75, opened 2019-01-02
Add createConsoleMessage functionality + logging configuration

MathiasKandelborg:
This enables users to set a strictMode boolean in their i18n config, that defines whether messages and errors are logged to the console.

So far I've not added an info message regarding traceLimit. I've added the `traceLimit` option because some errors would need a stack trace and that it's used in the function. Generally, the stack trace from pretty-error only contains standard modules and no helpful information, therefore it's quite unlikely that anyone would gain from being able to set the env variable. Nevertheless, I think, it does make sense to have it available.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "@babel/cli": "^7.1.5",
     "@babel/core": "^7.1.6",
     "@babel/node": "^7.0.0",
-    "@babel/plugin-transform-runtime": "^7.2.0",
+    "@babel/plugin-transform-runtime": "^7.1.0",
     "@babel/preset-env": "^7.1.6",
     "@babel/preset-react": "^7.0.0",
     "all-contributors-cli": "^5.4.1",
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "prepare": "yarn build",
     "run-example": "yarn build && cd example && rm -rf node_modules && yarn && yarn dev",
     "run-example:prod": "yarn build && cd example && rm -rf node_modules && yarn && yarn build && yarn start",
-    "test": "NODE_ENV=test jest",
+    "test": "NODE_ENV=test jest --silent",
     "contributors:check": "all-contributors check",
     "contributors:add": "all-contributors add",
     "contributors:generate": "all-contributors generate"
@@ -41,7 +41,7 @@
     "@babel/cli": "^7.1.5",
     "@babel/core": "^7.1.6",
     "@babel/node": "^7.0.0",
-    "@babel/plugin-transform-runtime": "^7.1.0",
+    "@babel/plugin-transform-runtime": "^7.2.0",
     "@babel/preset-env": "^7.1.6",
     "@babel/preset-react": "^7.0.0",
     "all-contributors-cli": "^5.4.1",
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
     "i18next-express-middleware": "^1.5.0",
     "i18next-node-fs-backend": "^2.1.0",
     "i18next-xhr-backend": "^1.5.1",
+    "pretty-error": "^2.1.1",
     "prop-types": "^15.6.2",
     "react": "^16.6.3",
     "react-i18next": "^8.4.0",
--- a/src/config/default-config.js
+++ b/src/config/default-config.js
@@ -5,8 +5,6 @@ const LOCALE_PATH = 'static/locales'
 const LOCALE_STRUCTURE = '{{lng}}/{{ns}}'
 const LOCALE_SUBPATHS = false
 
-const NEXT_I18NEXT_TRACE_LIMIT = parseInt(`${process.env.NEXT_i18NEXT_TRACE_LIMIT}`, 10) || 0
-
 export default {
   defaultLanguage: DEFAULT_LANGUAGE,
   otherLanguages: OTHER_LANGUAGES,
@@ -37,5 +35,5 @@ export default {
     wait: true,
   },
   strictMode: true,
-  traceLimit: NEXT_I18NEXT_TRACE_LIMIT,
+  errorStackTraceLimit: parseInt(`${process.env.NEXT_i18NEXT_TRACE_LIMIT}`, 10) || 0,
 }
--- a/src/config/default-config.js
+++ b/src/config/default-config.js
@@ -5,6 +5,8 @@ const LOCALE_PATH = 'static/locales'
 const LOCALE_STRUCTURE = '{{lng}}/{{ns}}'
 const LOCALE_SUBPATHS = false
 
+const NEXT_I18NEXT_TRACE_LIMIT = parseInt(`${process.env.NEXT_i18NEXT_TRACE_LIMIT}`, 10) || 0
+
 export default {
   defaultLanguage: DEFAULT_LANGUAGE,
   otherLanguages: OTHER_LANGUAGES,
@@ -34,4 +36,6 @@ export default {
   react: {
     wait: true,
   },
+  strictMode: true,
+  traceLimit: NEXT_I18NEXT_TRACE_LIMIT,
 }
--- a/src/config/default-config.js
+++ b/src/config/default-config.js
@@ -35,5 +35,5 @@ export default {
     wait: true,
   },
   strictMode: true,
-  errorStackTraceLimit: parseInt(`${process.env.NEXT_i18NEXT_TRACE_LIMIT}`, 10) || 0,
+  errorStackTraceLimit: 0,
 }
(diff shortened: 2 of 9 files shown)


src/config/default-config.js
```diff
@@ -5,6 +5,8 @@ const LOCALE_PATH = 'static/locales'
 const LOCALE_STRUCTURE = '{{lng}}/{{ns}}'
 const LOCALE_SUBPATHS = false
 
+const NEXT_I18NEXT_TRACE_LIMIT = parseInt(`${process.env.NEXT_i18NEXT_TRACE_LIMIT}`, 10) || 0
```
isaachinman: If this is just going to be merged into the config object, we might as well leave this out, as users can already pass overriding values via the user config.

src/config/default-config.js
```diff
@@ -34,4 +36,6 @@ export default {
   react: {
     wait: true,
   },
+  strictMode: true,
+  traceLimit: NEXT_I18NEXT_TRACE_LIMIT,
```
isaachinman: Is it possible to give a more descriptive name here? Which trace limit?

src/hocs/app-with-translation.js
```diff
@@ -6,6 +6,8 @@ import { lngPathCorrector } from 'utils'
 import { NextStaticProvider } from 'components'
 
 import hoistNonReactStatics from 'hoist-non-react-statics'
+import createConsoleMessage from '../utils/console-message'
```
isaachinman: Please use the import/export conventions laid out in other files. We're using the babel module resolver plugin, so you can import from `utils` as if it was an absolute path. Also, `utils/index.js` should export `console-message`.

And finally, everything should be in alphabetical order.

src/hocs/app-with-translation.js
```diff
@@ -71,7 +85,10 @@ export default function (WrappedComponent) {
       if (Array.isArray(pageProps.namespacesRequired)) {
         ({ namespacesRequired } = pageProps)
       } else if (process.env.NODE_ENV !== 'production') {
```
isaachinman: I had envisioned this logic being _within_ `console-message.js` so that we don't have to repeat it everywhere.

src/utils/console-message.js
```diff
@@ -0,0 +1,104 @@
...
+    traceLimit: 0,
+    strictMode: true,
+  }) {
+  const { traceLimit, strictMode } = options
+
+  if (!strictMode) {
+    return
+  }
+  const util = require('util')
+  const PrettyError = require('pretty-error')
```
isaachinman: Is this going to get pulled into prod bundles?

src/utils/console-message.js
```diff
@@ -0,0 +1,104 @@
...
+  /* Make room for new message */
+  console.log()
+  // Create a new empty error
+  const newLog = new Error()
+
+  // Make sure the message is a string
+  if (typeof message !== 'string') {
+    const metaError = new Error()
+    metaError.name = 'Meta'
+    metaError.message = `Param message needs to be of type: string. Instead, '${typeof message}' was provided.\n
```
isaachinman: This is fine, but it would be great if you can provide basic test coverage for `console-message.js`.

src/utils/console-message.js
```diff
@@ -0,0 +1,104 @@
...
+      setStackTraceLimit()
+      return
+    default:
+      newLog.name = messageTypes.info
+      console.info(pe.render(newLog))
+      setStackTraceLimit()
+  }
+}
+
+/* Debugging */
```
isaachinman: Please remove.

MathiasKandelborg:
I believe everything got fixed.

I'm quite sure imports won't happen if `process.env.NODE_ENV !== 'production'` now.

I've also written a test file that should be covering everything, I'm quite new with tests (still trying out in spare time) - so I apologize for any bad practices or poor structure.

package.json
```diff
@@ -27,7 +27,7 @@
     "prepare": "yarn build",
     "run-example": "yarn build && cd example && rm -rf node_modules && yarn && yarn dev",
     "run-example:prod": "yarn build && cd example && rm -rf node_modules && yarn && yarn build && yarn start",
-    "test": "NODE_ENV=test jest",
+    "test": "NODE_ENV=test jest --silent",
```
isaachinman: What's the `--silent` flag here?

package.json
```diff
@@ -41,7 +41,7 @@
     "@babel/cli": "^7.1.5",
     "@babel/core": "^7.1.6",
     "@babel/node": "^7.0.0",
-    "@babel/plugin-transform-runtime": "^7.1.0",
+    "@babel/plugin-transform-runtime": "^7.2.0",
```
isaachinman: Did you mean to bump the version?

src/config/default-config.js
```diff
@@ -34,4 +34,6 @@ export default {
   react: {
     wait: true,
   },
+  strictMode: true,
+  errorStackTraceLimit: parseInt(`${process.env.NEXT_i18NEXT_TRACE_LIMIT}`, 10) || 0,
```
isaachinman: I meant remove the `process.env` variable entirely. If the `errorStackTraceLimit` setting is within the config, users can just pass it into the class constructor. Unless you think it still makes sense to have an env var?

src/hocs/app-with-translation.js
```diff
@@ -24,8 +24,20 @@ export default function (WrappedComponent) {
               Router.replace(href, as, { shallow: true })
             }
           }
+
         })
       }
+
+      this.createConsoleMessage = (type, message, traceLimit = 0) => {
```
isaachinman: In general I prefer class properties unless something specifically requires `constructor` use. Is that possible here?

src/utils/console-message.js
```diff
@@ -0,0 +1,109 @@
+/* eslint-disable no-console */
+
+
+/**
+  * @readonly
+  * @enum {String} A console.log type
+  */
+const messageTypes = {
+  warn: 'Warning',
+  err: 'Error',
```
isaachinman: Sorry I didn't catch this the first time around. Can we use the same naming as `console` itself does? Thus, `warn, error, info`?

isaachinman:
Also, I do think the CodeClimate guidelines in this case are mostly useful. It would be good to eliminate most/all of the issues there. If you run into something that you can't fix or think doesn't make sense, do let me know.

package.json
MathiasKandelborg: It's described in this commit: https://github.com/isaachinman/next-i18next/pull/75/commits/116cd7c51169415f7792730386643e1b71be09cf

package.json
MathiasKandelborg: No, I'm not sure why I did it - I've kept the change out of my commits but forgot to un-stage the hunk with the --silent flag commit.

src/config/default-config.js
MathiasKandelborg: It's probably fine to remove it entirely. The use cases are few, there's none as far as I know. It _could_ be useful to the have the possibility of setting it with an env variable in the future.

MathiasKandelborg:
I'm soon hitting some personal deadlines and would be happy to see the main functionality get accepted and improve some dev experiences.

If you're willing to finish it up, that would be great. Thanks for the help and letting me contribute.

src/utils/console-message.js
MathiasKandelborg: Of course, I'll make the changes along with some CodeClimate changes.

isaachinman:
Okay great. Thanks again for your work on this. Just let me know when you need to step away from it and I will finish up the rest. Cheers!

MathiasKandelborg:
@isaachinman I just want to make sure you've seen the latest updates. :slightly_smiling_face:

isaachinman:
@MathiasKandelborg Yes I have. Sorry, it's been a busy week for me. I'm hoping to have time to review and merge this sometime this weekend.

isaachinman:
In regards to the CodeClimate duplication errors, we can easily fix this by doing something like:

```jsx
if (Object.values(messageTypes).includes(messageType)) {
  newLog.name = capitalize(messageTypes[messageType])
  console[messageType](pe.render(newLog))
} else {
  newLog.name = capitalize(messageTypes.info)
  console.info(pe.render(newLog))
}
```

isaachinman:
@MathiasKandelborg Thanks again for your help. I didn't notice until after the merge that the commit history is quite cluttered. In the future, please rebase your commits prior to a merge to simplify things.

I've added you as a contributor in the README.

MathiasKandelborg:
I'll remember to keep my commit history short and rebase, especially in OSS. Thanks for reminding. 

I'm happy to help. I've learned a lot so thanks again, I really appreciate it.

isaachinman:
Oh, sorry! I simply forgot to add you as a test contributor as well. I'll do that now.

PR #80, opened 2019-01-06
Use url package instead of urijs to reduce bundle size

kachkaev:
This PR is a follow-up to https://github.com/isaachinman/next-i18next/pull/50#issuecomment-449650826

Unlike `urijs`, `url` package is included into any next.js app by default:

```
cd new-project
yarn init -y
yarn add next react react-dom
```

```
yarn why url
yarn why v1.12.3
[1/4] 🤔  Why do we have the module "url"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "url@0.11.0"
info Reasons this module exists
   - "next" depends on it
   - Hoisted from "next#url"
   - Hoisted from "next#webpack#node-libs-browser#url"
```

```
yarn why urijs
yarn why v1.12.3
[1/4] 🤔  Why do we have the module "urijs"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
error We couldn't find a match!
```

This means that using `urijs` in `next-i18next` increases bundle size, while leveraging `url` does not. This PR switches between the packages, which should save people quite some bandwidth:

screenshot 2019-01-06 at 11 07 03

Next's `<Link />` accepts [URL objects](https://github.com/zeit/next.js#with-url-object) so there is no need to stringify URLs after they've been parsed. The tests are updated accordingly.

/cc @capellini, the author of https://github.com/isaachinman/next-i18next/pull/50


isaachinman:
I wasn't aware that the `Link` component could accept URL objects - that's a good find and makes our life a lot simpler.

This looks fine to me, but let's wait to hear from @capellini.

capellini:
Great catch @kachkaev - I also wasn't aware that you could pass a URL object as a prop into next's `Link` component.

@isaachinman - LGTM :+1:

PR #90, opened 2019-01-09
Refactoring next-18next-middleware

capellini:
Refactoring next-i18next middleware to behave more like common express
middleware.  This should allow users to use locale subpaths without any
extra configuration.  Configuring next-i18next now amounts to creating a
configuration file and adding a `server.use(nextI18NextMiddleware(nextI18next))`
line in the appropriate place in the server configuration file (i.e. before
any custom routing is performed).


capellini:
@isaachinman I refactored the middleware, modified the tests, and added the product page to the example.  It works in the simple example (almost -- see below), but I want to make sure that this is a robust solution that works on more comprehensive examples.  @kachkaev if you don't mind, I would really appreciate your input, as well.

One thing that I did notice (though it seems to not always occur):
1. Bring up `http://localhost:3000` or `http://localhost:3000/de`
2. Switch locales

Note that all of the translations work except for the `products` object, which appears to not be found.  If you subsequently reload the page, the products translation works fine, but when you switch locales for the first time, it doesn't work.  Strangely, if you click on a product link, then everything works (including navigating back to the index page and switching locales back and forth).  Are you seeing the same thing?  Any idea why this would be happening just for the product object translation? 

Let me know what you think.

capellini:
> hope to have the time to properly review this within the next week or so

Sounds good.  Sorry if I seemed like I was rushing you.  No worries -- take as much time as is needed.

> particularly in the context of the `as` prop and rendering routes _as if_ they were other routes

Note that it's still required to use the `as` prop, like I do [here](https://github.com/isaachinman/next-i18next/blob/efa23ed4dff4e7a79678dbb6e414f3520c0c298c/example/pages/index.js#L31) as well as define any custom routing like I do [here](https://github.com/isaachinman/next-i18next/blob/efa23ed4dff4e7a79678dbb6e414f3520c0c298c/example/server.js#L16-L18).  And since we're using the `next-i18next` `Link` component, it should properly handle the locale subpaths, right?

> I do tend to think that doing `req.url.replace` is going to run into some problems.

After some research in [the express docs for `req.originaUrl`](https://expressjs.com/en/4x/api.html#req.originalUrl) I noticed this comment:

> This property is much like req.url; however, it retains the original request URL, allowing you to rewrite req.url freely for internal routing purposes. For example, the “mounting” feature of app.use() will rewrite req.url to strip the mount point.

Seems like they are implying that rewriting `req.url` is it's intended purpose and, moreover, than stripping things off of `req.url` is OK.

capellini:
p.s. the bug that I was noticing seemed to disappear after I fixed some ordering and placement of namespaces in efa23ed4dff4e7a79678dbb6e414f3520c0c298c :man_facepalming: I guess I'm still very much an i18next n00b.

capellini:
> I do tend to think that doing `req.url.replace` is going to run into some problems.

I think I know what you mean now, so I changed `req.url.replace` and used `req.url.slice` instead.  I could have used a regex with `replace` to ensure that we replace only the first occurrence, but it's a bit more cumbersome and slow than just slicing off the part of req.url that we know we don't want.

example/pages/product-page.js
```diff
@@ -0,0 +1,34 @@
...
+    return {
+      id,
+      namespacesRequired: ['product-page', 'common']
+    }
+  }
+
+  render() {
+    const { id, t } = this.props
+    const products = t('products', { returnObjects: true });
+console.log('id =', id)
```
kachkaev: Is this console.log here on purpose? 🙂

capellini: :man_facepalming:  Thanks.  Fixed.

isaachinman:
> I think I know what you mean now, so I changed req.url.replace and used req.url.slice

Can you explain exactly what you saw, for clarity? Did you run into some buggy behaviour?

isaachinman:
[This line](https://github.com/isaachinman/next-i18next/blob/efa23ed4dff4e7a79678dbb6e414f3520c0c298c/example/pages/index.js#L31) is very good to bring up. I think it nullifies the "feature" of custom routing entirely if people have to do this. Users are not going to expect to have to manually put `as` on their links like this, they want to just be able to link to `/products/${productId}`.

capellini:
> [This line](https://github.com/isaachinman/next-i18next/blob/efa23ed4dff4e7a79678dbb6e414f3520c0c298c/example/pages/index.js#L31) is very good to bring up. I think it nullifies the "feature" of custom routing entirely if people have to do this.

Actually, next.js requires this for custom routing.  If you just put in an `href` prop without an associated `as` prop for a custom route, you'll note in dev tools that clicking the link will first produce a 404 error, then the page will reload, invoking the server-side custom routing logic.

Since the server-side custom logic is in place, it will appear that the `Link` without the `as` prop is working, but it's really not, since the `as` prop is required. See zeit/next.js#2833.  Specifically, [this comment](https://github.com/zeit/next.js/issues/2833#issuecomment-414919347).

So this is more of a feature of next.js than something to work around.

isaachinman:
Ah yes, of course. I had forgotten. That's the reason we're taking care of it for users in the `Link` component, etc. I need a solid chunk of time to review this instead of just a few minutes here or there - will have to wait a bit on that.

In the meantime I would welcome a second reviewer if anyone is available/interested.

isaachinman:
I should say that if this PR maintains 1:1 parity with the current `master` without any obvious bugs (tests are already all green), it will absolutely be merged. I prefer this approach, but might want to change some small organisational things about it.

capellini:
> Can you explain exactly what you saw, for clarity? Did you run into some buggy behaviour?

Actually this was just my misinterpretation of what `replace` does.  I thought that it would replace all occurrences of the string, but if the first argument is a string or a regex without the 'g' option specified, it only replaces the first occurrence.
```js
> '/de/page/delicious'.replace('/de', '')
'/page/delicious'
> '/de/page/delicious'.replace(/\/de/, '')
'/page/delicious'
> '/de/page/delicious'.replace(/\/de/g, '')
'/pagelicious'
```
So, it was safe to stick with `replace`, but `split` works just as well.  If you prefer to use `replace`, please let me know and I'll change it back.

andylacko:
hellow, really glad you are working on it, my CEO asks me, when this will be working, don't have time to help you with it :/ so I am just asking

isaachinman:
Hi @andylacko. This is open source software, and we are all working for free on our own time. Please be patient, or handle it within your own company.

isaachinman:
@kachkaev Did you get a chance to properly review this?

example/i18n.js
```diff
@@ -1,3 +1,6 @@
 const NextI18Next = require('next-i18next')
 
-module.exports = new NextI18Next({ otherLanguages: ['de'] })
+module.exports = new NextI18Next({
+  localeSubpaths: true,
```
isaachinman: As a small aside, I'd rather not have `localeSubpaths` in the example. The example is meant to be the simplest possible working app, and `localeSubpaths` is not an option which is enabled by default.

example/pages/index.js
```diff
@@ -7,11 +7,12 @@ import Footer from '../components/Footer'
 class Homepage extends React.Component {
   static async getInitialProps() {
     return {
-      namespacesRequired: ['common', 'footer']
+      namespacesRequired: ['common', 'footer', 'product-page']
```
isaachinman: I'm worried this will promote improper namespacing habits if other devs are looking to this example for conventions. If a namespace is called `{{something}}-page`, I would expect it to contain content only necessary for that page. The way you're doing it now, we'd be pulling in all content for the product page on the homepage itself, resulting in (potentially) a lot of unnecessary data.

Also, I have never seen an array within a namespace and am surprised that even works. The `products` array should be split into specific keys, and moved to the `common` namespace. The `h1` can remain in the `product-page` namespace as that is indeed where it belongs.

andylacko:
just asking 👌

example/pages/index.js
```diff
@@ -21,10 +22,23 @@ class Homepage extends React.Component {
         <Link href='/second-page'>
           <a>{t('to-second-page')}</a>
         </Link>
+        {
+          [1, 2, 3].map(
```
isaachinman: This is not a good pattern to use. We shouldn't have to loop/filter through arrays to reach content. Or, if you really wanted to, you should do so with keys, not an array:

```jsx
[1, 2, 3].map(num => t(`product-page:product${num}`)
```

This assumes a structure like:

```jsx
// products-page.json
...other keys
"product1" : "Product #1",
"product2" : "Product #2",
...and so on
```

This is just a general i18next thing.

example/pages/product-page.js
```diff
@@ -0,0 +1,34 @@
...
+
+class ProductPage extends React.Component {
+  static propTypes = {
+    id: PropTypes.number.isRequired,
+    t: PropTypes.func.isRequired
+  }
+
+  static async getInitialProps({ query: { id } }) {
+    return {
+      id: +id,
```
isaachinman: What's the `+` doing here?

example/pages/product-page.js
```diff
@@ -0,0 +1,34 @@
...
+      namespacesRequired: ['product-page', 'common']
+    }
+  }
+
+  render() {
+    const { id, t } = this.props
+    const products = t('products', { returnObjects: true });
```
isaachinman: Again, let's not use this pattern.

example/server.js
```diff
@@ -11,7 +11,11 @@ const handle = app.getRequestHandler();
   await app.prepare()
   const server = express()
 
-  nextI18NextMiddleware(nextI18next, app, server)
+  server.use(nextI18NextMiddleware(nextI18next))
+
+  server.get('/products/:id', (req, res) => {
+    app.render(req, res, '/product-page', { ...req.query, ...req.params })
```
isaachinman: Is there a better way of spreading/passing the `req` onwards?

Also just a small thing, but I'd probably prefer to have `server.get` return a func call:

```jsx
server.get('/products/:id', (req, res) =>
    app.render(req, res, '/product-page', { ...req.query, ...req.params }))
```

src/middlewares/next-i18next-middleware.js
```diff
@@ -1,24 +1,22 @@
...
-export default function (nexti18next, app, server) {
+export default function (nexti18next) {
 
   const { config, i18n } = nexti18next
   const { allLanguages, ignoreRoutes, localeSubpaths } = config
 
-  server.use(i18nextMiddleware.handle(i18n, { ignoreRoutes }))
-
-  const ignoreRegex = new RegExp(`^\/(?!${ignoreRoutes.map(x => x.replace('/', '')).join('|')}).*$`)
+  let middleware = [i18nextMiddleware.handle(i18n, { ignoreRoutes })]
```
isaachinman: Why not use a `const` and `push`:

```jsx
const middleware = [i18nextMiddleware.handle(i18n, { ignoreRoutes })]
...
middleware.push(
  forceTrailingSlash(allLanguages, ignoreRegex),
  lngPathDetector(ignoreRegex),
  handleLanguageSubpath(allLanguages)
)
```

src/utils/force-trailing-slash.js
```diff
@@ -1,15 +1,27 @@
...
-    if (pathname === `/${lng}`) {
-      res.redirect(301, pathname.replace(`/${lng}`, `/${lng}/`) + (search || ''))
-      performedRedirect = true
+const route = pathMatch()
+
+export default (allLanguages, ignoreRegex) => {
+  const ignoreRoute = route(ignoreRegex)
+
+  return (req, res, next) => {
+    const params = ignoreRoute(req.url)
```
isaachinman: Can you explain what this is doing? Shouldn't we not be calling `force-trailing-slash` on any `ignoreRoute` routes in the first place?

src/utils/force-trailing-slash.js
```diff
@@ -1,15 +1,27 @@
...
+    let performedRedirect = false
+
+    if (params) {
```
isaachinman: What does this evaluate to on simple routes, eg `myapp.com/`?

src/utils/handle-language-subpath.js
```diff
@@ -0,0 +1,17 @@
+import pathMatch from 'path-match'
+
+const route = pathMatch()
+
+export default allLanguages => (req, res, next) => {
+  const langRoute = route(`/:lng(${allLanguages.join('|')})/*`)
+
+  const params = langRoute(req.url)
+
+  if (params) {
```
isaachinman: Again, I'm unsure if this is the best pattern. I'd rather just not call the middleware int he first place.

kachkaev:
Hi @isaachinman! I won’t have time for a full review unfortunately. Just scrolled the changes on the phone and spotted console.log.

capellini:
@andylacko We appreciate your patience.  In the meantime, if you want to install this branch in your app and help us test it out, you can do so like this:
```bash
npm i git://github.com/capellini/next-i18next.git#middleware-refactor
```
or:
```bash
yarn add i git://github.com/capellini/next-i18next.git#middleware-refactor
```
if you run into any errors, please let us know.

example/pages/product-page.js
capellini: Ensuring that `id` is a number.  I can use `parseInt` if you prefer.

src/middlewares/next-i18next-middleware.js
capellini: Mutable vs. immutable...it doesn't really matter to me, though.  I'll change it.

src/utils/force-trailing-slash.js
capellini: What we are doing here is route filtering.  I opted for the simplest API possible for `nextI18nextMiddleware`.  That is, a user simply calls:
```js
```
This precludes the use of `server.get` within the `nextI18NextMiddleware` to filter `ignoreRoutes` -- every route is passed into the middleware.  Therefore, we must let the middleware determine which route is an `ignoreRoute`.

src/utils/force-trailing-slash.js
capellini: Assuming that 
```js
ignoreRegex === /^\/(?!_next|static).*$/
```
then `myapp.com/` would evaluate to `false`.  However, `req.url` in this case should be `/myapp.com/` in which case it would evaluate to `{}`.

src/utils/handle-language-subpath.js
capellini: This is path matching to ensure that we have a url that starts with a locale.  If not, we ignore the route.  If so, then we continue and modify the `query` and `url`.

capellini:
@isaachinman After reviewing your comments, it looks like you were expecting the route filtering to be performed in the main `nextI18nextMiddleware` function, whereas I had opted to just return an array of functions and let the individual functions do the route filtering themselves.

I've made the necessary changes in feed006260c8f14263f9a7998a657f18968f898c so that `nextI18nextMiddleware` now handles filtering routes.  Please let me know what you think when you get the opportunity.  Thanks again for your feedback.

capellini:
Reviewing the code some more, it occurred to me that we were also doing route filtering in `forceTrailingSlash`, so I opted to remove the route filtering done in that function and place it in `nextI18nextMiddleware` for consistency.  

There is still some route filtering done in `lngPathDetector`, but to the greatest extent possible, all route route filtering is done in `nextI18nextMiddleware`.

example/server.js
capellini: > Is there a better way of spreading/passing the `req` onwards?

You mean something [like this instead](https://github.com/isaachinman/next-i18next/blob/331fce22a767da15274cbe5cb3e857914c3d8364/example/server.js#L17-L19)?

> I'd probably prefer to have server.get return a func call

Done.

lucasfeliciano:
I'm really enjoying this PR.
Delegating the responsibility of rendering the page to the route give us much more control over things.
One example from the top of my mind ( because I was work on it recently ) is custom caching strategy this will make things much simpler.

src/utils/force-trailing-slash.js
capellini: Route filtering is now done inside of `nextI18NextMiddleware` [here](https://github.com/capellini/next-i18next/blob/dda95de7130a37e5e7d8e9d8fdd46bb98a3e46ef/src/middlewares/next-i18next-middleware.js#L24)

example/i18n.js
isaachinman: > Perhaps we can have more than one example in the example directory?

 Yes, that'd be the way forward. However, I really don't want to maintain multiple examples at this point in time. Until this project hits a 1.x release, I don't mind just answering people's questions.

example/server.js
isaachinman: I guess I was just surprised that you have to spread both `query` and `params` in. Why both?

src/utils/force-trailing-slash.js
isaachinman: So, are you expecting that value to be falsey? I'd rather have type-safe variables if we can.

src/utils/handle-language-subpath.js
isaachinman: Even in the `path-match` docs, they're specifically doing `params === false`.

Can't we perform one check for all these middlewares instead of having to check within each one?

isaachinman:
@capellini Thanks for making those changes.

This PR is getting pretty massive now, and I don't feel comfortable merging it until myself and hopefully at least one other have reviewed it thoroughly and properly.

@lucasfeliciano Do you have time for a review?

isaachinman:
Alternatively, if you think it would be possible to split this into several smaller PRs, that would really help get things merged incrementally/faster.

example/server.js
capellini: `query` has our original query parameters, `params` has our `id`.  I changed it to [this line](https://github.com/isaachinman/next-i18next/blob/331fce22a767da15274cbe5cb3e857914c3d8364/example/server.js#L19) which makes this more explicit.

src/utils/force-trailing-slash.js
capellini: Done

src/utils/handle-language-subpath.js
capellini: See https://github.com/isaachinman/next-i18next/pull/90/commits/feed006260c8f14263f9a7998a657f18968f898c

capellini:
> if you think it would be possible to split this into several smaller PRs

I think I can break off dda95de7130a37e5e7d8e9d8fdd46bb98a3e46ef (refactoring `appWithTranslation` and `lngPathCorrector` to work with custom routes) into a separate PR and probably deserves to be, since it's not middleware-related.

At a glance, everything else seems pretty integral to this PR.

capellini:
I also want to move [this code](https://github.com/isaachinman/next-i18next/blob/eb8c1178a525906679ef92ed25eb46348846ffe7/src/middlewares/next-i18next-middleware.js#L37-L42) inside the `if (isI18nRoute(req.url))` statement, but I will do that in a separate PR after this is merged to prevent this one from growing any larger.

lucasfeliciano:
@isaachinman On it

src/middlewares/next-i18next-middleware.js
```diff
@@ -1,24 +1,46 @@
...
+  const isI18nRoute = url => ignoreRoute(url)
+
+  const localeSubpathRoute = route(`/:lng(${allLanguages.join('|')})/*`)
+  const isLocaleSubpathRoute = params => params !== false
+
+  const isLocaleRootRouteWithoutSlash = pathname => allLanguages.some(lng => pathname === `/${lng}`)
+
+  return [
+    i18nextMiddleware.handle(i18n, { ignoreRoutes }),
+    (req, res, next) => {
```
lucasfeliciano: What you think about extract this to its own middleware and conditionally include to the array if needed?

Doing that we have some benefits:

1. We can make the tests clear testing the imported middleware instead deconstructing the array like you did here `const [, middleware] = nextI18nextMiddleware(nexti18next)`. At first I had no idea what this `middleware` was about, also, it is not dependent on nexti18next instance.

2. Since we are adding the middleware conditionally to the array, we would have -1 middleware in the chain since it is not needed

lucasfeliciano: What do you think about extracting this to its own middleware and conditionally add to the array if needed?

Doing that we have some benefits:

1. We can make the tests clear testing the imported middleware instead deconstructing the array like you did here `const [, middleware] = nextI18nextMiddleware(nexti18next)`. At first I had no idea what this `middleware` was about just by reading the variable name, also, this middleware it is not dependent on nexti18next instance so would make the test really encapsulated. :)

2. Since we are adding the middleware conditionally to the array, we would have -1 middleware in the chain since it is not needed 

3. The logic would be encapsulated where it is needed therefore the imports

capellini: @lucasfeliciano I like that a lot.  Thank you for the suggestion.

isaachinman:
Thanks for the review @lucasfeliciano. Remind me - you have had PRs merged on this repo, right? I need to add you as a contributor to the README.

@capellini Although I haven't had time for a proper review, this PR seems to keep moving. Let me know when you think it is stable/complete and I will set aside a chunk of time.

capellini:
@lucasfeliciano Thank you for the review. I've incorporated your changes and also cleaned up the tests a little.

@isaachinman The only thing I hope to do with this branch henceforth is rebase it with master when master changes (even if there are no conflicts) since I'm currently using this branch on my project and would like to keep it up to date with the latest changes elsewhere in this repo.

We should be good for a review when you have the opportunity.  I regret the amount of necessary changes and appreciate that reviewing it may take some time.  Thanks.

isaachinman:
@capellini I've installed autorebase so PRs on this repo should rebase automatically from now on. That being said, there appear to be two conflicts to be resolved. Can you resolve those sometime soon?

isaachinman:
@capellini I'm reviewing this now.

First, I'd like the actual custom routing to be removed from the main example, and added to the documentation for now. In the very near future we might end up having multiple example dirs, and one can be "custom routing". As long as we have a single example, though, it should be the simplest use case, as it's become obvious a lot of people are simply copy/pasting it.

isaachinman:
Having multiple example dirs will not solve #118 by the way, as we'd still have a single `jest-puppeteer` instance. Still not sure about how to solve that.

I'm going to remove all your work in the example dir so that the diff between your branch and master is as small as possible.

This should land on master this afternoon unless I find a problem.

We can add your products example back in shortly thereafter as a "custom routing" example.

isaachinman:
@capellini Looks like `handle-language-subpath` has changed a lot since I last looked at it. Can you explain what it's doing with this `slice` call? Can you think of a more specific name? "Handle" doesn't make it clear to me what it's doing. It seems to be reassigning `req.url` and `req.query`.

isaachinman:
This util is only used in one place, and it's only two lines of logic. I don't think this needs to be pulled out as a util in the first place, and the test coverage is somewhat circular. Hope you don't mind if I just pop it straight into `next-i18next-middleware`.

isaachinman:
@capellini Also, I'd much prefer to do:

```jsx
req.url = req.url.replace(`/${lng}`, '')
```

I think it's a lot more readable and makes it explicitly clear what's happening. Also, a lot less prone to bugs as we're matching a specific pattern, not just slicing out a length.

I'm going to proceed this way unless you foresee any complications.

isaachinman:
@capellini Merged... thanks for your patience throughout that process. I'm going to release this via v0.23.0. If we start getting a lot of bug reports, we'll have to revert. Otherwise, job done.

capellini:
> This util is only used in one place, and it's only two lines of logic. I don't think this needs to be pulled out as a util in the first place, and the test coverage is somewhat circular. Hope you don't mind if I just pop it straight into `next-i18next-middleware`.

I was thinking the same thing myself after thinking about it.

lucasfeliciano:
Nice work @capellini !
Tomorrow it will be a nice day at the office for me since I'll have the opportunity to refactor to this pattern.
Also when it gets into staging it will be a good opportunity to test it and see if we finds any bugs.

isaachinman:
Yes @lucasfeliciano, we will eagerly await your findings. Thanks everyone for really good participation on this one.

capellini:
Thanks very much @isaachinman.  I'm really enjoying seeing this project progress.  If there are some bugs reported, if you don't mind, I'd like the opportunity to fix them before we revert. 

And thank you for the kind words and the review @lucasfeliciano - please do let us know if you run into problems.

Also, many thanks to @kachkaev for the review, as well.

PR #91, opened 2019-01-12
Prevent caching 301 redirect

petertulala:


isaachinman:
This is fixed with 5352521. Closing in favour of that.

PR #95, opened 2019-01-14
Properly parse href in Link when it's an object

### LINKED ISSUE #89: Passing href as object in Link causing wrong navigation when enable Locale subpaths
When enable Locale subpaths and passing href as object in Link like below
```
<Link href={{ pathname: '/foo', query: { bar: '123' } }} passHref>
   ...
</Link>
```
When we are on default language path like `http://localhost:3000/`. It will correctly navigate to `http://localhost:3000/foo?bar=123`

But when we are on the other language path like `http://localhost:3000/fr/`. It will navigate to wrong path like `http://localhost:3000/fr/fr[object%20Object]`

capellini:
Fixes #89.

### DIFF
--- a/__tests__/components/Link.test.js
+++ b/__tests__/components/Link.test.js
@@ -110,7 +110,7 @@ describe('Link component', () => {
   })
 
   describe('https://github.com/isaachinman/next-i18next/issues/89', () => {
-    describe('href as object', () => {
+    describe('when href is an object, properly parse it', () => {
       beforeEach(() => {
         props.href = {
           pathname: '/foo/bar',
--- a/__tests__/components/Link.test.js
+++ b/__tests__/components/Link.test.js
@@ -30,29 +30,86 @@ describe('Link component', () => {
     <Link {...props} {...otherProps}>click here</Link>,
   ).find('Link').at(1)
 
-  describe('localeSubpaths === false', () => {
-    describe('href as string', () => {
-      beforeEach(() => {
-        props.href = '/foo/bar'
-      })
+  it('renders without lang if localeSubpaths === false', () => {
+    // without 'as' prop
+    let component = createLinkComponent()
 
-      it('renders without lang', () => {
-        const component = createLinkComponent()
+    expect(component.prop('href')).toEqual('/foo/bar')
+    expect(component.prop('as')).toBeUndefined()
 
-        expect(component.prop('href')).toEqual('/foo/bar')
-        expect(component.prop('as')).toBeUndefined()
-      })
+    // with 'as' prop
+    props.as = '/foo?bar'
+    component = createLinkComponent()
 
-      it('renders without lang (with "as" prop)', () => {
-        props.as = '/foo?bar'
+    expect(component.prop('href')).toEqual('/foo/bar')
+    expect(component.prop('as')).toEqual('/foo?bar')
+  })
 
-        const component = createLinkComponent()
+  it('renders without lang if props.lng is undefined', () => {
+    props.nextI18NextConfig.config.localeSubpaths = true
+    props.lng = undefined
 
-        expect(component.prop('href')).toEqual('/foo/bar')
-        expect(component.prop('as')).toEqual('/foo?bar')
-      })
-    })
+    // without 'as' prop
+    let component = createLinkComponent()
 
+    expect(component.prop('href')).toEqual('/foo/bar')
+    expect(component.prop('as')).toBeUndefined()
+
+    // with 'as' prop
+    props.as = '/foo?bar'
+    component = createLinkComponent()
+
+    expect(component.prop('href')).toEqual('/foo/bar')
+    expect(component.prop('as')).toEqual('/foo?bar')
+  })
+
+  it('renders without lang if props.lng === defaultLanguage', () => {
+    props.nextI18NextConfig.config.localeSubpaths = true
+    props.nextI18NextConfig.config.defaultLanguage = 'en'
+    props.lng = 'en'
+
+    // without 'as' prop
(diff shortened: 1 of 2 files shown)


isaachinman:
Thanks @capellini. Can you explain what all the changes to the test suite are? They are quite hard to follow. I would not expect existing tests to be modified - if anything this PR should be accompanied by one or several tests to cover the pre-existing bug, and not much else.

capellini:
> Can you explain what all the changes to the test suite are?

I only added tests to cover the bug in #89.  The rest is reorganizing the test file for better readability.  Looking at the test file in it's entirety, rather than the diff, should clear things up.  But I'll change it back to the old file and tack on the added tests to the end.  Perhaps I'll submit another PR to fix the test readability at some point.

capellini:
@isaachinman Test file changed.

__tests__/components/Link.test.js
```diff
@@ -133,4 +133,76 @@ describe('Link component', () => {
     expect(component.prop('href')).toEqual({ pathname: '/foo/bar', query: { lng: 'de' } })
     expect(component.prop('as')).toEqual('/de/foo?bar')
   })
+
+  describe('https://github.com/isaachinman/next-i18next/issues/89', () => {
```
isaachinman: Please use a real/descriptive name here so that people can understand the test output. You can include the link in a comment if you'd like.

capellini: If you look at the `it` block contained in this `describe`, it is a real/descriptive name:

screenshot from 2019-01-16 13-08-57

I also prefer this method because I can now click on the link in the test output on my terminal if I want more detail and it opens the GitHub issue page with one click.

Does that suffice?  If not, no problem - I just want to get an idea of how you would like things structured so that my future tests are structured similarly.

capellini: I modified the `describe` block within this `describe` to be a bit more descriptive.  Now the output looks like this:

screenshot from 2019-01-16 13-34-41

Would this suffice?  If not, just let me know and I think that I'm safe just removing the outer `describe` block.

capellini:
> I think it's going to need to be rebased

I rebased, but now we're running into code complexity issues.  I tried to reduce the complexity of `Link`'s `render` method by extracting some of the logic into static methods (75f15fa), but I still have to reduce the complexity a bit more.  I'm still looking through the code to see how I can do that without making the code awkward (e.g. moving some of the static methods out of the `export default` function entirely).

capellini:
Code complexity issue fixed.

__tests__/components/Link.test.js
capellini: Based upon [feedback](https://github.com/isaachinman/next-i18next/pull/102#issuecomment-455350029) on #102, I'm going to mark this one as resolved.  If people seem to not like this structure, I will submit a PR to change it.

src/components/Link.js
```diff
@@ -33,23 +35,31 @@ const removeWithNamespacesProps = (props) => {
   return strippedProps
 }
 
+const parseHref = (href) => {
+  if (typeof href !== 'object') {
```
isaachinman: Why not just explicitly check if it's a `string`? Also, because this function is so simple, I'd probably prefer it to return a ternary inline.

src/components/Link.js
```diff
@@ -69,6 +79,12 @@ export default function () {
     }
   }
 
+  Link.localeSubpathRequired = (lng) => {
```
isaachinman: Why did you split this out? Just to make CodeClimate happy? If you wanted to split this logic out, it should just be a standalone function, not a class method.

capellini: I didn't split it out to make CodeClimate happy I did it to increase code readability.

But I did make `parseHref` and `formatAsProp` standalone functions to make CodeClimate happy.  I would have rather made them static methods.  What is the objection to static methods?  I would understand using standalone functions if we were using functions in multiple places and extracted the functions into utility functions, but if the logic is used exclusively in the `Link` object, why not a static method?

src/components/Link.js
capellini: > Why not just explicitly check if it's a `string`?

Six to one -- half dozen to another.  I'll change it.

> Also, because this function is so simple, I'd probably prefer it to return a ternary inline.

Ok.

capellini: Fixed with 0bbe890944db372f9b9f603a60d113327afc1618

src/components/Link.js
capellini: "Fixed" with bf875c1cf0dfc01809d31763de45e39acdd9b976

capellini:
@isaachinman I just rebased this branch.  Given that we're now using a HoC, I think that the changes in cdeb0ce24f62fd3ba1b153c7ccecf9515cf6c5d1 should probably be reverted.

The reason is that, before rebasing, `localeSubpathRequired` was a static method because config was not internal to `Link` -- it was provided by being wrapped into a default function export.  Now that we're using a HoC, config is being passed in as a prop to `Link`.  Therefore, `localeSubpathRequired` is no longer static, as it requires `this.props.nextI18NextConfig`.  Therefore, in light of the rebased changes, I think that it is best to revert the changes in cdeb0ce24f62fd3ba1b153c7ccecf9515cf6c5d1 and leave localeSubpathRequired as a method in the `Link` component.

Also, if you don't mind, I think it makes sense to also make `formatAsProp` and `parseHref` static methods in the `Link` component, as the logic appears to be used exclusively within the `Link` component.

Of course, if we want to just merge this and move on to other 1.0 milestones, we can defer this until 1.0 is fielded.

Let me know what you think.

PR #96, opened 2019-01-14
Refactor appWithTranslation to work with custom routes

### LINKED ISSUE #126: query string is disappear after change language when localeSubpaths is true
When I enable `localeSubpaths`. If I change language, the query string will be removed from the url.

### Expected behavior
1. Start on the url like `http://localhost:3000/?foo=123`
2. Change language by using `i18n.changeLanguage('de')`
3. The url will changed to `http://localhost:3000/de/?foo=123`

### Actual behavior
1. Start on the url like `http://localhost:3000/?foo=123`
2. Change language by using `i18n.changeLanguage('de')`
3. The url will changed to `http://localhost:3000/de/`

capellini:
We cannot use window.location.pathname in lngPathCorrector if the route is a custom route, since window.location.pathname !== router.pathname (instead, it is equal to router.asPath).  Since this is the case, [we cannot use shallow routing](https://github.com/zeit/next.js/blob/canary/packages/next-server/lib/router/router.js#L374-L375) and a new page is requested, resulting in a 404 error whenever the i18n.changeLanguage function is invoked.

This commit changes appWithTranslation to instead use the withRouter Next HoC, which exposes a router prop.  The router prop has pathname, query, and asPath attributes which we can use instead of using window.location.pathname.  We then modify query and asPath in lngPathCorrector (as pathname will remain unchanged if we are only changing the language) and shallow route if the asPath has changed after correction (with query being modified, as necessary).

Also fixes #126 by ensuring that existing query parameters are preserved.

### DIFF
--- a/src/utils/lng-path-corrector.js
+++ b/src/utils/lng-path-corrector.js
@@ -1,6 +1,7 @@
-export default (config, i18n, asPath, query, currentLanguage = i18n.languages[0]) => {
+export default (config, i18n, currentRoute, currentLanguage = i18n.languages[0]) => {
 
   const { defaultLanguage, allLanguages } = config
+  const { asPath, query } = currentRoute
 
   if (!allLanguages.includes(currentLanguage)) {
     throw new Error('Invalid configuration: Current language is not included in all languages array')
--- a/src/utils/lng-path-corrector.js
+++ b/src/utils/lng-path-corrector.js
@@ -1,4 +1,4 @@
-export default (config, i18n, currentRoute, currentLanguage = i18n.languages[0]) => {
+export default (config, i18n, asPath, query, currentLanguage = i18n.languages[0]) => {
 
   const { defaultLanguage, allLanguages } = config
 
@@ -6,22 +6,18 @@ export default (config, i18n, currentRoute, currentLanguage = i18n.languages[0])
     throw new Error('Invalid configuration: Current language is not included in all languages array')
   }
 
-  let href = currentRoute
-  let as = href
+  let as = asPath
 
   for (const lng of allLanguages) {
-    if (href.startsWith(`/${lng}/`)) {
-      href = href.replace(`/${lng}/`, '/')
+    if (asPath.startsWith(`/${lng}/`)) {
+      as = as.replace(`/${lng}/`, '/')
       break
     }
   }
 
   if (currentLanguage !== defaultLanguage) {
-    as = `/${currentLanguage}${href}`
-    href += `?lng=${currentLanguage}`
-  } else {
-    as = href
+    return [`/${currentLanguage}${as}`, { ...query, lng: currentLanguage }]
   }
 
-  return [href, as]
+  return [as, { ...query, lng: defaultLanguage }]
 }
--- a/src/utils/lng-path-corrector.js
+++ b/src/utils/lng-path-corrector.js
@@ -20,5 +20,5 @@ export default (config, i18n, currentRoute, currentLanguage = i18n.languages[0])
     return [`/${currentLanguage}${as}`, { ...query, lng: currentLanguage }]
   }
 
-  return [as, { ...query, lng: defaultLanguage }]
+  return [as, query]
 }
--- a/__tests__/utils/lng-path-corrector.test.js
+++ b/__tests__/utils/lng-path-corrector.test.js
@@ -5,7 +5,7 @@ import lngPathCorrector from '../../src/utils/lng-path-corrector'
 describe('lngPathCorrector utility function', () => {
   let config
   let i18n
-  let query
+  let currentRoute
 
   beforeEach(() => {
     config = {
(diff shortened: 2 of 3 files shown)


src/utils/lng-path-corrector.js
```diff
@@ -1,27 +1,23 @@
...
   if (currentLanguage !== defaultLanguage) {
-    as = `/${currentLanguage}${href}`
-    href += `?lng=${currentLanguage}`
-  } else {
-    as = href
+    return [`/${currentLanguage}${as}`, { ...query, lng: currentLanguage }]
   }
 
-  return [href, as]
+  return [as, { ...query, lng: defaultLanguage }]
```
capellini: Is it necessary to include the `lng` query parameter with the defaultLanguage?  I wasn't sure.

capellini:
Note that the return value of `lngPathCorrector` has changed from returning `[href, as]` to returning `[as, query]`.

isaachinman:
> Note that the return value of `lngPathCorrector` has changed from returning `[href, as]` to returning `[as, query]`.

Why?

capellini:
`withRoute` gives us a `router` instance which, while also containing a 'route' key that contains the string representation of the URL, also contains the `pathname` string and `query` object.

Since `lngPathCorrector` really only modifies `as` and the query parameters and `router.replace` can take an object instead of a string, I thought it easier and less error-prone to modify a query object than to modify a URL's query string.

It's akin to the changes made in #80.  Once we realized that we could pass `href` as an object in `Link`, rather than having to mess around with strings, things became much easier.

capellini:
@isaachinman Let me know if you would prefer that I not to change the return value.

src/utils/lng-path-corrector.js
isaachinman: Why do you want to do that? That seems incorrect to me. What if the current language is not the `defaultLanguage`?

src/utils/lng-path-corrector.js
```diff
@@ -1,27 +1,23 @@
-export default (config, i18n, currentRoute, currentLanguage = i18n.languages[0]) => {
+export default (config, i18n, asPath, query, currentLanguage = i18n.languages[0]) => {
```
isaachinman: Probably makes sense to start using a single-argument object instead of adding more args here. TypeScript in the future will make this more reliable.

src/utils/lng-path-corrector.js
capellini: > What if the current language is not the `defaultLanguage`?

If `currentLanguage !== defaultLanguage` because of [these lines](https://github.com/isaachinman/next-i18next/blob/c9faead5bd759ecee1f6fcae688501d39d5d787c/src/utils/lng-path-corrector.js#L18-L20), that line will never be reached.  We only reach that line of code if the current language is the default language.

src/utils/lng-path-corrector.js
capellini: Fixed with 180c6abe1031e294389db58570360e2aacf0c243

src/utils/lng-path-corrector.js
capellini: Fixed with 004ea55c30d0cc2cf79d1467b25a146f642646d2

isaachinman:
I did:

1. `git fetch origin pull/96/head && git checkout FETCH_HEAD`
2. Modify `example/i18n.js` to include `localeSubpaths: true`
3. `yarn run-example`

I got:

> Invalid configuration: Current language is not included in all languages array

In a loop. Any idea what's going on here? Moreover, the e2e tests fail locally. Seems like we need to add an e2e test to cover `localeSubpaths`.

capellini:
Things should work now.  The problem was that I refactored `lngPathCorrector` to use a single argument for `asPath` and `routerQuery`, but I didn't make the appropriate change in the call to in `app-with-translation.js`.  I have corrected the bug and rebased the change [here](https://github.com/isaachinman/next-i18next/blob/c0afa43beeee720994b43a5d8ca929e63a14ae0a/src/hocs/app-with-translation.js#L24), so now the code works.

> Seems like we need to add an e2e test to cover `localeSubpaths`.

Yes, I think that would help a lot.  I normally manually change `i18n.js` to add `localeSubpaths` on the example to test PRs that require `localeSubpaths` and then remove the change prior to committing, but it appears I did not when I initially committed 
c0afa43 to this branch.

Ideally, I'd like to have unit and integration/e2e tests to cover as much as possible.  This way, a `yarn test` is all that would be needed to catch most bugs.

capellini:
@isaachinman In the case of our example in the repo, I can confirm that the issue reported on #126 happens on `master` (extra query parameters are removed from the URL), but when switching to this branch, it fixes the problem (query parameters are no longer removed).

Moreover [these tests](https://github.com/isaachinman/next-i18next/blob/1559b70104b3b2b252794325ea7f361035ea8e36/__tests__/utils/lng-path-corrector.test.js#L47-L57) should _help_ prevent regressions, though a full test of `app-with-translation.js` would probably have a test more specific to the case of #126.

isaachinman:
@capellini Thank you very much for your tireless work.

Issue #97, opened 2019-01-15
Passing withNamespaces props into Next Link causes devtools warning


capellini:
Spreading `props` in the `Link` component results in passing `withNamespaces` props down to Next.js' `Link` component, causing the following devtools warning:

![screenshot from 2019-01-15 08-17-37](https://user-images.githubusercontent.com/75311/51186783-6a63ab80-18a8-11e9-99cb-d338612dbb74.png)

isaachinman:
Fixed with f5fbdfc.

PR #102, opened 2019-01-16
Adding test to ensure removal of withNamespaces props

### LINKED ISSUE #101: Add test to ensure withNamespaces props are removed before passing to Next Link
Following up on #97.  I thought that it would be good to add a test to our suite to ensure that `withNamespaces` props are stripped before passing props to Next's `Link` component to prevent any regression bugs.

PR forthcoming.

capellini:
Adding tests to ensure that withNamespaces props are not passed on to Next's Link component.

Reference #97.

Resolves #101.

### DIFF
--- a/__tests__/components/Link.test.js
+++ b/__tests__/components/Link.test.js
@@ -31,10 +31,10 @@ describe('Link component', () => {
     }
   })
 
-  const createLinkComponent = () => {
+  const createLinkComponent = (otherProps = {}) => {
     const Link = linkFunction.apply(context)
 
-    return mount(<Link {...props}>click here</Link>).find('Link').at(1)
+    return mount(<Link {...props} {...otherProps}>click here</Link>).find('Link').at(1)
   }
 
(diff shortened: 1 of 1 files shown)


capellini:
Admittedly, the values that I set for the `withNamespaces` props do not reflect what their true values would be, but I think that's ok for testing purposes.  The important thing is that the props are removed prior to passing them on to Next's `Link`.

capellini:
Test output still references the GitHub issue with a more detailed description contained within.  Output looks like:
screenshot from 2019-01-17 14-06-23

Depending on what you think of [this comment](https://github.com/isaachinman/next-i18next/pull/95#discussion_r248394147) in #95, I can change it, but I respectfully ask that you please consider leaving the issue in to help with easy navigation to the issue via a click in a terminal window for reference.  Whatever you decide, though.  Just let me know when you have the opportunity.  Thanks.

isaachinman:
We can add these links for now if it makes you happy. They may be removed in the future if it bothers others. Thanks for this!

PR #109, opened 2019-01-18
Add customDetectors option 

lucasfeliciano:
I want to start a discussion about this feature

## Motivation

A while a go we were discussing custom language detectors and @isaachinman implemented the option to add plugins to next-i18next (https://github.com/isaachinman/next-i18next/commit/14d494fc0e871a509d4214d1523fa7cade33e13b).

However this was exclusively for plugins, which doesn't work with the custom language detectors interface provided by `i18n`.

Adding this feature makes it easier for users develop their own language detector and use it with the following interface and options:

Interface
```js
export default {
  name: 'myDetectorsName',

  lookup(options) {
    // options -> are passed in options
    return 'en';
  },

  cacheUserLanguage(lng, options) {
    // options -> are passed in options
    // lng -> current language, will be called after init and on changeLanguage

    // store it
  }
};
```

Options
```js
{
  // order and from where user language should be detected
  order: ['querystring', 'cookie', 'localStorage', 'navigator', 'htmlTag', 'path', 'subdomain'],

  // keys or params to lookup language from
  lookupQuerystring: 'lng',
  lookupCookie: 'i18next',
  lookupLocalStorage: 'i18nextLng',
  lookupFromPathIndex: 0,
  lookupFromSubdomainIndex: 0,

  // cache user language on
  caches: ['localStorage', 'cookie'],
  excludeCacheFor: ['cimode'], // languages to not persist (cookie, localStorage)

  // optional expire and domain for set cookie
  cookieMinutes: 10,
  cookieDomain: 'myDomain',

  // optional htmlTag with lang attribute, the default is:
  htmlTag: document.documentElement
}
```

## Questions

1. There is a specific reason to use `eval` here: `eval("require('i18next-express-middleware')")`
2. Why are we using `isNode` ? Isn't `process.browser` enough?

### DIFF
--- a/__tests__/create-i18next-client.test.js
+++ b/__tests__/create-i18next-client.test.js
@@ -1,0 +1,110 @@
+/* eslint-env jest */
+import i18next from 'i18next'
+import I18nextBrowserLanguageDetector from 'i18next-browser-languagedetector'
+import createI18nextClient from '../src/create-i18next-client'
+
+const i18nextMiddleware = require('i18next-express-middleware')
+
+jest.mock('i18next', () => ({
+  init: jest.fn(),
+  use: jest.fn(),
+}))
+
+jest.mock('i18next-express-middleware', () => ({
+  LanguageDetector: jest.fn(),
+}))
+
+jest.mock('i18next-browser-languagedetector', () => jest.fn())
+
+describe('initializing i18n', () => {
+  beforeEach(() => {
+    i18next.isInitialized = false
+  })
+
+  afterEach(() => {
+    i18next.init.mockClear()
+    i18next.use.mockClear()
+    i18nextMiddleware.LanguageDetector.mockClear()
+  })
+
+  it('should not initialize i18n if i18n is already initialized', () => {
+    i18next.isInitialized = true
+
+    createI18nextClient({
+      use: [],
+      customDetectors: [],
+      serverLanguageDetection: true,
+      browserLanguageDetection: true,
+    })
+
+    expect(i18next.init).not.toBeCalled()
+    expect(i18next.use).not.toBeCalled()
+  })
+
+  it('should initialize i18n if i18n is not yet initialized', () => {
+    i18next.isInitialized = false
+
+    createI18nextClient({
+      use: [],
+      customDetectors: [],
+      serverLanguageDetection: true,
+      browserLanguageDetection: true,
+    })
+
+    expect(i18next.use).toBeCalledTimes(2)
+    expect(i18next.init).toBeCalledTimes(1)
+  })
+
+  it('should add backend language detector if on node environment', () => {
+    i18next.isInitialized = false
+
+    createI18nextClient({
+      use: [],
+      customDetectors: [],
+      serverLanguageDetection: true,
+    })
+
+    expect(i18nextMiddleware.LanguageDetector).toBeCalled()
+    expect(I18nextBrowserLanguageDetector).not.toBeCalled()
+    expect(i18next.use).toBeCalledTimes(2)
+    expect(i18next.init).toBeCalled()
+  })
+
+  it('should call addDetector with the right detector if serverLanguageDetection is true', () => {
+    i18next.isInitialized = false
+    const addDetector = jest.fn()
+    i18nextMiddleware.LanguageDetector.mockImplementation(() => ({
+      addDetector,
+    }))
+
+    createI18nextClient({
+      use: [],
+      customDetectors: ['test_custom_detector'],
+      serverLanguageDetection: true,
+    })
+    expect(i18nextMiddleware.LanguageDetector).toBeCalled()
+    expect(I18nextBrowserLanguageDetector).not.toBeCalled()
+    expect(addDetector).toBeCalledWith('test_custom_detector')
+    expect(i18next.use).toBeCalledTimes(2)
+    expect(i18next.init).toBeCalledTimes(1)
+  })
+
+  it('should not call addDetector with the right detector if serverLanguageDetection is false', () => {
+    i18next.isInitialized = false
+    const addDetector = jest.fn()
+    i18nextMiddleware.LanguageDetector.mockImplementation(() => ({
+      addDetector,
+    }))
+
+    createI18nextClient({
+      use: [],
+      customDetectors: ['test_custom_detector'],
+      serverLanguageDetection: false,
+    })
+    expect(i18nextMiddleware.LanguageDetector).not.toBeCalled()
+    expect(I18nextBrowserLanguageDetector).not.toBeCalled()
+    expect(addDetector).not.toBeCalled()
+    expect(i18next.use).toBeCalledTimes(1)
+    expect(i18next.init).toBeCalledTimes(1)
+  })
+})
--- a/src/create-i18next-client.js
+++ b/src/create-i18next-client.js
@@ -1,7 +1,7 @@
 import isNode from 'detect-node'
 import i18next from 'i18next'
 import i18nextXHRBackend from 'i18next-xhr-backend'
-import i18nextBrowserLanguageDetector from 'i18next-browser-languagedetector'
+import I18nextBrowserLanguageDetector from 'i18next-browser-languagedetector'
 
 const i18n = i18next.default ? i18next.default : i18next
 i18n.nsFromReactTree = []
@@ -14,12 +14,16 @@ export default (config) => {
       const i18nextMiddleware = eval("require('i18next-express-middleware')")
       i18n.use(i18nextNodeBackend)
       if (config.serverLanguageDetection) {
-        i18n.use(i18nextMiddleware.LanguageDetector)
+        const serverDetectors = new i18nextMiddleware.LanguageDetector()
+        config.customDetectors.forEach(detector => serverDetectors.addDetector(detector))
+        i18n.use(serverDetectors)
       }
     } else {
       i18n.use(i18nextXHRBackend)
       if (config.browserLanguageDetection) {
-        i18n.use(i18nextBrowserLanguageDetector)
+        const browserDetectors = new I18nextBrowserLanguageDetector()
+        config.customDetectors.forEach(detector => browserDetectors.addDetector(detector))
+        i18n.use(browserDetectors)
       }
     }
 
--- a/README.md
+++ b/README.md
@@ -149,6 +149,7 @@ class SomeLink extends React.Component {
 | `localeSubpaths` | `false`  |
 | `serverLanguageDetection` | `true`  |
 | `use` (for plugins) | `[]`  |
+| `customDetectors` | `[]`  |
 
 _This table contains options which are specific to next-i18next. All other [i18next options](https://www.i18next.com/overview/configuration-options) can be passed in as well._
 
--- a/src/config/default-config.js
+++ b/src/config/default-config.js
@@ -24,6 +24,7 @@ export default {
   browserLanguageDetection: true,
   serverLanguageDetection: true,
   ignoreRoutes: ['/_next', '/static'],
+  customDetectors: [],
   detection: {
     order: ['cookie', 'header', 'querystring'],
     caches: ['cookie'],


src/create-i18next-client.js
```diff
@@ -1,31 +1,34 @@
 import isNode from 'detect-node'
 import i18next from 'i18next'
 import i18nextXHRBackend from 'i18next-xhr-backend'
-import i18nextBrowserLanguageDetector from 'i18next-browser-languagedetector'
+import I18nextBrowserLanguageDetector from 'i18next-browser-languagedetector'
```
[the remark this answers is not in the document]
lucasfeliciano: Eslint got red because of this rule ( https://eslint.org/docs/rules/new-cap ).

lucasfeliciano:
I addressed your comments @isaachinman

src/create-i18next-client.js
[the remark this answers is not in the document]
isaachinman: But it doesn't error for the previous two lines? I want these `i18next` imports to all be of the same casing (camel case). We can change the eslint settings if necessary.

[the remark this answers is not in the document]
lucasfeliciano: No, because the previous line we are not instantiating with `new` .

And for the new `i18nextMiddleware` the class constructor is on `new i18nextMiddleware.LanguageDetector()` and `LanguageDetector` is capital so doesn't eslint doesn't complain.

Also, it is good practice to class names having upper case.

I saw that we use airbnb styleguide:
https://github.com/airbnb/javascript#naming--PascalCase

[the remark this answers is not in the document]
isaachinman: Ah right. I don't understand though - linting is [part of our CI process](https://github.com/isaachinman/next-i18next/blob/master/.circleci/config.yml#L52). Why is this coming up now, and how was the lint passing previously?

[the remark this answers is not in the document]
lucasfeliciano: Because previously we were not using the class constructor, we were doing like this:

```js
i18n.use(i18nextBrowserLanguageDetector)
```

and now we are instantiating the class:

```
 const browserDetectors = new I18nextBrowserLanguageDetector()
 config.customDetectors.forEach(detector => browserDetectors.addDetector(detector))
 i18n.use(browserDetectors)
```

isaachinman:
@lucasfeliciano Didn't cross my mind until now - do you mind adding some basic test coverage?

lucasfeliciano:
Yes. I will implement it today

lucasfeliciano:
@isaachinman I'm writing the tests and I noticed something weird

https://github.com/isaachinman/next-i18next/pull/109/files#diff-4703877688e364f2894a5be64978398fR10

`i18n.isInitialized` is always `undefined` are you aware of it?

capellini:
> `i18n.isInitialized` is always `undefined` are you aware of it?

@lucasfeliciano `i18n.isInitialized` is initially undefined:
screenshot from 2019-01-22 10-06-01

But when invoked on the client, it is true:
screenshot from 2019-01-22 10-05-01

lucasfeliciano:
@capellini 

But on the test context I don't see this behaviour

```js
  it('should initialize i18n', () => {
    const i18n = createI18nextClient({
      use: [],
      customDetectors: [],

    })

    expect(i18n.isInitialized).toEqual(true)
  })
```

`i18n.isInitialized` is `undefined` no matter what :(

capellini:
Oh, I see what you mean.  The tests were meant to be more like unit tests, so we would probably want to mock out the behavior of `i18next`.  So we would do something more like this:

```js
import i18next from 'i18next'
jest.mock('i18next', () => ({
  isInitialized: false
  // most likely other  things
}))

...

it('should not initialize i18n if i18n is already initialized', () => {
    i18next.isInitialized = true;

    // expect().to...
})

describe('initializing i18n', () => {
  beforeEach(() => {
    i18n.isInitialized = false
  })

  // more tests
})
```

lucasfeliciano:
So I don't see what's the value on testing this file.
Since mostly of the things are covered by `i18next` it self

kachkaev:
Given that we're always in the Next.js environment, should not all  language detection (including the custom one) be a part of the server middleware? Client-side language is always dependent on whatever language has already been picked up on the server, otherwise the content will flickr on page load in case of a mismatch.

Keeping all language detection logic in the server middleware will help developers keep proprietary detection logic out of the client-side bundle. If this logic spills over, some private data can leak too.

capellini:
> So I don't see what's the value on testing this file.

We want to make sure that the behavior of `create-i18next-client.js` does what it is supposed to do and remains intact through any refactor.  I like to think of it like double-entry bookkeeping -- it may seem like it's not providing any value, but it's an added step in the process to ensure that mistakes are not made (though, unit tests are not meant to catch integration bugs, which I think is what you mean by "I don't see the value").  And one usually doesn't appreciate it's value until the test fails post-refactor.

Some things I think that are worth testing in this file:
1. If `i18n` is initialized we do nothing and just return the `i18n` instance
2. If `i18n` is not initialized:
    1. We apply any middleware with `config.use`
    2. We call `i18n.init` to initialize `i18next`
    3. If we're server-side:
        1.  We use `i18nextNodeBackend`...

And so on.

That said, there is value to an integration test to ensure that `i18next` is also doing what we expect and catch integration bugs.  Keeping the tests separate, however, helps us isolate our bugs from theirs.  

If the original intent of the request was to write an integration test, my apologies.  My understanding was that we were calling for a unit test.

capellini:
@lucasfeliciano If we're after an integration test, `i18n.isInitialized` is never set because of [this setTimeout call](https://github.com/i18next/i18next/blob/62e04664e5db84955d39c8958a48ff493cdf7598/src/i18next.js#L118), which results in the test `expect` code running before [this changeLanguage callback is invoked](https://github.com/i18next/i18next/blob/62e04664e5db84955d39c8958a48ff493cdf7598/src/i18next.js#L106-L111).  Which means we may want to use a [timer mock](https://jestjs.io/docs/en/timer-mocks).

lucasfeliciano:
> Given that we're always in the Next.js environment, should not all language detection (including the custom one) be a part of the server middleware? Client-side language is always dependent on whatever language has already been picked up on the server, otherwise the content will flickr on page load in case of a mismatch.

 @kachkaev  At my company we use server and client detectors without any problem and the main reason for it is that we also serve cached js files through service workers, which means that we don't hit the server at all.

Other reason for client side detection is when you call `i18n.changeLanguage()` and the application change the url, fetch new translations without hitting the server to get another rendered html.

> If you have doubts about other testing approaches, do you mind opening a new issue so we can track it there?
 @isaachinman, after @capellini's comment I get what I was missing, gonna implement the tests.
I was just not seeing how you guys wanted me to approach this.

lucasfeliciano:
Add added some tests, I couldn't mock `detect-node` since it is not a function, so I didn't figure out how to change the value from test to test.

Let me know what you think

capellini:
> I couldn't mock `detect-node` since it is not a function

That was a problem for me, as well.  The only solution I could think of was to create separate files - a 'node' test and a 'non-node' test.

In the 'node' test file, I included:
```js
jest.mock('detect-node', () => true)
```
and in the non-node test file, I put:
```js
jest.mock('detect-node', () => false)
```

I've been meaning to ask the `detect-node` folks to see if they have a better way, but have yet to get around to it. But that's what I did for the tests that required it for the time being.

lucasfeliciano:
What about `process.browser`?
Isn't it enough?

capellini:
I usually avoid reaching into the implementation logic of external packages in tests, but from the `detect-node` source code, it looks like as long as `process` is of type `[[Class]] process`, `isNode === true`.  So you would need to replace the entire `global.process` with something else, or somehow trick the test into thinking that the existing `process` object is not of type `[[Class]] process`.

Still, you would have to do this before you import the test file and would need separate files, as once you import the file you are testing, the `detect-node` value won't change.

I think it's more explicit and clearer to use `jest.mock`.

__tests__/create-i18next-client.test.js
```diff
@@ -0,0 +1,71 @@
...
+      customDetectors: [],
+      serverLanguageDetection: true,
+      browserLanguageDetection: true,
+    })
+
+    expect(i18next.init).not.toBeCalled()
+    expect(i18next.use).not.toBeCalled()
+  })
+
+  it('should not initialize i18n if i18n is already initialized', () => {
```
isaachinman: Why do we have duplicate case declarations here? What are you trying to test for?

__tests__/create-i18next-client.test.js
```diff
@@ -0,0 +1,71 @@
...
+
+jest.mock('i18next-express-middleware', () => ({
+  LanguageDetector: jest.fn(),
+}))
+
+jest.mock('i18next-browser-languagedetector', () => jest.fn())
+
+describe('initializing i18n', () => {
+  beforeEach(() => {
+    i18next.isInitialized = undefined
```
isaachinman: Shouldn't this be `false`?

__tests__/create-i18next-client.test.js
```diff
@@ -0,0 +1,71 @@
...
+    i18next.isInitialized = false
+
+    createI18nextClient({
+      use: [],
+      customDetectors: [],
+      serverLanguageDetection: true,
+      browserLanguageDetection: true,
+    })
+
+    expect(i18next.init).toBeCalled()
```
isaachinman: This is going to be truthy regardless of how many times it's called.

__tests__/create-i18next-client.test.js
```diff
@@ -0,0 +1,71 @@
...
+      customDetectors: [],
+      serverLanguageDetection: true,
+    })
+
+    expect(i18nextMiddleware.LanguageDetector).toBeCalled()
+    expect(I18nextBrowserLanguageDetector).not.toBeCalled()
+    expect(i18next.use).toBeCalledTimes(2)
+    expect(i18next.init).toBeCalled()
+  })
+})
```
isaachinman: This test suite is a good start, but it doesn't actually test if `customDetectors` works. Was that something you had trouble with?

lucasfeliciano: Added test for that

__tests__/create-i18next-client.test.js
lucasfeliciano: Specified the number of calls

lucasfeliciano:
I added the remaining tests.
The only thing that I didn't cover was the client side part.

Would be an option do iterate over the tests in a different PR?
This is the only thing that I need to use `next-i18next` in production 🙏

__tests__/create-i18next-client.test.js
```diff
@@ -0,0 +1,110 @@
...
+  it('should call addDetector with the right detector if serverLanguageDetection is true', () => {
+    i18next.isInitialized = false
+    const addDetector = jest.fn()
+    i18nextMiddleware.LanguageDetector.mockImplementation(() => ({
+      addDetector,
+    }))
+
+    createI18nextClient({
+      use: [],
+      customDetectors: [3],
```
isaachinman: Can you please make this a unique string instead of a number? Eg `"test_custom_detector"`. With that change, I will merge this. Thanks.

lucasfeliciano: Nice one.
Force pushed the change

lucasfeliciano:
@isaachinman I addressed your last comments.
Thank you for the review and inputs, everyone.

isaachinman:
@lucasfeliciano Merged, thank you very much. I will release this imminently so that you can start using it in prod. Just keep in mind we're still in a pre-release stage, so you might want to pin to exact versions in your prod deployments just to be safe.

lucasfeliciano:
@isaachinman Good to hear.
I'll deploy with at our company on Monday :)
exciting :)

RonnyVo:
Hi @isaachinman I don't know where to ask so I put my comment here, how can I change the cookie name to detect the language instead of the default one: i18next? Please help, thanks.

isaachinman:
@RonnyVo That should be `config.detection.lookupCookie`. Let me know if that works for you - don't think anyone else has needed to do that yet.

RonnyVo:
@isaachinman it doesn't work, here is my code in i18n.js
`const NextI18Next = require('next-i18next').default

    defaultLanguage: 'vi',
    otherLanguages: ['en'],
    detection: {
        lookupCookie: "lang"
    }
})
`
I want to use other config from https://github.com/i18next/i18next-browser-languageDetector too but I don't know how to use it in next-i18next

lucasfeliciano:
Hello @RonnyVo,

You can use any of the detector's config as inside of the `detection` attribute just like you are doing.

Try to add the `caches: ['cookie']` to it

```

    defaultLanguage: 'vi',
    otherLanguages: ['en'],
    detection: {
        lookupCookie: "lang"
        caches: ['cookies']
    }
})
```

lucasfeliciano:
@isaachinman I think I know what is happening.

We are overwriting the detection config instead of merging it.
So the default values are being overwritten.

I'm back from my holidays, I'll take a look when I arrive home from work tonight

isaachinman:
@RonnyVo Open a new ticket to track this new request, please.

isaachinman:
Opened #179.

Issue #113, opened 2019-01-19
Passing configuration to React components


capellini:
This issue is a follow-up on the first point mentioned in [this comment](https://github.com/isaachinman/next-i18next/issues/105#issuecomment-455535621) in #105.

What do you think about passing `next-i18next` configuration to our react components by using a HoC?  This allows our React component files to be less cluttered.  As a side benefit, we also won't run into cyclomatic complexity problems as easily with CodeClimate because our components won't be wrapped into a function.  Overall, IMO it seems to be much more "React-like" to use a HoC, since we're using props to pass our config.

In practice, we would need to add a HoC (say `withConfig`) and then in `src/index.js` change the lines like this:
```js
    this.Link = Link.apply(this)
```
to this:
```js
    this.Link = withConfig(Link, config)
```
Changes would all be transparent to the user, so no breaking change.  To be clear. this does not address the second point mentioned in the comment.  Thoughts?

PR with an example forthcoming.

isaachinman:
Fixed via #114.

PR #114, opened 2019-01-19
Adding HoC to handle passing config to React components

capellini:
An example implementation of how we could use a higher-order component to pass configuration on to our React components that require configuration.

Ref #113.

### DIFF
--- a/src/hocs/with-config.js
+++ b/src/hocs/with-config.js
@@ -1,0 +1,26 @@
+import React from 'react'
+
+export default (WrappedComponent, config) => {
+  class WithConfig extends React.Component {
+    static getInitialProps(ctx) {
+      if (WrappedComponent.getInitialProps) {
+        return WrappedComponent.getInitialProps(ctx)
+      }
+
+      return {}
+    }
+
+    render() {
+      return (
+        <WrappedComponent
+          {...this.props}
+          nextI18NextConfig={config}
+        />
+      )
+    }
+  }
+
+  WithConfig.displayName = `withNextI18NextConfig(${WrappedComponent.displayName || WrappedComponent.name || 'Component'})`
+
+  return WithConfig
+}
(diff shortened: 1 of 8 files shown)


capellini:
A few notes:
- I opted to wrap the `Trans` component in `withNamespaces` like we do with the `Link` component, since `withNamespaces` exposes an `i18n` prop.  I thought that adding another `i18n` prop to `withConfig` would be redundant in this case.  This removed the need to add configuration to the `Trans` component entirely.
- I also "snuck" in adding `propTypes` and `defaultProps` for the `lng` prop inside of the `Link` component into this PR.  As it is not really related to the HoC, If you would prefer that I submit a separate PR please let me know.

src/hocs/with-config.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+
+export default (WrappedComponent, config) => {
+  class WithConfig extends React.Component {
+    static getInitialProps(ctx) {
+      if (WrappedComponent.getInitialProps) {
```
isaachinman: Why would a wrapped component ever have a `getInitialProps`? We just need to add the config object into the props, nothing more.

capellini: I usually include these lines in my HoCs as a habit, as it avoids the headache of adding it later when I inevitably use it with a wrapped component that has a `getInitialProps` method.  But if we can guarantee that this will never be used with a wrapped component that has a `getIntialProps` method, then I suppose I'll throw caution to the wind and just remove the lines.

capellini:
> We just need to add the config object into the props, nothing more.

`getInitialProps` removed - 0ebed46febf645ddbe560e51735b796b3cd74cfc

> I'd rather this PR be a 1:1 replacement for the current solution instead of introducing new approaches.

Done - 72f75965b0e764e876e578abf21b36bd8a30a22c

> `propTypes` and `defaultProps` should be in a separate PR and would be merged immediately. 

Removed `lng` `propTypes` and `defaultProps` - 0066348bed138c2f5a831ba200491e76e7bdcf3e

> In fact, we should probably change the `eslint` config to require them.

Agreed.

Issue #120, opened 2019-01-22
Language selected in client is not being used on next server re-render


sudhirt4:
After changing the client side language using ``i18next.changeLanguage``method (or directly manipulating the ``i18next`` cookie value) and then refreshing the page the language selected is not being used by the SSR.

This used to work before, Facing this issue after upgrading from ``0.16.0`` to ``0.20.0``

isaachinman:
@capellini Any idea where this regression could be located? I assume it's within `app-with-translation`. Also, we'll need an e2e test to cover this once a bug fix is in place to prevent future regression.

isaachinman:
@sudhirt4 Very sorry for the trouble here, please bear with us while we debug this - it has the highest priority at the moment.

isaachinman:
@sudhirt4 Update - I cannot reproduce this via the [example](https://github.com/isaachinman/next-i18next/tree/master/example). If I load into an incognito window (served English), change the language to German on the client side, and do a hard reload, I get served an SSR page in German.

Please provide a reproducible repo/example for us to debug. It's either a bug in more complex use cases, or user error.

isaachinman:
Are you using the `i18n` instance that `next-i18next` exports to change the language, [as seen here](https://github.com/isaachinman/next-i18next/blob/master/example/pages/index.js#L18)?

capellini:
> I cannot reproduce this

@isaachinman - neither can I.  Both `changeLanguage` and changing the cookie work for me.

> we'll need an e2e test to cover this once a bug fix is in place to prevent future regression.

:+1: agreed

@sudhirt4 Would you mind installing v0.17, v0.18, and v0.19 and tell me what specific major version causes the code to break?

ngocketit:
I got the same problem. Calling `i18n.changeLanguage()` on client updates the `i18next` cookie to the correct langcode but then it's switched back with the default value when the page is refreshed.

ngocketit:
But I couldn't reproduce with the example either, weird!

ngocketit:
Alright, I can confirm that it works when NODE_ENV is set to production but not with development. Any ideas?

capellini:
I just came across one scenario (still looking for more) where this could happen with [this change](https://github.com/isaachinman/next-i18next/commit/e7ae318a6df47d2d8592ca61d9b140cf36bf60c2#diff-629eb664116efd2a55d4c6361f00c9bd) in `lngPathDetector`.

Scenario assumes:
[...]

PR #121, opened 2019-01-22
Implement lng-from-req to use req.i18n.language

sudhirt4:
Not exactly sure on the intention of ``allLanguages.includes(l)`` in this method (lng-from-req) but I am assuming [i18next-express-middleware](https://github.com/i18next/i18next-express-middleware) already provides the current language selected in req.i18n object. Ref [here](https://github.com/i18next/i18next-express-middleware/blob/master/src/index.js#L43)

Using these changes in PR solved this issue (Ref:  #120) for me without breaking other feature I have currently implemented.

### DIFF
--- a/src/utils/lng-from-req.js
+++ b/src/utils/lng-from-req.js
@@ -1,11 +1,9 @@
 export default (req) => {
-
   const { allLanguages, defaultLanguage, fallbackLng } = req.i18n.options
 
-  const language = req.i18n.languages.find(l => allLanguages.includes(l))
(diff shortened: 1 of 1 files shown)


isaachinman:
Sorry, I don't understand this PR. The `find` method expects you to return a bool. I'll look into #120 as a high priority bug, but this is not going to be the fix.

sudhirt4:
@isaachinman : It seems the ``find`` method whitelisting the current language. It seems to be expecting a language string, I don't understand what you mean by it is expecting a boolean.

PR #130, opened 2019-01-28
Consolidate create-config.js unit tests into one file

### LINKED ISSUE #128: Patch detect-node until there is a more testable version?
Currently, we're using `detect-node` by importing it straightaway:
```js
import isNode from 'detect-node'
// ...
if (isNode) { //...
``` 
Since `detect-node` exports only a boolean, this requires is to create two unit test files for every file that uses `detect-node`. which is less than ideal.  I have an issue in with detect-node (iliakan/detect-node#13) to see if there is a better way to do a unit test, or if they want to make a more testable version, but have not heard back yet.

In the meantime (and in case they don't agree), we can make our own `detect-node` by doing something like the following in a separate file:
```js
import isNode from `detect-node`

export { isNode }
```
And then we can use it as a named import, like so:
```js
import { isNode } from 'util/detect-node'
// ...
if (isNode) { // ...
```
The only thing that changes is the import, from a default to a named import.  But now we can test like so:
```js
import detectNode from 'util/detect-node'
jest.mock('util/detect-node', () => ({
  isNode: true
}))
```
and then just change the value of `detectNode.isNode` as required in our tests, allowing us to combine tests into one.

To me, it's a very small change that would allow us to greatly improve the readability of our tests.  What do you think?  If you agree, I'll create a PR for it.

capellini:
Utilize `jest.resetModules` to import a fresh copy of `create-config.js` for each test.  This helps us import a fresh boolean value for the `detect-node` package, depending on whether or not we wanted to test server-side or client-side logic, without the need for two separate test files.

Fixes #128.


isaachinman:
This looks fine. Can you rename `importCreateConfig` to something like `mockIsNodeCreateConfig`?

PR #132, opened 2019-01-28
$npm_execpath added, i18next to version 14.0.1

andylacko:
there is only 1 problem, husky fails to run :/ as if it doesn't recognize $npm_execpath

also, if you want this pull request, please check the version 14.0.1 of i18next

I don't have much time today, this is suggestion more than request

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # Dependencies
 node_modules
 
+
 # Dist
 dist
 .next
@@ -17,3 +18,7 @@ out
 
 # Testing
 coverage/
+
+
+.idea/
+package-lock.json
--- a/package.json
+++ b/package.json
@@ -24,9 +24,9 @@
     "lint": "eslint src examples __tests__",
     "lint:fix": "eslint src examples __tests__ --fix",
     "build": "babel src --out-dir dist",
-    "prepare": "yarn build",
-    "run-example": "yarn build && cd examples/simple && rm -rf node_modules && yarn && yarn dev",
-    "run-example:prod": "yarn build && cd examples/simple && rm -rf node_modules && yarn && yarn build && yarn start",
+    "prepare": "$npm_execpath run build",
+    "run-example": "$npm_execpath build && cd examples/simple && rm -rf node_modules && $npm_execpath && $npm_execpath dev",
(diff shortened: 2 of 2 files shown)


isaachinman:
Can you please explain what the purpose of this PR is?

lucasfeliciano:
@isaachinman I think it is because this project uses `yarn` to run scripts. 

And the `prepare` script runs even after `npm install`

Which means that who installs it as a dependency also needs to have `yarn` installed.

I would recommend to change all scripts on `package.json` to use npm since this is an open source library.

lucasfeliciano:
@isaachinman I'm not worried about the maintainers, the point is use `npm` on scripts because of the users because the `prepare` script runs after you install `next-i18next` into any project.

So after `npm install next-i18next` it will run the prepare script with `yarn` but if the user doesn't have it it will fail.

isaachinman:
This package doesn't need a `prepare` script for users to run, only a `prepublish` script. Last I checked they were deprecating certain scripts and it was a complete mess. The proper fix here would be to not have the end user run anything.

@lucasfeliciano Do you want to look into that?

lucasfeliciano:
I can take a look at the end of the week

isaachinman:
Basically, I think we need to change `prepare` to `prepublishOnly`.

I'm going to close this PR as we never got a response from @andylacko.

I've created #142 to track the npm scripts issue.

andylacko:
ok, sure, sry, I was working the whole day, just wanted to help with npm/yarn 👌if it is only for developers, everybody can install yarn

isaachinman:
No worries, I've already fixed #142 and will be publishing it soon.

andylacko:
@isaachinman , btw nice job with that custom subpath, it works well.

I was wondering, if something like next/router middleware is included in bundle, or should I create it myself. Or meybe if it's on schedule, because I think it would be better, if I used library functions which are tested rather my own code. If it's not, I will create it for the project and afterwards I can give you the code, but don't know if I have time to study this repo and implement it proper way into lib

isaachinman:
Hi @andylacko, I'm not sure I understand your question. Can you please open a new issue, clearly stating your question or request so that we can track the conversation separately?

PR #139, opened 2019-01-29
Properly set fallbackLng

### LINKED ISSUE #134: setting defaultLanguage to something other than 'en' breaks page after Update to 0.23.0
Hi, after updating to Version 0.23.0 I had to disable `serverLanguageDetection`. If it is enabled, you get the following error when visiting the page:

`TypeError: this.language.toLowerCase is not a function
    at I18n.loadResources (xxx/node_modules/i18next/dist/commonjs/i18next.js:209:42)`

capellini:
Currently, in a production environment, if a user specifies a defaultLanguage other than 'en' and does not specify 'en' in otherLanguages, our default configuration sets fallbackLng === 'en'.

This change properly sets the fallbackLng to the user-specified defaultLanguage, if the user hasn't already specified a fallbackLng.

Fixes #134.


capellini:
This fixes the problem with our simple example, but I'm waiting to hear back from @johnrackles to see if more is needed.

Issue #137, opened 2019-01-29
Get 'language is not included' error on first time page request


vimutti77:
When I request page at first time. I get `Invalid configuration: Current language is not included in all languages array` in the browser console.

Also, If I remove `i18next` cookie and then refresh the page. The error message is occurred too.

This error message is only occurred when I enable `localeSubpaths`.

isaachinman:
@capellini I won't have time to look into this until this evening, but this could be related to the recent middleware changes as well. This was something I explicitly fixed in the past.

capellini:
@vimutti77 Could you please post your `i18n.js` file?

vimutti77:
This is my `i18n.js` file.
```
const NextI18Next = require('next-i18next')

const options = {
  localeSubpaths: true,
  otherLanguages: ['th'],
  localePath: 'public/locales',
}
module.exports = new NextI18Next(options)
```

After I did some test. I found that sometime the error occurred and sometime not.

capellini:
Thank you.  Also, what languages does your app support? 'en' and 'th'?

capellini:
I think this issue my be complicated by the fact that we're serving static files outside of the `static` directory, which Next expressly forbids, see this note in [the docs](https://github.com/zeit/next.js/#static-file-serving-eg-images):

> Note: Don't name the static directory anything else. The name is required and is the only directory that Next.js uses for serving static assets.

That said, in reality we often do need to serve static files outside of `static`. I was able to get our 'simple' example that used files outside of `static` working with the following configuration (and one bug fix):

__`server.js`__
```js
const express = require('express')
const next = require('next')
const nextI18NextMiddleware = require('next-i18next/middleware')

const nextI18next = require('./i18n')

const app = next({ dev: process.env.NODE_ENV !== 'production' })
const handle = app.getRequestHandler();

(async () => {
  await app.prepare()
  const server = express()

  server.use(nextI18NextMiddleware(nextI18next))

  // use '/public' for static files
  server.use('/public', express.static('public'))

  server.get('*', (req, res) => handle(req, res))

  await server.listen(3000)
  console.log('> Ready on http://localhost:3000') // eslint-disable-line no-console
})()
```

__`i18n.js`__
```js
const NextI18Next = require('next-i18next')

module.exports = new NextI18Next({
  ignoreRoutes: ['/_next', '/static', '/public'], // note that we have to redefine ignoreRoutes
  otherLanguages: ['de'],
  localeSubpaths: true,
[...]

PR #140, opened 2019-01-29
Ensure client is properly configured with custom localePath

capellini:
Currently, if the user defines their own loadPath and/or localeStructure, the server-side is properly configured, but the client-side does not change config.backend to reflect this change.

This change properly configures config.backend on the client to account for any changes in custom loadPath or localeStructure.

Related to #137.

### DIFF
--- a/__tests__/config/create-config.test.js
+++ b/__tests__/config/create-config.test.js
@@ -91,6 +91,34 @@ describe('create configuration in non-production environment', () => {
       expect(config.backend.loadPath).toEqual('/home/user/static/translations/{{ns}}/{{lng}}.json')
       expect(config.backend.addPath).toEqual('/home/user/static/translations/{{ns}}/{{lng}}.missing.json')
     })
+
+    it('does not overwrite loadPath if user has specified one', () => {
+      const userConfigWithLoadPath = {
+        ...userConfig,
+        backend: {
+          loadPath: '/static/custom/path/{{ns}}/{{lng}}.json',
+        },
+      }
+
+      const config = createConfig(userConfigWithLoadPath)
+
+      expect(config.backend.loadPath)
+        .toEqual('/home/user/static/custom/path/{{ns}}/{{lng}}.json')
+    })
+
+    it('does not overwrite addPath if user has specified one', () => {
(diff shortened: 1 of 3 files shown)


isaachinman:
On hold due to ongoing discussion in #137.

capellini:
@isaachinman What do you think about the approach in c0f2a7a?  I think that it handles these the three use cases:
1. if zero config, then loadPath and addPath are defined by us
2. if user specifies a backend without a loadPath (or addPath) we define it for them
3. if user specifies a backend with a loadPath (or addPath), it remains untouched

p.s. I think the CodeClimate errors are a false positive, but let me know if you'd like me to resolve them and I'll think of something.

isaachinman:
This looks like a lot of responsibility for us as maintainers to assume. I'd rather this responsibility lie with the user to properly define their paths and backend if they do not want to use the default options.

I don't see a lot of value added by abstracting this logic away from users. Instead, I see it producing a lot of future problems/confusion when settings are not what people expect them to be.

I prefer the approach of: "it will work out of the box if you accept our paradigm - if not, you will need to get your hands dirty".

isaachinman:
Definitely open to discuss this more, but currently I'm seeing it as a won't-fix.

capellini:
Ok.  To be clear, do we just want to close this, or should I change the logic to zero config or custom config?  If the latter, what should be the demarcation between zero config and custom config.  For example, users will have to define `otherLanguages`.  Is any configuration that has more than that considered a custom configuration?

isaachinman:
In an ideal world, users just need to tell us what their default language is, and then what all the other languages are. No other config options are needed. This is what we will refer to as "zero config".

Zero config is already supported for backend/locale paths. What's necessary here might actually be removing some of the logic if users pass paths, so that all related options will need to be specified manually.

capellini:
> What's necessary here might actually be removing some of the logic if users pass paths, so that all related options will need to be specified manually.

Right, so move all the `backend` stuff to the `default-config.js`, right?

isaachinman:
Yes I believe that's correct. Mind putting together a PR?

capellini:
I don't mind at all.

Issue #135, opened 2019-01-28
Fix concurrency issues due to i18n singleton


lucasfeliciano:
We have a problem that is around since the examples of `with-i18next` on next.js repository which we should try to fix because is really bad issue.

The problem is that we have one instance of `i18next` and we pass it to the provider 

```js
<I18nextProvider
  i18n={i18n}
  initialLanguage={initialLanguage}
  initialI18nStore={initialI18nStore}
>
```

What happens here is that if a request hits the server and change the language before next renders the html it will render with the wrong language.

I solved the problem by cloning the `i18next` instance before passing to the provider

```js
...
render() {
  const clone = i18n.cloneInstance({
    initImmediate: false,
    lng: initialLanguage,
  })

  return (
    ...
    <I18nextProvider
      i18n={clone}
      initialLanguage={initialLanguage}
      initialI18nStore={initialI18nStore}
    >
    ...
  )
}
```

This also happens if you use the `i18n` instance outside of react context ( and this problem I couldn't think how to solve it yet ).
For example:

I have a file where I setup our `apollo-client` instance, and there I have some middlewares which inject some headers like this:

```js
import i18n from '../i18n' // <--- the file which exports the i18next instance
...

// Accept language link
const acceptLanguageLink = new ApolloLink((operation, forward) => {
  operation.setContext(({ headers = {} }) => {
    return {
      headers: {
        ...headers,
        'Accept-Language': i18n.language, // <--- Prone to concurrency issue
      },
    }
  })

  return forward(operation)
})
```

This issue is much more complex than looks like and I would love to think in a solution with you guys however I have some questions about the `next-18next` setup which @capellini or @isaachinman could give me some insights.

I'll start with: 
1. What is the context of `this` here: https://github.com/isaachinman/next-i18next/blob/master/src/hocs/app-with-translation.js#L12. I'm asking to know where the `i18n` comes from.

isaachinman:
This is an interesting problem indeed.

To answer your first question: `this` at that line refers to the `NextI18Next` class instance. We are basically creating a HOC around the `app-with-translation` and [binding the entire `NextI18Next` class instance to it](https://github.com/isaachinman/next-i18next/blob/master/src/index.js#L15), to be able to access things like the `i18n` instance, and config options.

There is almost certainly a better way to do things, but this is just what I came up with as a first iteration.

[...]

PR #141, opened 2019-01-29
Serialize i18n instance and pass it through getInitialProps

lucasfeliciano:
This PR fixes the issue that I described on issue #135.

I took a different approach because by cloning the instance I introduced some memory leak to my application to a point that the kubernetes pod were dying.

So the solution on the issue is not reliable, instead I'm making next.js serialize the server instance and using it if it is available.

I'll deploy this change tomorrow at work and keep an eye in our server's resources to see how it behave.

But it fix the problem.

For now I'm opening a PR to start a discussion and fix any comment if needed

### DIFF
--- a/src/hocs/app-with-translation.js
+++ b/src/hocs/app-with-translation.js
@@ -47,14 +47,12 @@ export default function (WrappedComponent) {
       const { req } = ctx.ctx
       let initialI18nStore = {}
       let initialLanguage = null
+      let i18nServerInstance = null
 
       // Step 1: Determine initial language
       if (req && req.i18n) {
 
         initialLanguage = lngFromReq(req)
-
-        // Perform a lang change in case we're not on the right lang
-        await i18n.changeLanguage(initialLanguage)
 
       } else if (Array.isArray(i18n.languages) && i18n.languages.length > 0) {
         initialLanguage = i18n.language
@@ -109,26 +107,29 @@ export default function (WrappedComponent) {
             .map(ns => new Promise(resolve => i18n.loadNamespaces(ns, () => resolve()))),
         )
         initialI18nStore = i18n.store.data
+      }
 
+      // Step 4: Overwrite i18n.toJSON method to be able to serialize the instance
+      if (req && req.i18n) {
+        req.i18n.toJSON = () => null
+        i18nServerInstance = req.i18n
       }
 
       // `pageProps` will get serialized automatically by NextJs
       return {
         initialI18nStore,
         initialLanguage,
+        i18nServerInstance,
         ...wrappedComponentProps,
       }
     }
 
     render() {
-      let { initialLanguage, initialI18nStore } = this.props
-      if (!process.browser) {
-        initialLanguage = i18n.language
-        initialI18nStore = i18n.store.data
-      }
+      const { initialLanguage, initialI18nStore, i18nServerInstance } = this.props
+
       return (
         <I18nextProvider
-          i18n={i18n}
+          i18n={i18nServerInstance || i18n}
           initialLanguage={initialLanguage}
           initialI18nStore={initialI18nStore}
         >
--- a/src/hocs/app-with-translation.js
+++ b/src/hocs/app-with-translation.js
@@ -54,6 +54,9 @@ export default function (WrappedComponent) {
 
         initialLanguage = lngFromReq(req)
 
+        // Perform a lang change in case we're not on the right lang
+        await i18n.changeLanguage(initialLanguage)
+
       } else if (Array.isArray(i18n.languages) && i18n.languages.length > 0) {
         initialLanguage = i18n.language
       }


isaachinman:
Hi @lucasfeliciano. If you don't mind, I'd rather we take some time on this issue of concurrency and come up with something as robust as possible, instead of merging half-fixes. Is that OK with you? Would it be too much to ask for you to maintain a fork for your professional work, and perhaps experiment with some new ideas there? I also need to look into this, but general maintenance of the repo is taking a lot of my time at the moment.

lucasfeliciano:
> If you don't mind, I'd rather we take some time on this issue of concurrency and come up with something as robust as possible, instead of merging half-fixes. 

Of course, this was more like a proof of concept and to start a discussion with some code on it. Let's just not forget about this issue because it is really important and seeing how this repository is starting to trend people will start to use it in production at some point and have these issues.

> Would it be too much to ask for you to maintain a fork for your professional work, and perhaps experiment with some new ideas there? 

I'm doing that already but I'm happy to share my findings and show some real world issues while I bring new ideas and help implementing the solution for them.

> I also need to look into this, but general maintenance of the repo is taking a lot of my time at the moment.

I can imagine, this is starting to get some momentum so more work is involved. I would love to help you maintaining it if you need help. Believe me, I think it is a mutual interest to get this library in a pristine state. It is already helping me a lot and I see potential to help a lot of people.

isaachinman:
I would actually like to get this merged ASAP. I feel a bit bad about @jamuhl, he tried to remind me of this issue a couple times and it still slipped through the cracks.

@lucasfeliciano You already are helping me, a lot!

My biggest goal right now is not to add any new features or extend the API, but instead simply get to a stable state.

Can you explain this:

> Overwrite i18n.toJSON method to be able to serialize the instance

I do remember something here about circularity within the JSON structure. Are you doing this to avoid that problem? Have you tested this specific PR against your real-world applications?

isaachinman:
> First I was trying to implement this without this "hack" posted something on tweeter and @jamuhl replied telling me about this and worked.

Okay, good to hear. I trust @jamuhl implicitly. Funny, I also heard from @timneutkens that you guys were talking. You're getting to know the two contact points between the packages. Good to see.

I will take a second look at this tomorrow, hopefully we can merge it soon. I have a feeling that even if this is not perfect, it's much better than what we're currently running with.

> Do you have what is needed to reach the milestone?

Yes, I set up a [1.0 milestone](https://github.com/isaachinman/next-i18next/milestone/2). It's somewhat of a moving target and I have a feeling TypeScript is going to get taken out of there.

lucasfeliciano:
I'm at DDD Europe today and tomorrow, I don't know if I'll be time to go back at this until the weekend.
But hold deploy until I check one more thing in production

lucasfeliciano:
False alarm, it was my mistake.

We need this, otherwise, the imported instance has language `null`
https://github.com/isaachinman/next-i18next/pull/141/files#diff-2a4d2aa19448a9cd757d2ce4e50beb39L57

isaachinman:
> False alarm, it was my mistake

Sorry, can you explain what you mean?

I see you've removed the `await i18n.changeLanguage(initialLanguage)` call, and removed the `if (!process.browser)` clause.

lucasfeliciano:
I'll explain later. I have to run to my talk.
I just had I insight during lunch and wrote it real quick.

I'll elaborate later

isaachinman:
@lucasfeliciano I wasn't familiar with the `toJSON` hack, but it makes perfect sense. I'm very familiar with the rest of this approach as it's similar to something that existed in the past example within `react-i18next`. 

It's correct that you should not remove the `await i18n.changeLanguage` call - that is there for a reason

All we need to add is serialisation, nothing else. It looks like this PR has been simplified to that now.

Hope your talk went well today. I am ready to merge this. Can you please give the OK so that we can proceed?

lucasfeliciano:
Sorry for the delay and confusion earlier.

Yes this PR is ready to be merged. :)

westpole:
Hi @lucasfeliciano you might the one who could answer a question "how would you get a current language on server side?" as you made changes.
Unfortunately, i18nServerInstance  does not contain language property as browser instance (i18n.language).

The issue is that I would not be able to set active class for a language link on server, therefor I will get an error message that "Prop `className` did not match".

Should we enhance i18nServerInstance object to keep consistence with browser object?

isaachinman:
@westpole You can use `initialLanguage`. Or you can use [lng-from-req](https://github.com/isaachinman/next-i18next/blob/master/src/utils/lng-from-req.js).

westpole:
@isaachinman thanks.
I should've investigate `props` object more. As turned out Component wrapped with `withNamespaces` should contain `lng` property.

isaachinman:
Yes that's correct, the `lng` prop from `withNamespces` will actually be fed from `initialLanguage` via context.

lucasfeliciano:
@westpole On server you could use `req.language` since `i18next-express-middleware` set the language for each request.

`withNamespaces` works as @isaachinman mentioned if you are on SSR context. :)

PR #149, opened 2019-02-01
feat: Export Router for Imperative Routing

minocys:
### CHANGES
1. Wrap NextRouter push, replace, and prefetch functions as pushRoute, replaceRoute, and prefetchRoute using logic from Link.js to handle locale subpath routing
1. Move util functions from Link.js into utils
1. Add tests for wrap-router.js
1. Update README

### REASONS
I needed imperative navigation in my current project. Having read through issue #4, I thought that exporting a wrapped router using locale subpath logic taken from Link.js would be a simple stop gap till a more robust solution and/or the withI18nRouter HOC is implemented. 

This was heavily inspired by [next-routes](https://github.com/fridays/next-routes).

### CATCHES
I have not tested prefetchRoute in action, and am currently leaning towards removing it from this PR since A) I am unsure if it will work properly with next.js, and B) the main pain point I wanted to solve was imperative routing


isaachinman:
@minocys In the future it would be good to communicate with us before investing a lot of time and effort on serious PRs like this.

@capellini Do you have time to review this in the near future?

I looked through it briefly, and although it might work it doesn't seem like the approach I'd take.

isaachinman:
While we wait for @capellini to respond, let me clarify: all the functionality we would need is already in the `i18n.on('languageChanged')` event, [which can be seen here](https://github.com/isaachinman/next-i18next/blob/master/src/hocs/app-with-translation.js#L20).

capellini:
@isaachinman I'll try to fit in a proper review this weekend.  A few things I noticed at first glace:
1. As you've mentioned, `lngPathCorrector` has the functionality that we need already built in.  In fact, we should incorporate it into our `Link` component.  Should we do that in a separate PR and rebase it into this one?
2. So far, we've aimed for a 1:1 replacement with Next's interface.  I think our router should have `push`, `replace`, and `prefetch` methods like Next router does, rather than having methods that are named differently.
3. I don't like that we're modifying Next's router.  I think we should create our own router with a next router instance within it.
4. It has suddenly occurred to me while looking over this PR that we do not incorporate `ignoreRoutes` when re-routing to locale-specific URLs in our `appWithTranslation` or `Link` components.  I will try to produce a repo that replicates this and submit a separate issue/PR.
5. I haven't had the time to look to see if it would be overwhelming to do so, but it would be nice to also incorporate a `withRouter` HoC (though perhaps that could wait for a different PR).

isaachinman:
`withRouter` and a wrapped `Router` are two separate concerns and are both on the table, although the working assumption is that `withRouter` will take more time.

Although I think this PR has the right idea, we might have to scrap it and start over.

isaachinman:
> In fact, we should incorporate it into our Link component. Should we do that in a separate PR and rebase it into this one?

That should be a separate PR entirely, no need for rebasing.

isaachinman:
> So far, we've aimed for a 1:1 replacement with Next's interface

Yes, everything will be a 1:1. There will be no new API.

isaachinman:
Your use of `lngPathCorrector` is incorrect. This implementation will send me to locale subpaths even when they are not enabled.

minocys:
@isaachinman Wow silly mistake on my part, my bad.

Updates:
1. Rebased against latest master branch
1. Moved localeSubpathRequired into utils
1. Only call lngPathCorrector after localeSubpathRequired check
1. Changed variable names to match the rest of the library more
1. Updated tests to check locale subpaths aren't being used when disabled

isaachinman:
@minocys Please squash these commits into one or two with sensible, meaningful commit messages.

minocys:
@isaachinman done.

minocys:
@isaachinman @capellini I have rebased this PR against the latest changes.

Is there something I'm not aware of that's blocking this feature?

Not trying to rush you guys, I'm just wondering if there is any work I can do to help move this forward.

isaachinman:
Thanks for rebasing. This looks good to me. @capellini Any chance you'll have time to review this soon?

capellini:
@isaachinman I'm tied up this morning, but I will make time to review it this afternoon/evening.

README.md
```diff
@@ -147,6 +147,23 @@ class SomeLink extends React.Component {
...
+We can also navigate imperatively with locale subpaths by importing `Router` from your `NextI18Next`.
+The exported Router shares the same API as the native Next Router. The push, replace, and prefetch functions will automatically prepend locale subpaths.
+
+```jsx
+import { Router } from '../i18n'
+
+function SomeButton() {
+  return (
+    <button
+      onClick={() => Router.pushRoute('/some-page')} // Navigates to /fr/some-page if users language is French
```
capellini:
```suggestion
      onClick={() => Router.push('/some-page')} // Navigates to /fr/some-page if users language is French
```

__tests__/router/wrap-router.test.js
```diff
@@ -0,0 +1,202 @@
...
+      }, `/de${href}`, options)
+
+      router.replace(hrefObj, undefined, options)
+      expect(NextRouter.replace).toHaveBeenNthCalledWith(4, {
+        pathname: href,
+        query: { test: 'something', lng: 'de' },
+      }, `/de${href}?test=something`, options)
+    })
+
+    it('calls NextRouter.prefetch with locale subpath prepended', () => {
```
capellini: `Router.prefetch` only takes a single string `url` argument.

src/router/wrap-router.js
```diff
@@ -0,0 +1,53 @@
...
+    Router[method] = (path, as, options) => {
+      const { config, i18n } = nextI18NextConfig
+
+      if (localeSubpathRequired(nextI18NextConfig, i18n.languages[0])) {
+        const href = parseHref(path)
+        const { pathname, query } = href
+        const asPath = as || format(href, { unicode: true })
+        const [correctedAs, correctedQuery] = lngPathCorrector(config, i18n, { asPath, query })
+
+        return NextRouter[method]({ pathname, query: correctedQuery }, correctedAs, options)
```
capellini: We need to do something different for `prefetch` as it expects only a single string argument and will fail if we pass in an object.

isaachinman: As far as I know, `prefetch` is a now-only feature, right? I _think_ some people have deployed `next-i18next` via `now`, but I'm not 100% sure. I don't know if we're going to be able to support it, and we might want to just leave that out.

capellini:
@isaachinman Review complete.  I had a few other suggestions for possible follow-on PRs:

1. Adding a withRouter HoC that mirror's Next's.
2. IMO, we should intercept [the events emitted by Next's router](https://github.com/zeit/next.js/tree/canary/packages/next#router-events) and apply an inverse `lngPathCorrector` function to the `url` before it's passed on to any event handlers.  In other words:
```js
const handleRouteChange = url => {
  console.log('App is changing to: ', url)
}
```
Will currently print the locale-specific URL, whereas I think it should print the locale-agnostic URL.  We should probably also pass other information into the event, such as the language, but I think it should be separate from the URL, as it would allow people to handle the URL without the locale subpath more cleanly.  Off the top of my head, I'm thinking of a signature like this:
```js
eventHandler(url: string, language: string)
```
An example event handler can look like this:
```js
const handleRouteChange = (url, language) => {
  console.log('App is changing to: ', url, 'with language', language)
}
```
This would allow a user to handle URL changes in the same way as they would with Next, if they wanted, or to also handle language-only changes, if they so chose.

At any rate, the details can be hashed out in a separate issue/PR, but I wanted to point out that it may be something we want to look at in the future.  I think the changes in this PR are a good start and will cover a large number of `Router` use cases.

src/router/wrap-router.js
capellini: We can support `prefetch` and it can be used in regular Next apps.  We just have to
```js
url.format({ pathname, query })
```
before we pass it in as the first argument.

minocys: @isaachinman @capellini Thank you for your review and comments.

I just spent some time thinking through prefetch, and I think we can't/shouldn't do anything to it at all.

Here are the scenarios I've walked through: 
1. For applications that do not require locale subpath, the given path shouldn't be touched. At the moment they would be using Next.js' router anyway. 

2. For applications that require locale subpath
    My initial reaction after reading your comments was that we could either:

    1. Add the current language to the path as a query variable
    1. Prepend the current language to the path for NextI18nextMiddleware to handle

    However, my understanding of prefetch is that it only downloads code, so therefore the current language is irrelevant, since Next.js would be expecting the base path.

    The only use case I could think of for wrapping prefetch is to also get the required namespaces  for that page, and my gut feeling is that it's beyond the scope of this library or a separate advanced feature/PR.

    Here is some quick pseudo code on how users might achieve this after looking at the [with-data-prefetch example](https://github.com/zeit/next.js/blob/3746d7d90b0dedc699cbd5f3ffdbeab2c68f192a/examples/with-data-prefetch/components/link.js#L8)
```
// Component with i18n
componentDidMount() {
  this.prefetch()
}

async prefetch() {
  const Component = await Router.prefetch('/some-route')

  // Not sure where ctx would come from since this is a prefetch
  const { namespacesRequired } = Component.getInitialProps(ctx)
  const { i18n } = this.props;

  await Promise.all(
      namespacesRequired
      .filter(ns => !i18n.hasResourceBundle(i18n.languages[0], ns))
      .map(ns => new Promise(resolve => i18n.loadNamespaces(ns, () => resolve())))
}
```

For now I have opted to just call prefetch without further action. What are your thoughts?

minocys:
Requested changes to README and jest.config.js have been made, I will squash/clean up my commits once the [issues with prefetch](https://github.com/isaachinman/next-i18next/pull/149#discussion_r255465219) have been resolved.

src/router/wrap-router.js
capellini: @minocys Thank you for that explanation.  I can see now that we don't need to do anything with `prefetch` and can use Next's implementation.  This package is only appending `lng` to the query parameters, which has no effect on `prefetch` (i.e. `Router.prefetch('/page?lng=en')` will fetch the same thing as `Router.prefetch('/page?lng=zh')` - it just fetches the code for `/page`).  We can leave it alone.

isaachinman:
@minocys Should be good to go.

minocys:
@isaachinman Awesome, commits have been squashed.

isaachinman:
@minocys Thanks.

@capellini Thank you for your review.

This will most likely get merged this evening.

isaachinman:
Thanks @minocys. I'll add a credit in the README for your contribution.

PR #150, opened 2019-02-01
Utilize lngPathCorrector within Link component

capellini:
Ref [this comment](https://github.com/isaachinman/next-i18next/pull/149#issuecomment-459787337) in #149.

### DIFF
--- a/__tests__/components/Link.test.js
+++ b/__tests__/components/Link.test.js
@@ -16,11 +16,14 @@ describe('Link component', () => {
   beforeEach(() => {
     props = {
       href: '/foo/bar',
-      lng: 'de',
       nextI18NextConfig: {
         config: {
+          allLanguages: ['en', 'de'],
           defaultLanguage: 'en',
           localeSubpaths: false,
+        },
+        i18n: {
+          languages: ['de', 'en'],
         },
       },
     }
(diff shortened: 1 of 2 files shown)


isaachinman:
Why are you using `i18n.languages[0]` instead of `props.lng`?

capellini:
I didn't think that there was a difference.  Apologies if I misunderstood.  I can change it back, if you'd like.

isaachinman:
In general I'd prefer PRs to have as little changes as possible to achieve the desired effect. Using `props.lng` is much safer as that is coming through the `withNamespaces` HOC.

capellini:
A few things I noticed while creating this PR:

1. In `Link`, we should probably be spreading `href` in `NextLink`'s `href` prop, rather than just using `pathname`.   In other words, instead of doing this:
```js
      return (
        <NextLink
          href={{ pathname, query: correctedQuery }}  // using just pathname
          as={correctedAs}
          {...removeWithNamespacesProps(props)}
        >
          {children}
        </NextLink>
      )
```
we should do this:
```js
      return (
        <NextLink
          href={{ ...href, query: correctedQuery }}  // spread href
          as={correctedAs}
          {...removeWithNamespacesProps(props)}
        >
          {children}
        </NextLink>
      )
```
For example, this would help us preserve things like `href.hash`.

2. We should probably change the function signature of `lngPathCorrector` to remove `i18n` entirely and enforce that a language is passed in.

Let me know if you would like me to submit PRs for these and I will.

capellini:
Requested corrections made.

capellini:
On second thought, instead of 1 and 2 above, what if we changed the `lngPathCorrector` to:

```js
lngPathCorrector(config: object, { href: object | string, as: string }, language: string): [string, object]
```

That is, take in a `config` object, an object with an `href` string or object and `as` string, and a `language` string and return `correctedAs` as a string and `correctedHref` as an object.

isaachinman:
This PR looks good to go. Let's discuss your comments about `lng-path-corrector` can be in a new issue, please.

PR #151, opened 2019-02-02
refactor: Moving setting of config.backend to default configuration

capellini:
In order to make it more explicit that users that define their own configurations will have to set their own `config.backend` options, moving the definition of `config.backend` to `default-config.js`.

Ref #140.

### DIFF
--- a/__tests__/config/test-helpers.js
+++ b/__tests__/config/test-helpers.js
@@ -13,6 +13,22 @@ const userConfig = {
   localeSubpaths: localeSubpathOptions.FOREIGN,
 }
 
+const userConfigClientSide = {
+  ...userConfig,
+  backend: {
+    loadPath: '/static/translations/{{ns}}/{{lng}}.json',
+    addPath: '/static/translations/{{ns}}/{{lng}}.missing.json',
+  },
+}
+
+const userConfigServerSide = {
+  ...userConfig,
+  backend: {
+    loadPath: '/home/user/static/translations/{{ns}}/{{lng}}.json',
+    addPath: '/home/user/static/translations/{{ns}}/{{lng}}.missing.json',
+  },
+}
+
 const setUpTest = () => {
   const evalFunc = jest.spyOn(global, 'eval').mockImplementation(() => ({
     readdirSync: jest.fn().mockImplementation(() => ['common', 'file1', 'file2']),
@@ -35,7 +51,8 @@ const tearDownTest = (evalFunc, pwd) => {
 }
 
 export {
-  userConfig,
+  userConfigClientSide,
+  userConfigServerSide,
   setUpTest,
   tearDownTest,
 }
--- a/src/config/default-config.js
+++ b/src/config/default-config.js
@@ -1,11 +1,11 @@
 import isNode from 'detect-node'
 
+export const isServer = isNode && !process.browser
 export const localeSubpathOptions = {
   ALL: 'all',
   FOREIGN: 'foreign',
   NONE: 'none',
 }
-export const runningServerSide = isNode && !process.browser
 
 const DEFAULT_LANGUAGE = 'en'
 const OTHER_LANGUAGES = []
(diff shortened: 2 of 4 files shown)


src/config/default-config.js
```diff
@@ -1,11 +1,15 @@
+import isNode from 'detect-node'
+
+export const runningServerSide = isNode && !process.browser
```
isaachinman: Can you please rename this `isServer`?

capellini: Done

isaachinman:
Hey @capellini - is this something we still want to get merged in?

capellini:
Hey @isaachinman - I suppose it depends on whether or not we want to allow users to use our `config.backend` or define their own.  If that's the case, then this makes our expectations more explicit.

But, looking at this more, I think that there's still some more work to be done, if we're going to support the `localePath` and `localeStructure` options.  If users specify those options, then we need to adjust the paths accordingly in `create-config.js`.  But since that is also the case without this change (at least client-side), I'm not sure if you want to incorporate that in this PR, or take care of it in a separate one.

isaachinman:
> If users specify those options, then we need to adjust the paths accordingly in `create-config.js`

Fair point, and I'd want that to be a separate PR that comes before this one.

isaachinman:
Hey @capellini, what are we doing with this one? Shall we try to get it updated and merged, or close it?

ckeeney:
If we wind up closing @258 in favor of this, maybe we should still expose a config option for `localeFileExtension` to making using YAML a simple one liner

isaachinman:
If two config options can produce the same end result, we don't need both. Each config option adds overhead to maintenance.

ckeeney:
I thought it would be straight forward to just default `localeFileExtension` to `json` and make it optional and then use that config option when building the default `backend` options.  I guess it might be confusing though if someone passed in their own `backend` and their own `localeFileExtension` so it's probably not worth the hassle.

capellini:
@isaachinman This should fix #258 by allowing a user to create their own `config.backend` and, consequently, their own `loadPath`.  I agree that we shouldn't add the complexity of another config option - what seems like a simple change may turn out to  be a nightmare to maintain as time passes.  The closer we stay to the `react-i18next` implementation, the better.

With this change, either users use our `backend` implementation, or create their own -- full stop.  Though I must admit, as I am only really using the default implementation, I'm not able to enumerate all of the possible configuration options to ensure that there aren't problems in places, this is a good place to start, IMO.

isaachinman:
@capellini Looks good to me. Can you resolve the conflicts?

ckeeney:
I opened https://github.com/capellini/next-i18next/pull/1 which merges `isaachinman:master` into this `capellini:backend-config-to-default` and resolves the 2 conflicts.

If preferred I'm happy to open a new PR in this repo to merge in `ckeeney:backend-config-to-default` directly into `isaachinman:master`, but since @capellini did all the work and I just resolved his conflicts hopefully @isaachinman can just accept PR after @capellini accepts https://github.com/capellini/next-i18next/pull/1

capellini:
@ckeeney Thank you for the push to rebase.

@isaachinman Rebased and ready for merging whenever you are ready.  Apologies on the delay.

src/config/default-config.js
isaachinman: @capellini Did this get lost in the mix somewhere?

capellini:
@isaachinman Yes, you are correct.  I missed that.  It's fixed now.

__tests__/config/test-helpers.js
```diff
@@ -13,6 +13,22 @@ const userConfig = {
   localeSubpaths: localeSubpathOptions.FOREIGN,
 }
 
+const userConfigClientSide = {
+  ...userConfig,
+  backend: {
+    loadPath: '/static/translations/{{ns}}/{{lng}}.json',
```
isaachinman: Why is this `/static/translations` instead of `/static/locales`?

capellini: It represents a custom user configuration, where they are specifying their own loadPath.

PR #154, opened 2019-02-04
Use rollup for build step

isaachinman:


kachkaev:
Just curious: who are planned consumers of umd, amd and iife bundles if this library is targeting Next.js apps only? 🤔 Unless I miss something, all that's needed is a set of files that Express and webpack can swallow. Ideally, these should be tree-shakable to reduce browser bundle size.

isaachinman:
@kachkaev This is a WIP - not ready for feedback really. I opened the PR so I could see how builds would run in CI. The approach is just a copy of rollup setups we've done in the past.

You're probably right that we could get away with just an `es` dir for NextJs to consume and a `commonjs` transpilation of the middleware for NodeJs to consume.

PR #156, opened 2019-02-05
fix: Lng path corrector unknown lang

capellini:
* Ensure that `fallbackLng` is an array
* `config.whitelist = config.allLanguages`
* `config.load = 'currentOnly'`
* Ensure that `req.i18n.languages` is an array in `lng-from-req` before we proceed with the `find` operation -- if it is not, use `defaultLanguage`

See [comment](https://github.com/isaachinman/next-i18next/issues/137#issuecomment-459844339) in #137.

### DIFF
--- a/src/utils/lng-from-req.js
+++ b/src/utils/lng-from-req.js
@@ -3,7 +3,7 @@ export default (req) => {
   const { allLanguages, defaultLanguage, fallbackLng } = req.i18n.options
 
   const language = req.i18n.languages.find(l => allLanguages.includes(l))
-    || fallbackLng
+    || fallbackLng[0]
     || defaultLanguage
 
   return language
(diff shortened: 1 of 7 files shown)


src/utils/lng-from-req.js
```diff
@@ -2,10 +2,9 @@ export default (req) => {
...
-  const language = req.i18n.languages.find(l => allLanguages.includes(l))
-    || fallbackLng
-    || defaultLanguage
-
-  return language
+  if (Array.isArray(req.i18n.languages) && req.i18n.languages.some(l => allLanguages.includes(l))) {
+    return req.i18n.languages.find(l => allLanguages.includes(l))
+  }
 
+  return fallbackLng[0] || defaultLanguage
```
capellini: We should also be checking to see if fallbackLng is an object here, but since that's a separate issue and this mimics the current functionality of `lngFromReq`, I decided not to change it in this PR.

PR #159, opened 2019-02-05
refactor: Babel setup

isaachinman:
After some further thought, I don't think `rollup` or `webpack` are truly necessary for a lib like this.

The primary consumer of this library are NextJs applications, which will have their own build process to compile both server and client code.

End users can modify their own Babel configuration to target whatever browsers they'd like.

In the end, we should primarily export `es` modules, but it's also useful to export `cjs` for those people who don't want to set up `es` modules for their NodeJs processes.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,27 +1,61 @@
 {
-  "presets": [
-    [
-      "@babel/env",
-      {
-        "targets": {
-          "node": "current"
-        }
-      }
-    ],
-    ["@babel/preset-react"]
-  ],
-  "plugins": [
-    [
-      "module-resolver",
-      {
-        "root": ["./src"]
-      }
-    ],
-    [
-      "add-module-exports",
-      {
-        "addDefaultProperty": true
-      }
-    ]
-  ]
+  "env": {
+    "es": {
+      "presets": [
+        [
+          "@babel/env",
+          {
+            "modules": false,
+            "targets": {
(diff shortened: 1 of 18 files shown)


lucasfeliciano:
> The primary consumer of this library are NextJs applications, which will have their own build process to compile both server and client code.

That is true but I would expect that I didn't have to manually change my build configuration in order to use a library.

I think that libraries should provide everything that is needed to run.


> After some further thought, I don't think `rollup` or `webpack` are truly necessary for a lib like this.

Not saying that was a bad decision but why did you decide that? Rollup make things easier ( IMO ).
We could use the same setup as i18next/react-i18next.

isaachinman:
> I think that libraries should provide everything that is needed to run.

What would we not be providing with this solution?

> Not saying that was a bad decision but why did you decide that?

Because we can get both `cjs` and `es` exports from Babel directly. Rollup is just going to end up using Babel at the end of the day.

isaachinman:
> I would expect that I didn't have to manually change my build configuration in order to use a library.

To clarify, you definitely would not need to change anything to use `next-i18next` in a NextJs application with this modified setup. You _would_ need to manually set up your own `.babelrc` file and manage targets if you want to support IE11, for instance. But that is true regardless of whether you're using `next-i18next` or not. Hope that makes more sense now.

PR #174, opened 2019-02-09
Refactor lngPathCorrector

### LINKED ISSUE #155: Change function signature for lngPathCorrector
1. Currently, the `lngPathCorrector` receives as an argument, the `i18n` object which is used only to determine a default argument for language.  However, I don't think that there's an instance where we don't know the right language to pass into `lngPathCorrector` and, in some cases (e.g. `Link`), it's better to use the `lng` prop anyway.
2. Also, the `lngPathCorrector` takes in an object with an `as` string and an `href` object, but we could easily change this to be more user-friendly by taking in an `href` object or string and, optionally, an `as` string.  If the `as` string is not provided, we can determine what it is from the `href` string/object.  Currently, we do this, but the logic for doing so is not perfomed within `lngPathCorrector` but, instead, is done in `Link`  and `appWithTranslation`.  But I think the logic belongs inside the `lngPathCorrector`.

1. Make the language a required parameter and remove the need to pass in `i18n`.
2. Have `lngPathCorrector` take in an object with the `href` string or object and, optionally, the `as` string and perform any necessary steps needed to just return the corrected `as` string and `href` object.

We could keep things the way they are, but it's better to keep all of the logic inside of `lngPathCorrector`.  Also, should we decide someday to expose this function, it's easier to use in the proposed form.

## Additional context
N/A

capellini:
Refactor lngPathCorrector to have the following signature:
```js
lngPathCorrector(
  { defaultLanguage: string, allLanguages: string[] },
  { href: object | string, as?: string },
  language: string
): { as: string, href: object }
```

Note that, to provide the greatest flexibility, I assumed that the `href` object need not contain a `query` key, if passed in by the user.  I also do some type checking to ensure that `href` is either a string or an object and `as`, if defined, is a string.

Also, in my testing, I noticed that, should a parsed `URL` object that contain both a `search` and `query` key, the `search` key string overrides any changes made to the `query` object.

```js
> const { format, parse } = require('url')
undefined

> const foo = parse('/foo?option1=value1#hash1', true)
undefined

> foo.query.lng = 'de'
'de'

> foo
Url {
  protocol: null,
  slashes: null,
  auth: null,
  host: null,
  port: null,
  hostname: null,
  hash: '#hash1',
  search: '?option1=value1',
  query: { option1: 'value1', lng: 'de' },
  pathname: '/foo',
  path: '/foo?option1=value1',
  href: '/foo?option1=value1#hash1' }

> format(foo)
'/foo?option1=value1#hash1'

> delete foo.search
true

> format(foo)
'/foo?option1=value1&lng=de#hash1'
```

Considering in past PRs we had such trouble with manipulating a string query, I opted to remove the `search` key from the `href` object, should it exist (for example, if a user passes in a `URL` object as the `href` object) and, if necessary, add a `query` object to `href`, should it not exist (for example, if a user passes in an object without the `query` object in it).  I thought that this was a better option than attempting to add `lng` to both the `query` object and the `search` string.

Resolves #155.


isaachinman:
@capellini Sorry, just getting back to this. Can you resolve the conflict?

capellini:
@isaachinman No worries.  Rebased.

isaachinman:
@capellini This looks good, not sure I have any technical feedback, actually. Did you do any manual regression testing?

capellini:
> Did you do any manual regression testing?

Yes, I did, but if you'd prefer, we can wait on this until #186 is merged, which should give us some added confidence that it works.

isaachinman:
I merged #186. Can you rebase here and force push to rerun the CI? Then we can merge this.

capellini:
Rebased.

PR #186, opened 2019-02-16
feat: Multiple e2e tests with different config options

### LINKED ISSUE #118: Add e2e test with localeSubpaths enabled
Following up on [this comment](https://github.com/isaachinman/next-i18next/pull/96#issuecomment-456379796), it would be good to add an e2e test which has `localeSubpaths` enabled.

This test would be nearly identical to our current e2e test.  The only change would be an added `localeSubpaths: true` in the `userConfig` object passed into `NextI18Next` in `i18n.js`.  Also, any changes to the simple example should be compatible with also using `localeSubpaths`.  I think that we can easily add another e2e test by using the existing example, but having the `localeSubpaths` option be configurable.

Some ideas:
1. Symlink all files in the current example to a new example, with the exception of `i18n.js`, which would be a new file with `localeSubpaths` enabled.  This will most likely cause problems with developers using Windows, as I'm not sure if symlinks work well in Windows.
2. Change the existing example to use an environment variable to set `localeSubpaths`.  `i18n.js` could look something like the following:
```js
const NextI18Next = require('next-i18next')
const localeSubpaths = process.env.USE_LOCALE_SUBPATHS || false

  otherLanguages: ['de'],
  localeSubpaths
})
```
In this way, the example can be used for cases where `localeSubpaths` is `true` and `false`, depending on whether or not `USE_LOCALE_SUBPATHS` is defined when we run our tests.

isaachinman:
Fixes #118.

### DIFF
--- a/examples/custom-routing/server.js
+++ b/examples/custom-routing/server.js
@@ -4,6 +4,7 @@ const nextI18NextMiddleware = require('next-i18next/middleware')
 
 const nextI18next = require('./i18n')
 
+const port = process.env.PORT || 3000
 const app = next({ dev: process.env.NODE_ENV !== 'production' })
 const handle = app.getRequestHandler();
 
@@ -21,6 +22,6 @@ const handle = app.getRequestHandler();
 
   server.get('*', (req, res) => handle(req, res))
 
-  await server.listen(3000)
-  console.log('> Ready on http://localhost:3000') // eslint-disable-line no-console
+  await server.listen(port)
+  console.log(`> Ready on http://localhost:${port}`) // eslint-disable-line no-console
 })()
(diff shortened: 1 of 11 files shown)


capellini:
@isaachinman Two comments/suggestions:
1. I noticed that you removed the `--silent` option when running tests.  Was that intentional?
2. Currently, the locale subpaths e2e test that exists tests our `appWithTranslation` usage of locale subpaths with a non-default language.  You may also want to include a few other tests:
    1. A test that also clicks the button to navigate to the second page, ensuring that the locale subpath exists on the second page.  This would verify that navigation through our `Link` element is also functioning properly when a non-default language is selected.
    2. A test that switches back to the default language, ensuring that do not use locale subpaths with the default locale. This would verify the behavior in `AppWithTranslation` when dealing with the default language.
    3. A test, with the default language set, that navigates to the second page by clicking the button.  This would verify the behavior in `Link` when dealing with the default language.

Other than that, this looks good to me.

isaachinman:
1. No, good catch. I removed it because I needed to actually log stuff out in the dev process.
2. This was a proof of concept basically, and is less concerned with overall test coverage. I just wanted to get the methodology in place.

PR #239, opened 2019-03-15
fix(types): remove extends of WithNamespaces

prokopsimek:
- Removed `extends WithNamespaces`, let the developer uses his own types for the component. 
- Do not enforce developer to define I18n `t` function (removal of extends - e.g. `React.SFC<{myProp: string}>`)
- If the developer wants to use the `t` function in his Component, he has to set props with I18nProps interface (`React.SFC<I18nProps>`) or extend his own interface with I18nProps (`interface CustomProps extends I18nProps { myProp: string }`)

Now the typing works correctly:

```
yarn run v1.13.0
$ rimraf ./build && next build && tsc
Creating an optimized production build ...

> Using external babel configuration
> Location: "/Users/prokop/Sites/_yoyo/xoxo/.babelrc"
Compiled successfully.

 ┌ /
 ├ /_app
 ├ /_document
 ├ /_error
 ├ /principles
 └ /signers

pages/index.tsx:21:10 - error TS2739: Type '{ totalCount: any; signers: any; }' is missing the following properties from type 'Pick<SignProps, "content" | "signers" | "totalCount" | "manifesto" | "currentUser">': content, manifesto

21         <Sign totalCount={props.totalCount} signers={props.signers} />
            ~~~~

pages/index.tsx:22:10 - error TS2739: Type '{ totalCount: any; signers: any; }' is missing the following properties from type 'Pick<AvatarsProps, "content" | "signers" | "totalCount" | "manifesto" | "currentUser">': content, manifesto

22         <Avatars totalCount={props.totalCount} signers={props.signers} />
            ~~~~~~~


Found 2 errors.

error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

@isaachinman Sorry for that mistake.

PR #240, opened 2019-03-19
fix(types): Extend config with options from i18next

kachkaev:
Resolves https://github.com/isaachinman/next-i18next/pull/239#issuecomment-473892273

See [definition of `i18next InitOptions`](https://github.com/i18next/i18next/blob/19b2785206b1024497dab78e8ee6c1fb9dd90472/index.d.ts#L156)

cc @felixmosh

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -11,19 +11,18 @@ import { LinkProps } from 'next-server/link';
 import { SingletonRouter } from 'next-server/router';
 import i18next from 'i18next';
 
-export interface INextI18NextConfig {
-  browserLanguageDetection: boolean;
-  serverLanguageDetection: boolean;
-  defaultNS: string;
-  defaultLanguage: string;
-  ignoreRoutes: string[];
-  localePath: string;
-  localeStructure: string;
-  otherLanguages: string[];
-  localeSubpaths: 'none' | 'foreign' | 'all';
-  use: any[];
-  customDetectors: any[];
-}
+export type InitConfig = {
+  browserLanguageDetection?: boolean;
+  serverLanguageDetection?: boolean;
+  defaultLanguage?: string;
+  ignoreRoutes?: string[];
+  localePath?: string;
+  localeStructure?: string;
+  otherLanguages?: string[];
+  localeSubpaths?: "none" | "foreign" | "all";
+  use?: any[];
+  customDetectors?: any[];
+} & i18next.InitOptions;
 
 export interface I18nProps {
   t(key: string, option?: object): string;
@@ -35,7 +34,7 @@ declare class NextI18Next {
   Router: SingletonRouter;
   i18n: i18next.i18n;
 
-  constructor(config: Partial<INextI18NextConfig>);
+  constructor(config: InitConfig);
 
   withNamespaces(
     namespace: Namespace | NamespaceExtractor,


types.d.ts
```diff
@@ -11,19 +11,18 @@ import { LinkProps } from 'next-server/link';
...
-  defaultLanguage: string;
-  ignoreRoutes: string[];
-  localePath: string;
-  localeStructure: string;
-  otherLanguages: string[];
-  localeSubpaths: 'none' | 'foreign' | 'all';
-  use: any[];
-  customDetectors: any[];
-}
+export type InitConfig = {
```
kachkaev: We can no longer prefix this typing with `I` as it is not an interface any more. I'm renaming it to `InitConfig` to avoid too much verbosity (there are no more primary objects with configs in this library).

Technically, renaming `INextI18NextConfig` is a breaking change because it is exported. I can add something like this:

```
    /**
     * @deprecated use InitConfig instead
     */
export type INextI18NextConfig = InitConfig;
```

Might be an overkill given the age of typings.

felixmosh: I do think that it is an overkill to add the deprecation.

isaachinman: @felixmosh Can you approve the PR if you believe it's mergeable?

felixmosh: Will check it in my project, just to be sure.

felixmosh: Yeah, it works :]

types.d.ts
```diff
@@ -11,19 +11,18 @@ import { LinkProps } from 'next-server/link';
...
+  browserLanguageDetection?: boolean;
+  serverLanguageDetection?: boolean;
+  defaultLanguage?: string;
```
isaachinman: This is not an optional param.

kachkaev: According to the README, `defaultLanguage` has a default value (`en`). `otherLanguages` though is marked as required. Did you mean that option here, just github placed your comment next to the neighbouring line?

kachkaev: Hmm. The minimal example in that same README defines both `defaultLanguage` and `otherLanguages`. Conceptually, they both seem to be required parameters, which I'm happy with. In this case, we can remove `?` from both of them and remove _Default value_ from the docs. Otherwise, there's a contradiction. WDYT?

kachkaev: | Key  | Default value |
| ------------- | ------------- |
| `defaultLanguage` (required) | - |
| `otherLanguages` (required) | - |
| `browserLanguageDetection`  | `true`  |
| `defaultNS` | `'common'`  |
| `ignoreRoutes`  | `['/_next', '/static']`  |
| `localePath` | `'static/locales'`  |
| `localeStructure` | `'{{lng}}/{{ns}}'`  |
| `localeSubpaths` | `'none'`  |
| `serverLanguageDetection` | `true`  |
| `use` (for plugins) | `[]`  |
| `customDetectors` | `[]`  |

isaachinman: Hm, good point. Not sure how we'd communicate default values. `otherLanguages` is required in that it must be an array of strings with a length more than zero... but let's just leave it out for now. I hope to bring an enterprise TS project on board to `next-i18next` soon, and will work out those kinks as we go.

kachkaev: Re. an array with a non-zero length. Turns out TypeScript allows defining such a typing in theory 👀
https://stackoverflow.com/a/49910890/1818285

PR #241, opened 2019-03-19
fix(types): remove T from return types

prokopsimek:
Remove T from return types because it requires `t` function in props and it's not necessary.

before:
image

after:
image

pls check if it works for you @felixmosh

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -42,7 +42,7 @@ declare class NextI18Next {
     options?: WithNamespacesOptions,
   ): <T extends React.ComponentType<any>>(
     component: T,
-  ) => T & (T extends React.ComponentType<infer P> ? React.ComponentType<Subtract<P, I18nProps>> : never);
+  ) => T extends React.ComponentType<infer P> ? React.ComponentType<Subtract<P, I18nProps>> : never;
 
   appWithTranslation<P extends object>(Component: React.ComponentType<P> | React.ElementType<P>): any;
 }


isaachinman:
@felixmosh @kachkaev Mind having a look?

felixmosh:
@prokopsimek it is not works with decorator usage.

prokopsimek:
@felixmosh ok, could you please try this?

```
Subtract<T, I18nProps> & (T extends React.ComponentType<infer P> ? React.ComponentType<Subtract<P, I18nProps>> : never);
```

felixmosh:
> ```
> Subtract<T, I18nProps> & (T extends React.ComponentType<infer P> ? React.ComponentType<Subtract<P, I18nProps>> : never);
> ```

It is not works, cause T is the component class type, not it's props

prokopsimek:
> @prokopsimek it is not works with decorator usage.
> We can change the type of `I18nProps` and make `t` optional

@felixmosh It's not possible, because of the type checking. TypeScript throws an error `Cannot invoke an object which is possibly 'undefined'.ts(2722)` where I call `props.t('more')`

prokopsimek:
I have found the problem. It's because my component `extends React.SFC<T>` and it makes return type of `withNamespaces(...): T & never` instead of the Subtract.

felixmosh:
> > @prokopsimek it is not works with decorator usage.
> > We can change the type of `I18nProps` and make `t` optional
> 
> @felixmosh It's not possible, because of the type checking. TypeScript throws an error `Cannot invoke an object which is possibly 'undefined'.ts(2722)` where I call `props.t('more')`

How is that related to type checking? this is a runtime error.
If we convert `t` to be optional this will solve the issue of requirement of `t` prop.

prokopsimek:
> > > @prokopsimek it is not works with decorator usage.
> > > We can change the type of `I18nProps` and make `t` optional
> > 
> > 
> > @felixmosh It's not possible, because of the type checking. TypeScript throws an error `Cannot invoke an object which is possibly 'undefined'.ts(2722)` where I call `props.t('more')`
> 
> How is that related to type checking? this is a runtime error.
> If we convert `t` to be optional this will solve the issue of requirement of `t` prop.

@felixmosh I have set `"strictNullChecks": true`

felixmosh:
Ha, sorry then :)

Maybe we will go to a different approach, by using overloading, maybe we can declare 2 interfaces, that will work in both of the scenarios

prokopsimek:
@felixmosh Have you tried to use react-i18n's decorators and not only next-i18next's decorators? Because I haven't seen some examples of how to use react-i18n's functions as decorators. Is it common to use it this way in react-i18n?

felixmosh:
@prokopsimek, yes I'm using it as a decorator for 2 years already

isaachinman:
Let me know when this is ready to go and I'll review it.

felixmosh:
@prokopsimek lets drop the decoration types for now, I can't find any proper way to support both, and decorators are proposal, so I will handle it in my app.

AppliftingHQ:
@felixmosh nope

this works for me:

```typescript
withNamespaces(
    namespace: Namespace | NamespaceExtractor,
    options?: WithNamespacesOptions,
  ): <P>(
    component: React.ComponentType<P>,
  ) => React.ComponentType<Subtract<P, I18nProps>>;
```

prokopsimek:
@felixmosh nope

this works for me:

```typescript
withNamespaces(
    namespace: Namespace | NamespaceExtractor,
    options?: WithNamespacesOptions,
  ): <P>(
    component: React.ComponentType<P>,
  ) => React.ComponentType<Subtract<P, I18nProps>>;
```

kachkaev:
Do we need `I18nProps` as such ? Can we use `WithNamespaces` from `react-i18next` when subtracting? It contains genuine typings for `t` and other props, see https://github.com/isaachinman/next-i18next/pull/239#issuecomment-475916007 for context.

prokopsimek:
@kachkaev it seems that we can :)

```
withNamespaces(
    namespace: Namespace | NamespaceExtractor,
    options?: WithNamespacesOptions,
  ): <P>(
    component: React.ComponentType<P>,
  ) => React.ComponentType<Subtract<P, WithNamespaces >>;
```

works for me

prokopsimek:
@felixmosh @isaachinman So should I update the PR to merge it?

isaachinman:
Update how? Just let me know when this is ready to go and I'll review it.

prokopsimek:
@isaachinman As is the last my code snippet in comments. Instead of having working types for decorators.

isaachinman:
Looks fine to me.

kachkaev:
+1 for using `WithNamespaces` from `react-i18next` instead of `I18nProps`. This custom typing can be deleted as it's not used anywhere else.

PR #242, opened 2019-03-20
fix(types): Add config to NextI18Next declaration

pavelvladimir:
I have a problem with typescript error because I use allLanguages and defaultLanguage key from config when I generate the sitemap.

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -29,7 +29,7 @@ export type NextI18nextConfig = {
   allLanguages: string[];
   whitelist: string[];
   preload: string[];
-} & InitConfig;
+} & i18next.InitOptions;
 
 export interface I18nProps {
   t(key: string, option?: object): string;
--- a/types.d.ts
+++ b/types.d.ts
@@ -1,4 +1,4 @@
-import * as React from 'react';
+import * as React from "react";
 import {
   TransProps,
   Namespace,
@@ -6,10 +6,10 @@ import {
   WithNamespacesOptions,
   WithNamespaces,
   Subtract
-} from 'react-i18next';
-import { LinkProps } from 'next-server/link';
-import { SingletonRouter } from 'next-server/router';
-import i18next from 'i18next';
+} from "react-i18next";
+import { LinkProps } from "next-server/link";
+import { SingletonRouter } from "next-server/router";
+import i18next from "i18next";
 
 export type InitConfig = {
   browserLanguageDetection?: boolean;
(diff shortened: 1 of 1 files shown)


isaachinman:
@kachkaev What's going on with the typing? Why are we running into constant problems?

kachkaev:
@pavelvladimir what’s the reason for creating a new type given that InitConfig exists already? Also your type makes too many options mandatory, which does not match what README says.

types.d.ts
```diff
@@ -1,15 +1,15 @@
-import * as React from 'react';
+import * as React from "react";
```
kachkaev: You’ve probably used prettier locally, which introduced unnecesssy diffs. These changes, if need to be made, should to to a separate PR IMO.

@isaachinman I can help with setting up Prettier throughout the project to avoid unncesssry formatting diffs in future.

pavelvladimir:
@kachkaev I filled config with InitConfig content and constructed from createConfig

types.d.ts
pavelvladimir: So please add a prettier config to root directory

kachkaev:
Your new type looks to me like a duplicate of InitCoding, where all fields are required. i don’t understand what it can be useful for.

pavelvladimir:
@kachkaev For example ..In your InitConfig the defaultLanguage is set as optional

types.d.ts
kachkaev: I’m for it, just in a separate PR. These extra formatting diffs generally make it harder to make sense of the changes required.

kachkaev:
This was discussed in https://github.com/isaachinman/next-i18next/pull/240#discussion_r267082249

defaultLanguage and otherLanguages are just made mandatory, will this solve your problem?

pavelvladimir:
@kachkaev Yes

kachkaev:
I still don’t understand what the additional type is solving. If it’s the same config that people pass down, then it wont necessary have all properties that need to be availalve.

pavelvladimir:
@kachkaev For example ..allLanguages comes from createConfig (not from custom configuration)

kachkaev:
I see. Should it be an extended version of InitConfig then?

pavelvladimir:
@kachkaev Does InitConfig include the allLanguages key? What do you suggest?

kachkaev:
It does not: https://github.com/isaachinman/next-i18next#options

`allLanguages` comes from [create-config.js](https://github.com/isaachinman/next-i18next/blob/c15256e16070b101467ab21be51d0170aad7bf11/src/config/create-config.js), same as a few other values.

I believe your change makes sense now. The only thing I'd do differently is naming type as `Config` rather than `NextI18nextConfig`. We are in the context of `NextI18next` all the time so probably need to prefix types in addition. That's a matter of taste though.

We currently use two naming strategies (`InitConfig` – no prefix vs `NextI18nextConfig` – has prefix). Consistency in one way or another seems better to me than what we have now.

isaachinman:
Is this change still desired? If so, can you resolve the conflicts?

isaachinman:
@kachkaev Are you happy with these changes?

kachkaev:
LGTM

PR #258, opened 2019-04-06
Add yaml support

ckeeney:
Here are some failing tests.
I'm a little bit confused about the expected output for `createConfig`.

Here is an excerpt from a test already merged into master:
https://github.com/isaachinman/next-i18next/blob/a4444008dc2be024b4ab415c9f781e241b4e2802/__tests__/config/create-config.test.js#L131-L148

I would expect the `backend.loadPath` and `backend.addPath` to need to reference the `localePath: '/static/translations'` `localefileExtension: 'yml'` passed into the `userConfig`:

```js
expect(config.backend.loadPath).toEqual('/static/translations/{{lng}}/{{ns}}.yml')
```

I don't believe this is an actual deep merge, which may be related to this issue:

https://github.com/ckeeney/next-i18next/blob/fac15872c696908f322923796010d15418fd82b3/src/config/create-config.js#L5-L23


isaachinman:
@capellini Mind having a look? I think you wrote the config tests.

capellini:
This is related to #140, [this comment](https://github.com/isaachinman/next-i18next/pull/48#discussion_r242015868), and [this one too](https://github.com/isaachinman/next-i18next/pull/151#issuecomment-467858373).

The problem isn't so much with the test as with the logic in `create-config.js`.  Yes, I agree, if the user specifies an `addPath` and/or `loadPath`, intuitively it seems like that should be reflected in the config that is created.  However, that's not what `create-config.js` does.  The problem is not in the deep merge, but instead is related to [these lines](https://github.com/isaachinman/next-i18next/blob/2e4cf1732979c16c46a3ff41c0b27e9ad93e7bfb/src/config/create-config.js#L51-L54).  That is, at the moment, we unconditionally overwrite `addPath` and `loadPath` if we're server-side.  

I seem to remember having a discussion as to why we do this, and we've had many discussions on how difficult it is to manage configuration and what we will or won't allow the user to specify, but I can't seem to find the discussion(s) in my searching.

To me, it makes sense to allow a user to override `addPath` and `loadPath`, but since I'm not doing so in my projects, this issue has never been a big deal for me and, consequently, I don't really know the repercussions of allowing a user to specify this.  But I seem to recall that there was a reason.

#151 should solve this specific issue (sorry @isaachinman that I did not respond), though there are a few more issues that we would need to deal with after it is merged.  If you want to re-open the PR, I will do what is necessary to ensure that it's merged and then take care of any follow-on PRs to fix the remaining issues.

p.s. @isaachinman - sorry I have been "off the grid" the past few weeks.  I've had an unexpected surge in work lately.  I know that I have a PR that I still have to submit and I'm hoping to get to it soon.

isaachinman:
Yeah - will #151 actually make this `localeFileExtension` option unnecessary? If so, I prefer that approach.

@capellini No worries, very happy to have any level of contribution you can manage.

ckeeney:
> The problem is not in the deep merge, but instead is related to these lines. That is, at the moment, we unconditionally overwrite addPath and loadPath if we're server-side.

[I made changes](https://github.com/isaachinman/next-i18next/pull/258/files#diff-e588a4bb9e37c5a5304146435a628df8) to those lines in my branch so they take into account the `localeFileExtension` and they pass the test that is built for them.  The tests that fail are for building the client-side configuration.  I think possibly those tests are broken or at least some more documentation should be added about why the client-side tests for using the custom configuration are written to pass when the `loadPath` and `addPath` ignore the value set for `localePath`.

Specifically, why isn't this line:
https://github.com/isaachinman/next-i18next/blob/master/__tests__/config/create-config.test.js#L146

Written as 
```js
      expect(config.backend.loadPath).toEqual('/static/translations/{{lng}}/{{ns}}.json')
```

I still think the "deep merge" is problematic.  What is the point of this line:

https://github.com/isaachinman/next-i18next/blob/2e4cf1732979c16c46a3ff41c0b27e9ad93e7bfb/src/config/create-config.js#L5

This is not a deep merge.

https://github.com/isaachinman/next-i18next/blob/2e4cf1732979c16c46a3ff41c0b27e9ad93e7bfb/src/config/create-config.js#L18-L23


If we want to close this in favor of #151 that's fine, but what does it take to get that merged in?

ckeeney:
Sorry I misunderstood the deep merge.  I've added `backend` to the `deepMergeObjects` array and added a `backend` key to the test `userConfig`.  Now only the end to end tests are failing.

isaachinman:
@ckeeney Thanks for your continued work on this. Let's wait and see what @capellini says about #151.

isaachinman:
For what it's worth, if the e2e tests are failing, it probably means the build is fully broken.

ckeeney:
I'm not sure I'm running the tests properly.  The tests on master fail for me unless i run `yarn` in the example app.  I'm sure it has something to do with how `next-i18next` is specified in `package.json`.

https://github.com/isaachinman/next-i18next/blob/f47538de9a83a7cded9cb373698d9ee17195fa44/examples/simple/package.json#L25-L27

Would I have to re-run `yarn` in the example app before running tests to see how the changes affect the e2e tests, or is it enough to just run `yarn` once in the example app?

ckeeney:
Nevermind I just saw this:  https://github.com/isaachinman/next-i18next/pull/260/files

Thanks.

isaachinman:
@ckeeney Please bear with me - I am going to finally get around to setting up linking properly.

isaachinman:
Fixed by #302.

Issue #237, opened 2019-03-15
Add a "How To Contribute" section to docs


isaachinman:
See #234 and #235.

prokopsimek:
Some ideas: 
- https://help.github.com/en/articles/setting-guidelines-for-repository-contributors
- https://gist.github.com/PurpleBooth/b24679402957c63ec426
- https://github.com/nayafia/contributing-template

ckeeney:
Documentation on how to run the tests would be a good place to start. I think you can't run them unless you install the dependencies in the example app first.

isaachinman:
@ckeeney Yes - that's something that's come up several times. It's a mistake, not something that needs documentation. Fixed with #260.

ckeeney:
I still can't run the tests without some extra steps:

```
 ckeeney@ckeeney-Serval-WS  ~/tmp  git clone https://github.com/isaachinman/next-i18next.git
Cloning into 'next-i18next'...
remote: Enumerating objects: 17, done.
remote: Counting objects: 100% (17/17), done.
remote: Compressing objects: 100% (16/16), done.
remote: Total 2303 (delta 6), reused 1 (delta 0), pack-reused 2286
Receiving objects: 100% (2303/2303), 1.80 MiB | 3.22 MiB/s, done.
Resolving deltas: 100% (1401/1401), done.

 ckeeney@ckeeney-Serval-WS  ~/tmp  cd next-i18next
 ckeeney@ckeeney-Serval-WS  ~/tmp/next-i18next   master  yarn
yarn install v1.12.3
[1/5] Validating package.json...
[2/5] Resolving packages...
[3/5] Fetching packages...
info fsevents@1.2.4: The platform "linux" is incompatible with this module.
info "fsevents@1.2.4" is an optional dependency and failed compatibility check. Excluding it from installation.
info fsevents@1.2.7: The platform "linux" is incompatible with this module.
info "fsevents@1.2.7" is an optional dependency and failed compatibility check. Excluding it from installation.
[4/5] Linking dependencies...
[5/5] Building fresh packages...
Done in 13.59s.

 ckeeney@ckeeney-Serval-WS  ~/tmp/next-i18next   master  yarn test
yarn run v1.12.3
$ yarn build:example && NODE_ENV=test jest --maxWorkers=1 --silent
$ cd examples/simple && yarn build
$ next build

● Client █████████████████████████ chunk asset optimization (92%) TerserPlugin 
 

✖ Server
[...]

PR #261, opened 2019-04-08
fix: Add linking to simple example

isaachinman:


isaachinman:
@ckeeney Let's continue discussion from #237 here. I'm going to go forward with this approach now as I believe it provides the best local dev experience. I had forgotten that it's not safe to use the `postinstall` script, as that'll run when end users install from registries as well.

By adding these changes, you should be able to clone a fresh copy of the repo and simply do `yarn && yarn test` to run the tests.

Moreover, the example will be linked to the root source, so you can just run a build in the root and the example will hot reload (if you're running it in dev mode).

ckeeney:
This worked for me.  Thanks.

Issue #290, opened 2019-05-02
Babel doesn't polyfill according preset-env


StarpTech:
## Describe the bug
As mentioned in https://github.com/isaachinman/next-i18next/issues/107 I expected that es6 functions are polyfilled but it's not. If you check the `dist/es/utils/lng-path-corrector` there is no polyfill for `Array.includes`. This throws in IE Edge.

isaachinman:
We transpile to the same target that NextJs does. Polyfilling is up to the user. Am I misunderstanding?

StarpTech:
mmh.. babel doesn't transpile `ǹode_modules` by default. Any idea how I can solve it with current nextjs version? I tried the `with-polyfill` example but it doesnt work.

isaachinman:
Transpiling and polyfilling are two separate concerns. See https://github.com/zeit/next.js/issues/4572#issuecomment-395960145.

Ideally we'd like to support all browsers that NextJs does out of the box as well.

Does this throw in IE11? If so, we can think about improving that.

StarpTech:
Yeah, I already read it but it doesn't my answer how to polyfill a third party lib with a regular next setup. From my understanding next-i18next should provide a compatible version for all browsers because it's not part of the next babel process right?

isaachinman:
> a compatible version for all browsers 

No, absolutely not. We aim to support the same subset of browsers that NextJs does.

StarpTech:
Sry, I mean all browsers that NextJs that includes ie11+.

isaachinman:
If `next-i18next` breaks in IE11, please provide a detailed report and we'll fix it.

StarpTech:
I could fix it by adding the 

```js
import 'core-js/es6/array';
import 'core-js/es7/array';
```
to the polyfills.js as shown in the `with-polyfill` example. We should add it to the readme.

kachkaev:
Related: https://github.com/isaachinman/next-i18next/issues/107

IMO this lib does not have to support IE11 out of box but for any steps enabling that, a section in README would be helpful.

StarpTech:
I'm corious how it works if you import these as functions `/fn/... . From my understanding in that way it can't work because the prototype isn't touched.

StarpTech:
@isaachinman at the end the module will fail due to exact such missing methods. This means next-i18n doesn't support the same browsers as nextjs. I'm fine with documenting it in the Readme. I will send a PR.

isaachinman:
[...]

PR #291, opened 2019-05-02
docs: Add readme note to inform about polyfill solution for browsers with restricted ES support.

StarpTech:


isaachinman:
https://github.com/isaachinman/next-i18next/issues/290#issuecomment-488767289

StarpTech:
This is temporary thing until the fix was implemented.

isaachinman:
Polyfilling is not something which is specific to IE. I'll merge this as a temporary measure if the message can be more generic.

StarpTech:
Propose something....

isaachinman:
"Support for older browsers may require polyfills until #290 is resolved"

StarpTech:
This is also not accurate. Edge is not old it's the default browser of every windows 10 installation. We only talking about ES compatibility.

StarpTech:
"Support for browsers with limited support for modern ES6/7 features may require polyfills until #290 is resolved."

isaachinman:
Sure, fair enough. Please give the PR a semantic title and I'll merge it.

Issue #152, opened 2019-02-03
Duplicate core-js module in the app's prod bundle due to polyfilling


kachkaev:
## Describe the bug

After upgrading `next-i18next` from version 0.18.1,  I've started seeing two copies of `core-js` in my prod bundle. This may be related to the fact that it's in `next-i18next`'s `dependencies` and is also because of `@babel/polyfill` in `.babelrc` ([9c96479e](https://github.com/isaachinman/next-i18next/commit/9c96479e27297e300d04b62b17db3d3aa80ae4c4)).

Before / after:

<img width="771" alt="screenshot 2019-02-03 at 16 53 06" src="https://user-images.githubusercontent.com/608862/52179598-70c4b380-27d4-11e9-80d7-f08901ec4b84.png">

<img width="804" alt="screenshot 2019-02-03 at 16 53 15" src="https://user-images.githubusercontent.com/608862/52179596-702c1d00-27d4-11e9-8e45-e86a76d2d907.png">

The problem may be to do with the fact that `core-js` is now used as both an es module and a commonjs one by webpack when the app is built.

## Occurs in next-i18next version
0.20.0...0.25.0

## Steps to reproduce

Run bundle analyzer for the Next.js app

## OS (please complete the following information)
 - Device:  MBP 2017 15"
 - Browser: N/A

## Additional context
Perhaps, polyfilling should be removed from the library completely – doing it only once on the app level ensures that it's not done twice and so there are no duplicate items in the final bundle. As a replacement, the README could advise on what an app developer should do to support older browsers. Some thoughts are here: https://github.com/isaachinman/next-i18next/issues/107

kachkaev:
Yes, I had two versions of `core-js`, but even if I forced one of them via `yarn` and `package.json` `resolutions`, this did not help.

Looking into `node_modules`, reveals that `core-js` is shipped in multiple variants:

<img width="188" alt="screenshot 2019-02-03 at 17 34 18" src="https://user-images.githubusercontent.com/608862/52180103-fac34b00-27d9-11e9-844e-dd03ac8a9ec8.png">

`next-i18next` is built as commonjs, which in turn uses a commonjs variant of `core-js`. On the other hand, the polyfills I mention in my app's js modules, resolve into a different variant and this is how two copies of `core-js` end up in the bundle.

isaachinman:
I think we should reopen #129, and figure out how to properly transpile our source without relying on runtime polyfills.

@kachkaev @capellini Thoughts?

[...]

PR #292, opened 2019-05-02
add builtins inline, polyfills for ie11 and higher support

### LINKED ISSUE #290: Babel doesn't polyfill according preset-env

StarpTech:
This PR will fix #290 and include all required pollyfills in order to support browsers with restricted set of ES6/7 features. This is aligned to next.js [strategy](https://github.com/zeit/next.js#browser-support). Next.js is already delivered with `core.js` and builtIns are only imported once as documented [here](https://babeljs.io/docs/en/babel-preset-env#usebuiltins-usage-experimental)

> Adds specific imports for polyfills when they are used in each file. We take advantage of the fact that a bundler will load the same polyfill only once.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -6,8 +6,10 @@
           "@babel/env",
           {
             "modules": false,
+            "useBuiltIns": "usage",
             "targets": {
-              "node": "current"
+              "node": "current",
+              "ie": "11"
             }
           }
         ],
@@ -23,6 +25,13 @@
             "regenerator": true,
             "useESModules": false
           },
+          "preset-env": {
+            "useBuiltIns": "usage",
+            "targets": {
+              "node": "current",
+              "ie": "11"
+            }
+          }
         }]
       ],
       "plugins": [


isaachinman:
@kachkaev Mind reviewing? I remember you had previously found some duplication in webpack outputs. Let's make sure we're not adding any new deps here.

StarpTech:
Sure, go on!

isaachinman:
@StarpTech Mind linking those NextJs docs?

StarpTech:
Which docs links do you mean?

isaachinman:
Was hoping you had documentation somewhere that the above Babel settings will not pull in new `core.js` source. I believe the NextJs Babel preset can be found [here](https://github.com/zeit/next.js/blob/canary/packages/next/build/babel/preset.ts).

StarpTech:
No, according to the babel [docs](https://babeljs.io/docs/en/babel-preset-env#usebuiltins-usage-experimental) "usage" only add specific and required polyfills to the file.

isaachinman:
That's correct, but our main concern is not pulling in any more deps, as `next-i18next` is already a pretty sizeable dep.

StarpTech:
Then this PR make no sense at all because it will definitly increase size.

kachkaev:
@isaachinman yes tweaking `.babelrc` in the library may result duplicate modules in the apps' bundles as noted in #152.  If simply providing instructions for `≤IE11` in README is not an option, tweaking `.babelrc` should be done via the next.js [plugin](https://github.com/zeit/next-plugins) API rather than via the local `.babelrc` directly. In that case, polyfills will be added only once during app build rather than twice during lib build and then during app build while polyfilling for other parts of the code base.

StarpTech:
@kachkaev I will check it but in your issue it looks differently. In builtins "usage" mode with a single corejs lib provided by nextjs there should be only one corejs file.

> should be done via the next.js plugin API rather than via the local .babelrc directly

Is it possible to polyfill external modules with a plugin? The more I think about it the more we should just recommend the polyfill approach at app level.

kachkaev:
One more thought. Polyfilling i18n is somewhat more challenging than polyfilling many other things because not only the `I18n` object and methods need to be added, but also the locales. Their set depends on the website and in any case, requires quite a lot of extra javascript.

Because Next.js [produces only a single output bundle for all browsers](https://github.com/zeit/next.js/issues/3056), polyfilling i18n for few people on IE11 means requiring _a lot_ of people to download _a lot_ of unneeded JavaScript. To prevent that, I personally use `https://polyfill.io/` as an alternative to changing babel config. 

```jsx
// in _document.tsx
https://cdn.polyfill.io/v2/polyfill.min.js?features=Map,fetch,Intl.~locale.ru,Intl.~locale.en&unknown=polyfill

// inside <Head />
<link rel="preload" href={polyfillUrl} as="script" />

// before <NextScript />
<script src={polyfillUrl} async={false} />
```

This gives IE11 all the extra JS it needs at a cost of an extra zero-sized request for the modern browsers (`polyfill.io` only returns the JavaScript the given User Agent needs as polyfills). This trick enables i18n in IE11, but does not slow down the load time on frequently mobile browsers as bad as an extra 100+KB of useless JS would do. 

The bottom line is that polyfilling I18n and locales is not a one-size-fits-all task. Therefore, I believe that the solution should be described in README rather than attempted automatically. Neither `.babelrc` approach nor `polyfill.io` one are ideal, so they can be presented as options with own pros and cons.

StarpTech:
> Because Next.js produces only a single output bundle for all browsers, polyfilling i18n for few people on IE11 means requiring a lot of people to download a lot of unneeded JavaScript.

We only talking about some es6/7 polyfills like `Array.includes`, `Object.assign` no drama :smile:

I don't trust third party services like `polyfill.io` for none personal projects. They could inject any kind of js and subresource-integrity [can't](https://caniuse.com/#feat=subresource-integrity) be fully supported.

isaachinman:
This is frustrating. Is there truly not a build-time option?

StarpTech:
> This is frustrating. Is there truly not a build-time option?

Sure, the current PR but I need to check if the issue #152 really appears in a fresh project.

isaachinman:
I mean a build-time solution, not pulling in runtime polyfills.

StarpTech:
@isaachinman What about to rewrite some places in ES5 like `array.includes` and do the rest via babel-trasnform-runtime?

https://gist.github.com/CMCDragonkai/7fb6b279bb667f3c194994b2f2ccedae

kachkaev:
> I don't trust third party services like polyfill.io for none personal projects. They could inject any kind of js and subresource-integrity can't be fully supported.

I agree with this statement to a degree and would not use polyfill.io myself in projects that deal with user data in any form. However, for websites that simply serve content like read-only blogs or catalogues, why not? One project requires more security and another one needs to be fast even on slow networks. There is not silver bullet here.

> This is frustrating. Is there truly not a build-time option?

There is, but it would require preloading all locales in the world into a common bundle, which the latest Chrome and Firefox would have to download. Not sure this is what developers would want to discover when auditing bundle size because the website is loading for too long on a slow network.

isaachinman:
> preloading all locales in the world into a common bundle

Sorry, I do not understand your point about locales relating to ES targets. Can you explain further?

kachkaev:
For i18n to work in IE11 (which this discussion is about) we not only have to polyfill I18n browser API, but also load the locale data (pluralization rules, names of months etc.). Without that, the polyfill for IE11 is incomplete.

isaachinman:
I do not understand how names of months relates to ES targets.

kachkaev:
Names of months relate to the final result, being whether a Next.js app with `next-i18next` works in IE11 out of box or not 🤷‍♂️ No matter how ES targets are configured, client-side calls to `Intl` methods like `NumberFormat` / `DateTimeFormat` / `PluralRules` will crash the app in IE11. Basic things like `t('blah')` may work, but it will be still far from saying that `next-i18next` works in IE11 out of box.

StarpTech:
@kachkaev no i18next library relys on `Intl` :confused:

isaachinman:
@StarpTech That link to "Babel-Polyfill vs Babel-Transform-Runtime" seems very sensible. Our current build process is dead-simple Babel transpilation so we'll be able to use plugins like [babel-plugin-array-includes](https://www.npmjs.com/package/babel-plugin-array-includes).

StarpTech:
@isaachinman if you run babel in debug mode with useBultIns and preset-env you will see several other polyfills which are required for IE11. I won't add such a transformation for 1 or 2 ocurrunces replace it with `ìndexOf`

kachkaev:
Talking about non `Intl` polyfills for this plugin, should not `next-i18next` just do the same as in https://github.com/zeit/next.js/tree/canary/examples/with-polyfills? This example is linked from [_Browser Support_](https://github.com/zeit/next.js#browser-support) section in README.

The API for devs would be as for any other Next plugins

```ts
// next.config.js
const withNextI18next = require('next-i18next')

// Option A
module.exports = withNextI18next()

// Option B
const withCss = require("@zeit/next-css");
const withPlugins = require("next-compose-plugins");

module.exports = withPlugins([
  withNextI18next,
  withCss,
])
```

`withNextI18next` would do similar to what's happening in [`example/with-polyfills/next.config.js`](https://github.com/zeit/next.js/blob/canary/examples/with-polyfills/next.config.js). Importantly, babel transformation will take place only once when building the app and not the `next-i18next`. So #152 wont re-occur.

kachkaev:
`withNextI18next` can options, roughly like `withNextI18next({ polyfillIntl: true, polyfillLocales: ['en', 'de']})`. If these are provided, there's no need for `https://cdn.polyfill.io/v2/polyfill.min.js?...", but the bundle size is bigger. Otherwise, people are instructed how to use `polyfill.io` and what pros and cons are.

isaachinman:
@kachkaev The `next-i18next` package doesn't do anything with webpack config - we don't need to follow the pattern that TypeScript/SASS plugins do.

I still do not understand your point about "polyfilling locales".

Is everyone happy if I proceed with `transform-runtime` and `babel-plugin-array-includes`, etc? The goal here is to get our CJS output to run in IE11 and other older browsers from a purely ECMAScript perspective.

kachkaev:
> we don't need to follow the pattern that TypeScript/SASS plugins do

If you want to polyfill things, then it is necessary to to create `withNextI18next()` function, which people would use inside their apps' `next.config.js`. Just changing `.babelrc` inside the `next-i18next` will bloat the bundle size by potentially putting two identical polyfills in it. [`examples/with-polyfills/next.config.js`](https://github.com/zeit/next.js/blob/canary/examples/with-polyfills/next.config.js) shows the right way of polyfilling things.

The reason why changing the plugin's `.babelrc` results duplicates is because its output is a commonjs file with all polyfill functions built in. Say, you've polyfilled `Array.includes`. When Next.js builds the app in which someone else polyfilled `import "core-js/fn/array/includes";` for themselves, this import won't be able to recognize a duplicate to what we have via `next-i18n` already and thus the client-side bundle output will contain two copies of Array.includes. Polyfills are sometimes quite massive, which effects the website load time quite badly. We've been there: https://github.com/isaachinman/next-i18next/issues/152

isaachinman:
I don't intend to polyfill anything.

kachkaev:
Using things like `babel-plugin-array-includes` / `"targets": { "node": "current", "ie": "11"}` is polyfilling AFAIU.

isaachinman:
The idea is that we'd add some Babel plugins to do build-time transformations so that our CJS output is directly consumable by targets like IE11, without adding any runtime polyfills. I'd really rather not get into the whole NextJs build process - I'd rather just export and be done with it.

kachkaev:
I understand the intention, I’m just warning that this approach will leave developers with larger bundle sizes because of duplicate polyfills. Once a library is built, it can no longer be tree-shaken, which is quite  unfortunate, but that’s the reality we have. This limitation is why Next.js suggest doing polyfills the way it does and not via the libraries’ .babelrc.

StarpTech:
The discussion is no longer related to the purpose of this PR. Please open a different issue.

Issue #295, opened 2019-05-03
Get negotiated locale statically server/client-side


haf:
## Is your feature request related to a problem? Please describe.
I'd need the locale for fetching localised content from the blog provider in `static getInitialProps`. `req` is null client-side and `navigator.language`, server-side. Also, I need to take cookies and any other mechanism of lang neg into account. How do I do this?

## Describe the solution you'd like
Being able to see the currently active `language` in the static function.

## Describe alternatives you've considered
- the utils function
- readinig the cookie
- reimplementing lang neg

## Additional context
Would be great to have this in the sample.

isaachinman:
The `withNamespaces` HOC adds a language prop and should be used for this sort of thing.

You should set your cookie preferences via the config. If you find yourself doing imperative stuff with cookies, you've lost a lot of the benefit of this package.

Let me know if you have any other questions.

isaachinman:
Ah, if you need it inside your `static` method you'll need to check `req.lng` or `req.i18n`. Neither of these approaches are recommended or officially supported, though.

haf:
> The withNamespaces HOC adds a language prop and should be used for this sort of thing.

Not statically

> method you'll need to check req.lng or req.i18n

No, that doesn't work on the client

isaachinman:
On the client you can check the actual `i18n` instance I suppose. Can you please explain your use case?

haf:
I'm loading i18n-ed blog posts via; it works both in the browser and on the server https://buttercms.com/docs/api/?javascript#get-your-blog-posts

Some blog posts are in Swedish and therefore not optimal for international consumption, whilst others are in English (all can see) and perhaps going forward, in more languages.

haf:
Current workaround:

```

BlogIndex.getInitialProps = async ({ query, req }: any) => {
  const page = query.page || 1
  const tag_slug = req != null ? req.language : i18next.i18n.language
  const listing = await list({
    page,
    page_size: 10,
    exclude_body: true,
    tag_slug
  })
  return {
    namespacesRequired: [ 'blog', 'common' ],
    ...listing
  }
}
```

However, this doesn't adjust "live" when the language is changed globally, but requires a navigation event away and back again.

isaachinman:
[...]

PR #301, opened 2019-05-07
Update README.md

haf:
Documenting how to find the negotiated language from `req` and `i18next` (the instance) both client/server side. See #295


isaachinman:
I would expect this to explain _why_ we have to switch over client and server contexts, and not say specific things about Accept headers and so on. I appreciate the PR but I'll probably write something up quickly that gives more context/background. Thank you!

isaachinman:
@haf Mind checking #303 and see if this addresses your concerns?

Issue #297, opened 2019-05-04
localePath not working after upgrade to 0.38.0+


vimutti77:
## Describe the bug
After upgrade next-i18next to version 0.38.0+, if I set localePath to `public/locales`. the language will not be loaded on the server side.

## Occurs in next-i18next version
0.38.0 - 0.39.0

## Steps to reproduce
i18n.js
```
const NextI18Next = require('next-i18next/dist/commonjs')

module.exports = new NextI18Next({
  defaultLanguage: 'th',
  ignoreRoutes: ['/_next', '/static', '/public', '/app', '/icons', '/images'],
  otherLanguages: ['en'],
  localePath: 'public/locales',
  localeSubpaths: 'all',
})
```
My project structure
```
.
├── public
|     └── locales
└── src
      └── server.js
```

## Expected behaviour
The language are loaded correctly on server side in version 0.38.0.

## OS (please complete the following information)
 - Device: PC, Windows 10
 - Browser: Chrome lastest

## Additional context
The previous version (0.37.0) is working correctly.

vimutti77:
I found that if I replicate the folder `public/locales` to `static/locales`, the language will be loaded. Look like the server is still load the language from `static/locales` although I set `localePath` to `public/locales`.

novascreen:
I believe this happened here: https://github.com/isaachinman/next-i18next/commit/56363e4e3cf7c89f55675430efe7a251a7207eaa
`config.backend` is now always generated based on the default config. that logic needs to be in `create-config` to take user config into account

isaachinman:
Sorry, I have been away on holiday this weekend. This does indeed sound like a genuine bug and we'll work on getting it resolved this week. Thank you for your patience.

capellini:
@isaachinman After reviewing #140 again, perhaps c0f2a7af68baefe45f66493a006fe5ec0c57295e is a better solution than what we implemented in 56363e4e3cf7c89f55675430efe7a251a7207eaa?

RonnyVo:
Hey guys, can you provide a config for define the url to get the translations on browser?
My scenario is I installed nextjs app in an subdirectory domain, e.g: www.example.com/my-app, the translation in server load OK but then in browser, they will load the translation json in www.example.com/static/locales instead of www.example.com/my-app/static/locales which will result in 404

Issue #294, opened 2019-05-03
Use custom express middleware


haf:
Hello,

I'm trying to use the `i18next-node-locize-backend` middleware, so I can speed up my translating of my site. I've asked the Locize team [in this issue](https://github.com/locize/locize-examples/issues/2#issuecomment-488768926) and I've tried a few alternatives to add their middleware too this module, but currently to no avail. (Details in the linked comment)

I've tried; adding the middleware to the `use` array, and adding an instantiated `Backend` instance to the same, with different error messages.

## Occurs in next-i18next version
`"next-i18next": "^0.37.0"`

## Behaviour

I get an error Error: Invalid URI "/Users/h/dev/qv2/static/locales/en/common.json"

## Screenshots

![image](https://user-images.githubusercontent.com/193115/57158964-062c0f80-6de5-11e9-98dd-7b749caa3786.png)

## Summary

Thanks for a great project; it's working fabulously for me. I hope you can find the time to answer this question.

isaachinman:
Can you check #151 and see if that helps with your use case?

haf:
I read through 151 and it goes through how to configure custom file locations as far as I can tell; it doesn't seem to go through how to instantiate dynamic backends that themselves fetch. This is consistent with the behaviour I'm getting, with an error mentioning the file system rather than anything from the API call.

isaachinman:
@haf Perhaps let's start like this: can you copy/paste a full `react-i18next` configuration that you would expect to work with your desired setup? I can then work backwards to see what it would take for us to support it, if we don't already.

I am going on holiday shortly but know that I am committed to helping users as much as humanly possible - just might take a bit longer.

haf:
```js
import NextI18Next, { defaultConfig } from 'next-i18next'

const locizeConfig = {
  // the id of your locize project
  projectId: 'CHANGEME',
  // add an api key if you want to send missing keys
  apiKey: 'CHANGEME',
  // the reference language of your project
  referenceLng: 'en',
  saveMissing: true,
  // version - defaults to latest
  version: '2.0.0',
  // private - set to true if you version on locize is set to use private publish
  private: false
}

// https://github.com/i18next/react-i18next/tree/master/example/v9.x.x/locize
const defaultConfig = {
  defaultLanguage: 'en',
  otherLanguages: ['sv'],
  use: process.browser ? [] : [ require('i18next-node-locize-backend') ],
  backend: locizeConfig
}

[...]

PR #302, opened 2019-05-07
refactor: Add localeExtension config option and clean up config procedure

### LINKED ISSUE #257: Add YAML Support
The backend supports YAML but the loader is hard coded with a JSON extension and it is not possible to override as written right now.

https://github.com/isaachinman/next-i18next/blob/c15256e16070b101467ab21be51d0170aad7bf11/src/config/create-config.js#L42-L54

If we want to keep the `backend` options as rigid as they are, we could expose a `fileExtension` option and use that.

If it is possible to use YAML with this library can you please show an example?  I know [you've already mentioned](https://github.com/isaachinman/next-i18next/issues/106#issuecomment-455552249) that [`i18next-node-fs-backend` already supports YAML](https://www.npmjs.com/package/i18next-node-fs-backend), but I can't figure out how to get it working without doing more work than I feel I should have to when using this library.

isaachinman:
General cleanup of the config creation procedure.

Fixes #257, #294, and #297.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,7 @@ server.get('*', (req, res) => handle(req, res))
 | `defaultLanguage`  | `'en'`  |
 | `ignoreRoutes`  | `['/_next', '/static']`  |
 | `otherLanguages` (required) | `[]`  |
+| `localeExtension` | `'json'`  |
 | `localePath` | `'static/locales'`  |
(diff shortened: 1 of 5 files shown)


isaachinman:
@capellini Mind reviewing? Note that I had less than a half hour free this evening so haven't had a chance to write tests for the new config option. Appreciate any help you can offer. Thanks!

PR #303, opened 2019-05-08
docs: Add section on accessing current language

isaachinman:


README.md
```diff
@@ -195,6 +195,21 @@ server.get('/products/:id', (req, res) => {
...
+
+```jsx
+// This is our initialised `NextI18Next` instance
+import { i18n } from '../i18n'
+
+MyPage.getInitialProps = async({ req }) => {
+  const currentLanguage = req === null ? i18n.language : req.language
+}
+```
+
```
haf: Doesn't address changing the language though.

PR #306, opened 2019-05-08
feat: Introduce option to disable shallow routing

### LINKED ISSUE #305: Support for unshallow routing when language has changed
Hi,
it would be great if we could provide an option to allow a reload of `getInitialProps` of the active page. Currently, [shallow routing](https://github.com/isaachinman/next-i18next/blob/6f69e7c62dc230cb087e3c3d6749a9858ab053c7/src/hocs/app-with-translation.js#L27) is used.

The problem I'm trying to solve is really common. I have static translation which works great but my product data is fetched from an external source. How should I fix it without to trigger a real browser reload? 

This is also useful if you use `routeChangeComplete` for analytics. Right know, `routeChangeComplete` has to be fired twice if you want to trigger a refetch of `getInitialProps` after a language change. 

Related:
https://github.com/isaachinman/next-i18next/issues/299

### Proposal:

```js
  shallowRouteChange: false
})
```

StarpTech:
This PR will Fix #305

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,7 +239,7 @@ MyPage.getInitialProps = async({ req }) => {
 | `serverLanguageDetection` | `true`  |
 | `use` (for plugins) | `[]`  |
 | `customDetectors` | `[]`  |
-| `shallowRouteChange` | `false`  |
+| `shallowRouteChange` | `true`  |
 
 _This table contains options which are specific to next-i18next. All other [i18next options](https://www.i18next.com/overview/configuration-options) can be passed in as well._
 
--- a/README.md
+++ b/README.md
@@ -195,6 +195,20 @@ server.get('/products/:id', (req, res) => {
 server.get('*', (req, res) => handle(req, res))
 ```
 
+### Reload `getInitialProps` on language change
+
+In some cases, you'll want to refetch your page content in a different language. The default behaviour won't trigger a reload of `getInitialProps`. This is called [shallow](https://github.com/zeit/next.js#shallow-routing) routing.
+
+```jsx
+import NextI18Next from 'next-i18next'
+
+const NextI18NextInstance = new NextI18Next({
+  shallowRouteChange: false
+})
+
+export default NextI18NextInstance
+```
+
 ## Accessing the Current Language
 
 In many cases, you'll need to know the currently active language. Most of the time, to accomplish this, you should use the `withNamespaces` HOC, which will pass a `lng` prop to the wrapped component and further asserts your component will get re-rendered on language change or changes to the translation catalog itself (loaded translations). More info can be found [here](https://react.i18next.com/legacy-v9/withnamespaces).
@@ -225,6 +239,7 @@ MyPage.getInitialProps = async({ req }) => {
 | `serverLanguageDetection` | `true`  |
 | `use` (for plugins) | `[]`  |
 | `customDetectors` | `[]`  |
+| `shallowRouteChange` | `false`  |
 
 _This table contains options which are specific to next-i18next. All other [i18next options](https://www.i18next.com/overview/configuration-options) can be passed in as well._
 
--- a/src/config/default-config.js
+++ b/src/config/default-config.js
@@ -17,6 +17,7 @@ const LOCALE_SUBPATHS = localeSubpathOptions.NONE
 const config = {
   defaultLanguage: DEFAULT_LANGUAGE,
   otherLanguages: OTHER_LANGUAGES,
+  shallowRouteChange: true,
   load: 'currentOnly',
   localePath: LOCALE_PATH,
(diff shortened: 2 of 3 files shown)


isaachinman:
Hi @StarpTech - did you test this PR? I believe to accomplish what you want, we will need to modify the `app-with-translation` HOC to wrap `WrappedComponent` in `withNamespaces` inside the i18next provider.

The `shallow` option you changed here only applies to a small subset of `localeSubpath` cases.

StarpTech:
Yes, I tested it and it works and all tests are green. What was the reason to close it?

isaachinman:
It doesn't accomplish what it set out to, unless I misunderstand. You'd like `_app.js` to fully re-render when a language is changed, is that correct?

Perhaps I've misunderstood and we can reopen if you explain the purpose, but as I said, the line you changed only affects a very specific case with `localeSubpaths`. What about all other cases?

Also - any new features would need to be accompanied with _new_ test coverage.

StarpTech:
> You'd like _app.js to fully re-render when a language is changed, is that correct?

Yes

> Perhaps I've misunderstood and we can reopen if you explain the purpose, but as I said, the line you changed only affects a very specific case with localeSubpaths. What about all other cases?

Which other place have you in mind? There is only one `languageChanged``listener and only one place where the route is pushed.

> Also - any new features would need to be accompanied with new test coverage.

That's no problem. I will address it when it has a go.

StarpTech:
The problem I'm trying to solve is really common. I have static translation which works great but my product data is fetched from an external source. How should I fix it without to trigger a real browser reload?

isaachinman:
I understand the problem. I will work on it in due time. It is not as simple as a find and replace for route push.

StarpTech:
Thank you, feel free to request for help.

isaachinman:
Something [along these lines](https://github.com/isaachinman/next-i18next/compare/replace-route-on-lang-change) would work, however I'm not sure if calling `router.replace` is an anti-pattern.

isaachinman:
@timneutkens Do you have an opinion on this?

StarpTech:
@isaachinman this shouldn't work in 'foreign' mode

isaachinman:
@StarpTech It is not a full solution, but a proof of concept. I'm not sure I'd like to proceed down this road. It's easy enough for users to hook into `i18n.on` events themselves.

StarpTech:
> @StarpTech It is not a full solution, but a proof of concept. I'm not sure I'd like to proceed down this road.

@isaachinman I really don't understand you. The only thing I propose to change is how `getInitialProps``is handled by next.js. The functionality should work as it is because we use HOC. 

> It's easy enough for users to hook into i18n.on events themselves.

That's not true! How is it possible? I can't intercept the registered listener by next-i18next it will result in a double route change.

isaachinman:
The listener registered by `next-i18next` performs a `router.replace` _if and only if the route changes_. What you are describing is triggering NextJs router events when the route _has not changed_. I hope this clarifies the matter.

StarpTech:
> The listener registered by next-i18next performs a router.replace if and only if the route changes.

A route change happens every time when the language is changed due to the lcoale change in the url.

> What you are describing is triggering NextJs router events when the route has not changed. I hope this clarifies the matter.

No, I just don't want to use `shallow` routing in 'foreign' mode in order to update my complete page. Is it clear now? There is no option to hook into i18n.on events when I want to prevent the double route event as described in my bug ticket.

isaachinman:
> A route change happens every time when the language is changed due to the lcoale change in the url.

Nope. This package supports more than one way of working with languages/paths.

> No, I just don't want to use shallow routing in 'foreign' mode

As I've said, this is _one small subset of `localeSubpath` cases_, and we should take a more general approach.

StarpTech:
I don't want to change the behavior of this module except the way how a route push/replace action is handled by nextjs. It's hard to understand your points. What do you mean with more general? What are you trying to solve? Where is the issue? On order to push it forwards you have to be more transparent.

isaachinman:
The point is that users are doing data fetching and external dependency reconciliation based on current locale inside the `getInitialProps` hook of their `_app`, and are now expecting that hook to re-run when the locale changes.

If we only re-run `getInitialProps` when route changes, we would not be supporting this functionality for users who are not using `localeSubpaths`.

We aren't going to be merging any half-measures that support some users and not others.

@StarpTech I hope this is clear, but if not please leave it with me, as the request itself is clear.

StarpTech:
I'm trying to understand your doubts. The only thing I propose is to enable this feature behind a flag. I can't speak for more than myself but I think the scenario to reload your page content based on a different locale is really common. It would also help to enable it selectivly per `changeLanguage` call.

PR #308, opened 2019-05-09
fix: Do not call next() after redirect

### LINKED ISSUE #304: The response is not ended after redirected to the new Url
## Describe the bug
```
server.use(nextI18NextMiddleware(nextI18next));
...
server.get('/', (req, res) => {
res.setHeader('Content-Type', 'text/html');
...
})
```
The above code will report ```UnhandledPromiseRejectionWarning: Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client```

v0.37.0

## Steps to reproduce
https://github.com/Ktchan825/next-i18next
```
git clone https://github.com/Ktchan825/next-i18next
cd examples/simple
yarn && yarn dev
```
In your browser, navigate http://localhost:3000/da or any page not exist
In your server console, You can see res.headersSent is true 

## Expected behaviour
I don't expect a middleware will send the response header to the client, such that I cant use res.setHeader in rest of my route.

 - Device: Mac mini
 - Browser: Version 74.0.3729.131 (Official Build) (64-bit)

capellini:
If we perform a redirect in lngPathDetector (using localeSubpaths option), we do not want any subsequent processing of routes or middleware.  Thus, next() should not be called in our middleware if we redirect.  This change ensures that we do not process any further middleware or routes if we redirect in lngPathDetector.

Appropriate changes in unit and e2e tests have been made to ensure no regression occurs.

Additionally, documentation changes have been made to make it explicit that we may perform redirects within our middleware and, thus, the order in which you run our middleware is important.

Resolves #304.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,8 @@ After creating and exporting your `NextI18Next` instance, you need to take the f
 1. Create an `_app.js` file inside your `pages` directory, and wrap it with the `NextI18Next.appWithTranslation` higher order component (HOC). You can see this approach in the [examples/simple/pages/_app.js](./examples/simple/pages/_app.js). 
 2. Create a `server.js` file inside your root directory, initialise an [express](https://www.npmjs.com/package/express) server, and use the `nextI18NextMiddleware` middleware with your `nextI18Next` instance passed in. You can see this approach in the [examples/simple/server.js](./examples/simple/server.js). For more info, see [the NextJs section on custom servers](https://github.com/zeit/next.js#custom-server-and-routing).
 
+    When using the `localeSubpaths` option, our middleware may redirect without calling any subsequent middleware.  Therefore, if there are any critical middleware that must run before this redirect, ensure that you place it before the `nextI18NextMiddleware` middleware.
+
 That's it! Your app is ready to go. You can now use the `NextI18Next.withNamespaces` HOC to make your components or pages translatable, based on namespaces:
 
 ```jsx
@@ -134,6 +136,8 @@ myapp.com/fr/
 myapp.com/de/
 myapp.com/es/
 ```
+
+When using the localeSubpaths option, our middleware may redirect without calling any subsequent middleware.  Therefore, if there are any critical middleware that must run before this redirect, ensure that you place it before the `nextI18NextMiddleware` middleware.
 
 The main "gotcha" with locale subpaths is routing. We want to be able to route to "naked" routes, and not have to worry about the locale subpath part of the route:
 
--- a/README.md
+++ b/README.md
@@ -69,8 +69,6 @@ After creating and exporting your `NextI18Next` instance, you need to take the f
 1. Create an `_app.js` file inside your `pages` directory, and wrap it with the `NextI18Next.appWithTranslation` higher order component (HOC). You can see this approach in the [examples/simple/pages/_app.js](./examples/simple/pages/_app.js). 
 2. Create a `server.js` file inside your root directory, initialise an [express](https://www.npmjs.com/package/express) server, and use the `nextI18NextMiddleware` middleware with your `nextI18Next` instance passed in. You can see this approach in the [examples/simple/server.js](./examples/simple/server.js). For more info, see [the NextJs section on custom servers](https://github.com/zeit/next.js#custom-server-and-routing).
 
-    When using the `localeSubpaths` option, our middleware may redirect without calling any subsequent middleware.  Therefore, if there are any critical middleware that must run before this redirect, ensure that you place it before the `nextI18NextMiddleware` middleware.
-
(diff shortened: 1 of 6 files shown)


capellini:
@isaachinman A few notes:

1. I have opted to use a return value to indicate whether or not a redirect has been performed in `lngPathDetector`, which is nice and simple, but I know is a style that some do not like.  If you would prefer a different solution (e.g. throwing an error and catching it), please let me know and I will change the code appropriately.
2. I have modified our simple e2e test to incorporate calling `res.setHeader` to avoid regression bugs, but I know that you also want to keep our example repos as simple as possible.  If you would prefer that I remove that line, please let me know.
3. I have included the disclaimer about middleware ordering in two places where I think that it may be appropriate.  If you think it's too redundant to include it in both of those places, just let me know which one you would like me to remove.

isaachinman:
Initial thought is that this doesn't seem like the cleanest pattern to me. Nor would throwing an error - it's not an error.

It seems like our `next-i18next-middleware` is fundamentally doing three things:

1. Forcing trailing slash (will be removed at some point)
2. Performing language path redirects via `lngPathDetector`
3. Doing `req.url.replace` if `params !== false`

Would it be possible to split these into three separate middlewares and chain them? We'd then have finer control over whether or not to call `next` after each of the three steps.

Let me know if that makes sense and if you think it sounds reasonable.

capellini:
> Would it be possible to split these into three separate middlewares and chain them? We'd then have finer control over whether or not to call `next` after each of the three steps.

I'll take a look and see what I can do.

isaachinman:
@capellini Any update?

capellini:
@isaachinman I have split up the middleware into the three distinct units, as requested.  I'm still using a return value to determine whether or not to call `next()`.  Please let me know what you think.

src/utils/lng-path-detector.js
```diff
@@ -3,6 +3,8 @@ import redirectWithoutCache from './redirect-without-cache'
 import { localeSubpathOptions } from '../config/default-config'
 
 export default (req, res) => {
+  let redirectPerformed = false
```
isaachinman: I think I'd prefer if this function returns some known config object, like:

```
interface LngPathDetectorRedirectConfig {
    originalUrl: string;
    correctedUrl: string;
    redirectRequired: boolean;
}
```

And then we can perform redirects in the consumer middleware instead.

Sound alright? I'm happy to make these changes if you don't have the time at the moment. Thanks!

capellini: :+1: Sure thing.  I'll work on this at some point soon.

capellini: @isaachinman After reviewing some code climate suggestions, I occurred to me that `redirectRequired` is not strictly necessary, since we can always compare `originalUrl` and `correctedUrl` to determine if a redirect is required.  If you would prefer I add it back in, just let me know.

capellini:
@isaachinman Other than potentially adding back in `redirectRequired` as discussed in https://github.com/isaachinman/next-i18next/pull/308#discussion_r287011881, this should be ready.  Please let me know if you'd like me to add `redirectRequired` back in and I will.

README.md
```diff
@@ -69,6 +69,8 @@ After creating and exporting your `NextI18Next` instance, you need to take the f
 1. Create an `_app.js` file inside your `pages` directory, and wrap it with the `NextI18Next.appWithTranslation` higher order component (HOC). You can see this approach in the [examples/simple/pages/_app.js](./examples/simple/pages/_app.js). 
 2. Create a `server.js` file inside your root directory, initialise an [express](https://www.npmjs.com/package/express) server, and use the `nextI18NextMiddleware` middleware with your `nextI18Next` instance passed in. You can see this approach in the [examples/simple/server.js](./examples/simple/server.js). For more info, see [the NextJs section on custom servers](https://github.com/zeit/next.js#custom-server-and-routing).
 
+    When using the `localeSubpaths` option, our middleware may redirect without calling any subsequent middleware.  Therefore, if there are any critical middleware that must run before this redirect, ensure that you place it before the `nextI18NextMiddleware` middleware.
```
isaachinman: Why is this block of text duplicated?

examples/simple/server.js
```diff
@@ -14,7 +14,13 @@ const handle = app.getRequestHandler();
 
   server.use(nextI18NextMiddleware(nextI18next))
 
-  server.get('*', (req, res) => handle(req, res))
+  server.get('*', (req, res) => {
```
isaachinman: Is it possible to not have this in the example? A lot of people do straight copy/pastes of this dir and I'd rather not have this floating around.

capellini: Done

README.md
capellini: Removed

PR #309, opened 2019-05-10
refactor: Added the TypeScript declaration file for the middleware module

d-dmytro:
In order to setup _next-i18next_, we need to import the package's Express middleware creator function `import nextI18NextMiddleware from 'next-i18next/middleware'` and add the middleware to the express server: ` server.use(nextI18NextMiddleware(nextI18Next));` The middleware module misses the type definitions and TypeScript treats this import as `any`, so I added the definitions in the _middleware.d.ts_

### DIFF
--- a/middleware.d.ts
+++ b/middleware.d.ts
@@ -1,0 +1,6 @@
+import NextI18Next from './types';
+import { Handler } from 'express';
+
+declare function nextI18NextMiddleware(nexti18next: NextI18Next): Handler;
+
+export default nextI18NextMiddleware;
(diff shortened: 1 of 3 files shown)


isaachinman:
Needs a semantic title.

@lucasfeliciano Is this something you would be able to review? I believe you have a TS consumer of this package?

isaachinman:
You've added a dep but not updated the lockfile. Please do so.

isaachinman:
Otherwise looks good to me.

PR #312, opened 2019-05-14
fix: Load main language when initial language is locale specific

lucasfeliciano:
In order to support [locals resolving](https://www.i18next.com/principles/fallback#locals-resolving) feature from i18next we need to load the main language of the locale-specific initial language.

Here is the use case:

The website supports the `fr` language however there is a legal related copy that only apply to France.

By creating a folder `/locales/fr-FR` we can add these locale-specific copies in there and make i18next fall back to the general `/locales/fr` language for the keys that it doesn't find on the specific folder.

ps: I implemented this in a way that doesn't impact the current usage. So if you only use global languages ( `fr, en, de, etc..`) nothing changes.

### DIFF
--- a/__tests__/utils/lngs-to-load.test.js
+++ b/__tests__/utils/lngs-to-load.test.js
@@ -47,4 +47,8 @@ describe('lngsToLoad utility function', () => {
     expect(lngsToLoad(undefined, { it: ['dk', 'en'] })).toMatchObject([])
   })
 
+  it('returns main language when initialLng is locale specific', () => {
+    const analysis = lngsToLoad('fr-FR', 'en', ['fr', 'nl', 'de', 'fr-BE', 'en-US'])
+    expect(analysis).toMatchObject(['fr-FR', 'en', 'fr'])
+  })
 })
--- a/src/hocs/app-with-translation.js
+++ b/src/hocs/app-with-translation.js
@@ -85,7 +85,7 @@ export default function (WrappedComponent) {
 
         // Detect the languages to load based upon the fallbackLng configuration
         const { fallbackLng } = config
-        const languagesToLoad = lngsToLoad(initialLanguage, fallbackLng)
+        const languagesToLoad = lngsToLoad(initialLanguage, fallbackLng, config.otherLanguages)
 
         // Initialise the store with the languagesToLoad and
         // necessary namespaces needed to render this specific tree
--- a/src/utils/lngs-to-load.js
+++ b/src/utils/lngs-to-load.js
@@ -1,4 +1,4 @@
-export default (initialLng, fallbackLng) => {
+export default (initialLng, fallbackLng, otherLanguages) => {
   const languages = []
 
   if (initialLng) {
@@ -23,6 +23,15 @@ export default (initialLng, fallbackLng) => {
     if (fallbackLng.default) {
       languages.push(fallbackLng.default)
     }
+  }
+
+  if (initialLng && initialLng.includes('-') && Array.isArray(otherLanguages)) {
+    const [languageFromLocale] = initialLng.split('-')
+    otherLanguages.forEach((otherLanguage) => {
+      if (otherLanguage === languageFromLocale) {
+        languages.push(otherLanguage)
+      }
+    })
   }
 
   return languages


isaachinman:
Hey @lucasfeliciano, nice to hear from you. Are you using regional locales with `next-i18next` somewhere in production? I was under the assumption that this simply didn't work.

I think the changes required to properly support regional locales would be much more involved in this, and include considerations like `localeSubpaths`, etc.

lucasfeliciano:
Hey @isaachinman!
It's not in production yet, however, I'm working on this now and it is working ~properly~.

The only problem that I found was that once the javascript takes over the server side rendered document ( which is translated correctly ) it replaces the keys that are missing from the locale-specific translations with the fallback language since the main language is not loaded therefor is not served together with the initial store.

This pr changes this behaviour and add the necessary languages. And it is working like a charm, at least for my requirements ( we don't use `localeSubpaths` ), however, I don't understand why the problem with `localeSubpaths + regional locales`. What do experience?

isaachinman:
The main problem with this is that we do not explicitly support regional locales at the moment. This PR would add "halfway" support, but would lead to future problems.

What if locale subpaths ended up looking like this:

```
myapp.com/en-us
myapp.com/en-gb
```

If that were the case, we specifically _wouldn't_ want to load in multiple locales. I think the general misunderstanding here is that a regional locale is some sort of subset of a language-generic locale. In my understanding, that is not true at all - a locale is a distinct entity.

To achieve the behaviour you want, you'd either need to:

1. Maintain regional locales as "full locales"
2. Switch over your fallback language so that `en-GB` falls back to `en` (for example)

Let's keep discussing and figure out a solution!

lucasfeliciano:
>  I think the general misunderstanding here is that a regional locale is some sort of subset of a language-generic locale.

I don't agree with that, most of the cases regional locale is used to represent a regional way of expression or grammar.
The simplest example is English from GB and US with words like `localise` and `localize`.
If one is not a subset of the other we will have to maintain a lot of duplicated translations keys just to support a few translation keys that have these differences.


I think this is a broader discussion that have general discussion and not related to the implementation.

https://www.i18next.com/overview/configuration-options#languages-namespaces-resources

lucasfeliciano:
>  I think the general misunderstanding here is that a regional locale is some sort of subset of a language-generic locale.

I don't agree with that, most of the cases regional locale is used to represent a regional way of expression or grammar.
The simplest example is English from GB and US with words like `localise` and `localize`.
If one is not a subset of the other we will have to maintain a lot of duplicated translations keys just to support a few translation keys that have these differences.


I think this is a broader discussion that has an impact further than this project so we need to follow standards in this matter.


> If that were the case, we specifically wouldn't want to load in multiple locales.

We don't have to i18next has an option to let the user decide what he wants to load.

We can pass the `load` option and i18next will take care of it. This is working already and the default is `all` so it **will** load all the regional and general language by default.

More details here:
https://www.i18next.com/overview/configuration-options#languages-namespaces-resources


And here is an example of the config that I'm passing when instantiating `next-i18next`

```js
{
  defaultLanguage: 'en',
  fallbackLng: 'en',
  otherLanguages: ['nl', 'de', 'fr', 'fr-FR', 'hu', 'es'],
  load: 'all',
  serverLanguageDetection: true,
  customDetectors: [languageByDomainDetector],
  localeSubpaths: 'none',
  detection: {
    order: ['querystring', 'domain'],
  },
}
```

It works as expected and like a charm :)

isaachinman:
Yup, that'll work because you're passing `load: 'all'`. Unfortunately, regional locales do directly affect our implementation of `localeSubpaths` and I'm not really sure what to do about that.

In the end, it might be easier to require users to simply pass in a mapping of locales to routes, eg:

```
{
    defaultLanguage: 'en',
    fallbackLng: 'en',
    otherLanguages: ['nl', 'de', 'fr', 'fr-FR'],
    localeSubpaths: {
        'nl': 'nl',
        'de': 'de',
        'fr': 'fr'
        'fr-FR': 'custom-france-subpath'
    }
}
```

I think we can merge this PR as a short term measure so that you're not blocked, but I do think we need a more holistic approach in general to locales and subpaths.

lucasfeliciano:
Thanks!

I have some ideas on how to make it work with subpaths I'll create an issue after react europe with some proposals.

isaachinman:
> I'll create an issue after react europe with some proposals and ideas

That would be great - thanks. Enjoy!

PR #323, opened 2019-05-27
Adds useSuspense:false to react-i18next default config

MiguelFreire:
As stated in the migration guide from (https://react.i18next.com/latest/migrating-v9-to-v10) this commit sets useSuspense to false in react-i18next default config.

### DIFF
--- a/src/config/default-config.js
+++ b/src/config/default-config.js
@@ -41,6 +41,7 @@ const config = {
   },
   react: {
     wait: true,
+    useSuspense: false,
   },
   strictMode: true,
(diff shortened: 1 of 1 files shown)


MiguelFreire:
@isaachinman both the local tests and Circle CI are using the linked version of the package from npm registry, shouldn't it be using the local dist? Or am I missing something?

isaachinman:
I don't understand your point. The `useSuspense ` option is new in v10 and as far as I know would need to be accompanied by a full migration - it's not useful on its own.

PR #327, opened 2019-05-30
add a config for setting the dir for nextjs project

thientanngo:
Hello, I found a problem when using your code with locating a dir for NextJS project. In this case, the localePath in server and client should be different. Otherwise, when I change a language, the browser cannot get JSON file.

### DIFF
--- a/src/config/create-config.js
+++ b/src/config/create-config.js
@@ -25,6 +25,7 @@ export default (userConfig) => {
     localeExtension,
     localePath,
     localeStructure,
+    nextjsDir,
   } = combinedConfig
 
   if (isServer) {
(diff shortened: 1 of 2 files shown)


isaachinman:
I assume what you are referring to is running your app under a `/subpath`? If so, this has already been discussed in previous issues - please search for that.

thientanngo:
it's not a problem about subpath. You should try to run your next js project in a dir. Like that
https://github.com/zeit/next.js/#custom-server-and-routing
`Supported options:

dev (bool) whether to launch Next.js in dev mode - default false
dir (string) where the Next project is located - default '.'`

isaachinman:
Are you saying you are running your scripts from a different dir than your root dir? I do not understand the point being made here.

thientanngo:
Yes, in my project, `static` folder is in `web` folder, so I set `localePath: 'web/static/locales'`, when running the project, I have a 404 error on `http://localhost:3030/web/static/locales/en/common.json`, the correct link is `http://localhost:3030/static/locales/en/common.json`

PR #344, opened 2019-06-14
feat: Update types.d.ts with useTranslation

Jonesus:
This PR adds the typings of `useTranslation` to the `types.d.ts` for better dev experience and typescript support :)

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import {
+  useTranslation,
   TransProps,
   Namespace,
   NamespaceExtractor,
@@ -38,6 +39,7 @@ declare class NextI18Next {
   Router: SingletonRouter;
   i18n: i18next.i18n;
   config: Config;
+  useTranslation: typeof useTranslation;
 
   constructor(config: InitConfig);
 


types.d.ts
```diff
@@ -38,6 +39,7 @@ declare class NextI18Next {
   Router: SingletonRouter;
   i18n: i18next.i18n;
   config: Config;
+  useTranslation: typeof useTranslation;
```
isaachinman: Is `typeof` the right thing to do here? Does `react-i18next` not have actual typings for this function?

Jonesus: It does have the types, I wasn't sure if it would be better to just create the function type signature from the other types in `react-i18next` or use `typeof`, my reasoning is that if the type of the `react-i18n`'s `useTranslation` would change in the future there would be no need to update your package other than the dependency

isaachinman: @Jonesus Have you confirmed that the output is what you expect? When I clone your repo and checkout this branch, the output I get for `new NextI18Next({}).useTranslation` is `any`.

Am I misunderstanding?

Jonesus: Hmm how do you get that error? It works perfectly for me in VSCode with TypeScript 3.5.1 at least...

isaachinman: Not an error - I mean that the type is `any`. Can you screenshot your output and upload it here?

Jonesus: Not sure how to run that oneliner, but this is what VSCode shows on hover:
yah

isaachinman:
@Jonesus Good enough for me - thanks. Please just give the PR a semantic title.

Jonesus:
@isaachinman done (I hope 😅 ) Thanks!

Issue #328, opened 2019-05-31
localePath with src directory makes locale file request 404


anhdd-hnl:
## Describe the bug
`localePath` inside `src` dir makes client request for locales file returns 404 due to incorrect path

## Screenshots
My project structure:
![pj-structure](https://user-images.githubusercontent.com/37301129/58700338-f09d0c00-83c9-11e9-8683-6c348d85f148.jpg)

My npm scripts:
```json
"scripts": {
    "test": "echo \"Error: no test specified\" && exit 1",
    "dev": "node ./src/server.js",
    "build": "next build ./src",
    "start": "next start ./src"
  }
```

My i18n.js:
```
const NextI18Next = require('next-i18next/dist/commonjs')

module.exports = new NextI18Next({
  defaultLanguage: 'vi',
  otherLanguages: ['en'],
  localePath: 'src/static/locales'
});
```

Bug:
- Client navigation for locales file returns 404 due to incorrect path:
![incorrect_path](https://user-images.githubusercontent.com/37301129/58700331-ea0e9480-83c9-11e9-81d3-00d8221987f7.png)

kachkaev:
I also noticed this issue after upgrading. Solved like this in the end: https://gitlab.com/kachkaev/website-frontend/blob/master/src/i18n.ts#L54

isaachinman:
@kachkaev Do you think this is a regression? Is this potentially what #327 is about?

Can anyone identify where this regression was introduced?

brosbunbo:
@kachkaev It works for me, thanks alot.

kachkaev:
> Do you think this is a regression?
I first thought it was, but then decided it was me not using the library as documented. Hard to say now which exact version introduced the change.

isaachinman:
> Hard to say now which exact version introduced the change

Do you mind installing some previous versions and finding out, though? It sounds like a regression to me.

isaachinman:
This is indeed a regression between v0.37.0 and v0.38.0.

@capellini This is almost certainly a result of 56363e4e3cf7c89f55675430efe7a251a7207eaa.

isaachinman:
After looking through 56363e4, I'm not actually convinced this is a regression. Previously, we were assuming that all users were loading locales from the filesystem, and we were doing some "magic" with `process.cwd`. You'll need to do that manually now if you are specifying a `config.backend` object.

Here's an example from a working, production project on v0.40.1:

```jsx
const NextI18NextInstance = new NextI18Next({
[...]

PR #364, opened 2019-06-25
feat: Validate defaultNS exists via fs

### LINKED ISSUE #358: Server rendered markup is empty if you don't have a `common` namespace.
## Describe the bug
After adding next-i18next to my Next app, I noticed that the HTML source returned from the server didn't contain my app's markup. Specifically, `<div id="__next"></div>` was empty in the markup. When viewed in the browser, it rendered fine because the JS execution would construct the views via DOM manipulation, but I'd be missing out on the benefits of SSR.

After a bit of digging, I traced it back to the [`<NextStaticProvider />` component](https://github.com/isaachinman/next-i18next/blob/master/src/components/NextStaticProvider.js#L9), which does `return tReady ? children : null`.

And `tReady` is provided by [react-i18next's useTranslation hook](https://github.com/i18next/react-i18next/blob/master/src/useTranslation.js#L38-L40), which checks if all relevant namespaces have been loaded.

The issue is that the list of namespaces ends up coming from `i18n.options.defaultNS`, and next-i18next sets `defaultNS` to `common` by default. My app doesn't have a `common` namespace, so `tReady` would never be true during SSR.

I appreciate that you've documented the default values of next-i18next options. However, what was missing for me was that `defaultNS` would determine whether there's server rendered markup or not. In addition, users might not realize that SSR is broken because client-side render works. 

`0.43.0`

## Steps to reproduce
* Create a new next-i18next app as described in the README
* However, instead of having `common.json` in each language directory, have something else, for example `strings.json`
* Start the next dev server, load a page, then view source
* Notice that markup doesn't contain app view code
* Modify i18n config `defaultNS` to be `[]` or `["strings"]`
* Restart dev server, load page, view source
* Notice that markup *does* contain app view code

## Expected behaviour
I'm not really sure the best path forward. If you want to maintain that default value, users should be guided to either use a common.json translation file, or change the `defaultNS` option to suit their project. If you want the project to just work out of the box for all users, then `defaultNS`'s default value should be `[]` instead of `common`. Or perhaps it'd be better to throw if the default namespace doesn't exist in the project. That way the error would be surfaced more obviously.

If anything, this issue can serve as guide for anyone else who runs into this problem.

## Screenshots
If applicable, add screenshots or a GIF to help explain your problem.

 - Device: Mac
 - Browser: Chrome (version won't matter)

isaachinman:
Fixes #358.

### DIFF
--- a/__tests__/config/create-config.test.js
+++ b/__tests__/config/create-config.test.js
@@ -41,6 +41,18 @@ describe('create configuration in non-production environment', () => {
     )
   })
 
+  it('throws if defaultNS does not exist', () => {
+    createConfig = mockIsNodeCreateConfig(true)
+    evalFunc.mockImplementation(() => ({
+      readdirSync: jest.fn().mockImplementation(() => ['universal', 'file1', 'file2']),
+      existsSync: jest.fn().mockImplementation(() => false),
+    }))
+
+    expect(() => createConfig({ localeSubpaths: 'all' })).toThrow(
+      'Default namespace not found at /home/user/static/locales/en/common.json',
+    )
+  })
+
   describe('server-side', () => {
     beforeEach(() => {
       createConfig = mockIsNodeCreateConfig(true)
@@ -86,6 +98,7 @@ describe('create configuration in non-production environment', () => {
     it('creates custom non-production configuration', () => {
       evalFunc.mockImplementation(() => ({
         readdirSync: jest.fn().mockImplementation(() => ['universal', 'file1', 'file2']),
+        existsSync: jest.fn().mockImplementation(() => true),
       }))
 
       const config = createConfig(userConfigServerSide)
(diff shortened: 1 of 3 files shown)


radeno:
This PR breaks custom path for server side locales.
As i said in another Issue, nexti18next should respect `dir` for SSR from NextJS

isaachinman:
@radeno Can you explain what you mean in detail?

radeno:
https://github.com/isaachinman/next-i18next/blob/50276b8f576bfb77399c3a7eb765c0577423aa9f/src/config/create-config.js#L37

Doesn't respect custom settings for `options.backend`

isaachinman:
I do not understand your point, please either explain in detail or submit your own PR with desired changes.

radeno:
Your solution doesn't work anymore
https://github.com/isaachinman/next-i18next/issues/328#issuecomment-499008251 because you are check for defaultNS in same way.

isaachinman:
Sorry, what does `config.backend` have to do with `config.defaultNS`? Feel free to set `defaultNS` to an empty array, or whatever you want.

radeno:
Do you see your code i have posted? You are checking for DefaultNS file before `config.backend` is processed. So it always break, because file doesn't exist.

isaachinman:
If `defaultNS` is a string, we will check the filesystem for it. Feel free to either:

1. Set `defaultNS` to an empty array, or whatever you want
2. Submit an enhancement PR with the functionality you desire

radeno:
And in your comment i have linked, you recommends to make own paths. But that paths are not used.

Issue #369, opened 2019-07-01
Translation files read from disk on every request


bywo:
## Describe the bug
On subsequent requests to the same next.js page with the same locale, i18next reads all translations files for all locales from disk and [copies them (via object spread) to the i18next ResourceStore](https://github.com/i18next/i18next/blob/master/src/ResourceStore.js#L93).

As far as I can tell, this happens because [useSSR replaces the shared ResourceStore's `data` when `initialI18nStore` is passed](https://github.com/i18next/react-i18next/blob/master/src/useSSR.js#L11). Doing so makes sense on the client-side, where we assume the translation bundles have been serialized somehow in the HTML and we want to initialize the client's i18next with those bundles. However, doing this on the server means that any bundles that were loaded on previous requests get thrown away on the next request, so the next request will trigger resource bundle loads again.

Deploying to production showed memory usage growing rapidly, causing lots of GCs, and a CPU pegged at 100%. Stat'ing and reading files from disk for every request didn't help either.

## Occurs in next-i18next version
Latest (0.43.0)

## Steps to reproduce
* Connect Chrome debugger to your server
  * Launch your next.js server that uses `next-18next`, but add `--inspect`, e.g. `node --inspect myServer.js`
  * Go to `chrome://inspect` in Chrome and you should see a process to connect to
* In the "Search" pane of the debugger, search for `addResourceBundle`. A method from the `i18next` npm package should show up. Add a breakpoint at the top of the method.
* Load a page from your next.js service in the browser
* Observe that `addResourceBundle` is hit on every request, and the async stack trace shows `fs.readFile`

Here's what it looks like for me
![Screen Shot 2019-07-01 at 4 12 40 PM](https://user-images.githubusercontent.com/1434481/60471880-64297780-9c1b-11e9-85c7-44cc16984c7d.png)

## Expected behaviour
The first request to the server process should trigger resource bundle loads. Those bundles should then be cached by i18next's ResourceStore. Any subsequent requests should use the cache and shouldn't trigger any file system reads. `addResourceBundle` shouldn't be called again for any bundles that were already added.

## OS (please complete the following information)
 - Device: Mac
 - Browser: Chrome

isaachinman:
Nice find! I wonder what @jamuhl thinks about this. We're specifically using `withSSR` from `react-i18next`. Although there are many ways we could work around this, I imagine it'd be a problem for every consumer of `withSSR`.

jamuhl:
[...]

PR #370, opened 2019-07-01
don't pass initialI18nStore to withSSR on the server-side.

bywo:
* passing initial store on the server causes useSSR (used by
  withSSR) to overwrite the server's global i18next ResourceStore
* frontends will still pass the initial store so rehydration works

resolves https://github.com/isaachinman/next-i18next/issues/369

I don't know if this is the ideal fix, but this has resolved my issues in production. Another potential solution is to have [`useSSR`](https://github.com/i18next/react-i18next/blob/master/src/useSSR.js#L11) merge `initialI18nStore` into `ResourceStore` instead of replacing it.

### DIFF
--- a/src/hocs/app-with-translation.js
+++ b/src/hocs/app-with-translation.js
@@ -133,9 +133,9 @@ export default function (WrappedComponent) {
         >
           <NextStaticProvider>
             <WrappedComponentWithSSR
-              initialLanguage={initialLanguage}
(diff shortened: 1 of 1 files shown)


isaachinman:
How then will localisation happen server side, without a store?

isaachinman:
Closing in favour of #372.

PR #385, opened 2019-07-11
feat: Custom next server dir

radeno:
This is just proof of concept.

If developer set custom server directory with `dir` option https://github.com/zeit/next.js/#custom-server-and-routing nextI18next doesn't accept this path and still checks with `process.cwd()` what is base directory, not directory where NextJS code is located. And `static` directory also.

Because it could be complicated to get this directory from next server, there should be an option to developers to set it without any hard way to rewrite it by own.


isaachinman:
I do not understand what value this adds, and why you can't just prepend that to `localePath`. Please explain?

radeno:
I did explained before.
You are checking for defaultNS hardcoded. What do you don't understand? 
And if it is hardcoded it makes error because defaultNS file doesn't exists.

src/config/create-config.js
```diff
@@ -31,11 +32,12 @@ export default (userConfig) => {
 
     const fs = eval("require('fs')")
     const path = require('path')
+    const nextPath = path.join(process.cwd(), nextServerDir)
 
     // Validate defaultNS
     // https://github.com/isaachinman/next-i18next/issues/358
     if (process.env.NODE_ENV !== 'production' && typeof combinedConfig.defaultNS === 'string') {
-      const defaultNSPath = path.join(process.cwd(), `${localePath}/${defaultLanguage}/${combinedConfig.defaultNS}.${localeExtension}`)
```
radeno: This is one of reason why next-i18next doesn't work with custom with `dir` option. There is no way how to change this variable. So if NextJS directory is for example `src` this validation fails every time.

src/config/create-config.js
```diff
@@ -44,15 +46,15 @@ export default (userConfig) => {
 
     // Set server side backend
     combinedConfig.backend = {
-      loadPath: path.join(process.cwd(), `${localePath}/${localeStructure}.${localeExtension}`),
```
radeno: If there is option to set NextJS custom source directory no need to overwrite this two options.

radeno:
I have added comments to original code why i think PR is necessary.

adascal:
@radeno @isaachinman any updates here? Would love to have that PR merged

isaachinman:
@adascal Want to take a stab at explaining to me what value this PR adds, when you can already modify `localePath` and `defaultNS` to achieve the same result?

radeno:
@isaachinman as i said. You are checking for defaultNS file. So when you define `common` defaultNS you always get error about non-existed defaultNS file.
So, you can't overwrite path. It is hardcoded.

PR #387, opened 2019-07-12
feat: Re-export types I18n and TFunction from i18next

### LINKED ISSUE #375: next-i18next doesn't export types of i18next
## Describe the bug
`next-i18next` doesn't export types of `i18next`

v0.44.0

## Steps to reproduce
1. Install package
2. Try to import any type of `i18next` (TFunction for example)

## Expected behaviour
`next-i18next` exports namespace `i18next` with all types

 - Device: no matter
 - Browser: no matter

masone:
Following up on the change in https://github.com/isaachinman/next-i18next/pull/361 ...

We also consume the types for `I18n` and `TFunction`. We currently have to import the `i18next` package, just to get a hold of these types. The version differences between `next-i18next`'s dependencies and ours regularly cause a type mess. This has also been requested in https://github.com/isaachinman/next-i18next/issues/375

Now that we made the first step towards re-exporting dependencies' types in the PR mentioned above, can we consider exposing these too? The reasoning is the same...

> Re-exporting the typing from `react-i18next` would be cool because direct importing from this package isn't needed anymore. This causes higher updatability.

Fixes #375

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -46,6 +46,8 @@ declare class NextI18Next {
   appWithTranslation<P extends object>(Component: React.ComponentType<P> | React.ElementType<P>): any;
 }
 
+export type TFunction = i18next.TFunction;
+export type I18n = i18next.i18n;
 export type WithTranslation = WithTranslation;
 
 export default NextI18Next;


isaachinman:
I wonder if there's a more flexible way of re-exporting all named exports/types without having to manually handle each one? Otherwise this is fine - I agree that `next-i18next` should be exporting all types necessary to accomplish typical use cases.

masone:
I'm not a typescript expert, but this is the pattern I've seen in a couple of places so far...
Open to suggestions!

isaachinman:
This is certainly fine for now, I'll update it if I think of something more clever. Thanks!

isaachinman:
Can we add test coverage to prevent regressions? Just import the expected type from the root and do something like `expect(MyType).toBeDefined()`.

masone:
That's a cool idea! I _think_ that would mean setting up typescript in the example project. Afaik there is no way to import typescript types in plain javascript. Do you see another option or should I go down that path?

If yes, I might need some help with running the tests. `yarn test` fails for me on master with:
```
Attempted import error: 'useTranslation' is not exported from 'react-i18next'.
```
which I can't follow as `useTranslation` is clearly exported by `react-i18next`. Any advice?

isaachinman:
No, there's no need to do that. We should be able to achieve everything necessary with `ts-jest` and a `transform` on the Jest config. However, please leave this with me as I need to think about the simplest possible approach.

masone:
Thanks, I'm looking forward to seeing your solution! 👀

isaachinman:
@masone Should be as simple as #390 - import of named export will fail during compilation if it's `undefined`. Mind reviewing?

PR #404, opened 2019-07-18
fix: Root path redirect for localeSubpaths

### LINKED ISSUE #401: Redirect in middleware when locale is missing from url
## Describe the bug
Hi, not a real bug report, but more of a question if I'm misunderstanding a feature or I've found a bug. In my understanding, if I configure `localeSubpaths: 'all'` the locale should always be present in the URL. I visit for `https://www.google.be/hello` it should make sure this is rewritten to `https://www.google.be/en/hello`. This currently kinda works but seems to be handled on the client side and not on the server in the middleware.

0.45.0

## Expected behaviour
I would expect the middleware to detect that the current request url is missing the locale and redirect the entire request via a 302 redirect.

I've tracked the issue down to the `lng-path-detector.js` file which seems to be not setting the correct `config.correctedUrl`.
I think the issue is in this section: https://github.com/keustma/next-i18next/blob/master/src/utils/lng-path-detector.js#L35

The `if` statement is entered, the `allLanguages.forEach` does not set the correct url. Before leaving the `if` statement, I added a part to the if-statement which seems to fix the issue.

```javascript
if (!languageChanged && languageNeedsSubpath && !req.url.startsWith("/".concat(language, "/"))) {
      allLanguages.forEach(function (lng) {
        if (req.url.startsWith("/".concat(lng, "/"))) {
          config.correctedUrl = req.url.replace("/".concat(lng, "/"), '/');
          languageChanged = true;
        }
      });

      if(!languageChanged) {
        config.correctedUrl = '/' + language + req.url;
      }
}
```

isaachinman:
Fixes #401.

### DIFF
--- a/src/utils/lng-path-detector.js
+++ b/src/utils/lng-path-detector.js
@@ -11,6 +11,10 @@ export default (req) => {
     const language = lngFromReq(req)
     const { allLanguages, defaultLanguage, localeSubpaths } = req.i18n.options
     let languageChanged = false
+    const languageNeedsSubpath = (localeSubpaths === localeSubpathOptions.FOREIGN
+      && language !== defaultLanguage)
+      || localeSubpaths === localeSubpathOptions.ALL
+
     /*
       If a user has hit a subpath which does not
       match their language, give preference to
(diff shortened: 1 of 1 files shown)


src/utils/lng-path-detector.js
```diff
@@ -22,26 +26,25 @@ export default (req) => {
...
     if (!languageChanged && languageNeedsSubpath && !req.url.startsWith(`/${language}/`)) {
-      allLanguages.forEach((lng) => {
-        if (req.url.startsWith(`/${lng}/`)) {
-          config.correctedUrl = req.url.replace(`/${lng}/`, '/')
+      allLanguages.forEach((otherLng) => {
+        if (req.url.startsWith(`/${otherLng}/`)) {
+          config.correctedUrl = req.url.replace(`/${otherLng}/`, '/')
         }
       })
+      config.correctedUrl = req.url.replace('/', `/${language}/`)
```
keustma: Shouldn't this line be dependent if we already set the `config.correctedUrl` on line 38 ?

isaachinman: No, line 38 here is removing any other potential language subpaths.

keustma: LGTM then

PR #409, opened 2019-07-21
feat: Add module & jsnext:main entries

felixmosh:
This PR reduces the bundle size

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,8 @@
   "repository": "git@github.com:isaachinman/next-i18next.git",
   "author": "Isaac Hinman <isaac@isaachinman.com>",
   "main": "dist/commonjs/index.js",
+  "module": "dist/es/index.js",
+  "jsnext:main": "dist/es/index.js",
   "types": "./types.d.ts",
   "license": "MIT",
   "engines": {


isaachinman:
This does not "reduce the bundle size", it causes Webpack to consume a different target, the es target, which most consumers do not want to do.

We provide the es target for consumers who specifically want to consume esnext code, and are fine with handling their own browser targeting/transpilation.

Feel free to use the es target if you want - no code changes necessary! The current state of JavaScript modules, especially with Webpack and Babel, is a bit messy. We'd rather keep it such that all tooling will use the cjs target by default.

kachkaev:
> which most consumers do not want to do

Why? Just wondering.

IMHO given that this package is a Next.js plugin and not a standalone thing makes ES target the only necessary one at all. The decision to support this or that browser is made at the app level, not at the Next.js plugin level. That's the only way to achieve proper tree-shaking and polyfilling that does not produce duplicate code.

isaachinman:
@kachkaev Unless you're transpiling your `node_modules`, that is unfortunately not the case. Your "app-level" will control the target of your _proprietary_ code, not your deps. Again, this is an ongoing discussion in the JavaScript world, but not something we'll get into here.

Just trying to make the package make the most sense for the most users.

felixmosh:
Ha, Ok, got your point, your `es` target is not only `es modules` but all the code is es.

I think that this is wrong approach to a plugin, I think that the `es` version should only not transpile the import / export syntax.

Checkout how `react-semantic-ui` are doing it: https://github.com/Semantic-Org/Semantic-UI-React/blob/master/.babel-preset.js#L64
Check how `i18next` are doing it: https://github.com/i18next/i18next/blob/master/rollup.config.js#L29

WDYT?

felixmosh:
@isaachinman ping?

isaachinman:
@felixmosh Regarding your comments: there is no need to repeat yourself across multiple issues/PRs. 

The fact that we're exporting a esnext version of our source has nothing to do with alignment with the community in regards to tree shaking. It's there so that consumers who wish to transpile their deps to their own target may do so. This is, I believe, the future of JavaScript dependency management.

The fact that the current standard is to distribute packages with es modules _and also_ transpile to some arbitrary browser target is indicative of how troubled the current state of JavaScript modules is.

As a package author, I _do not care_ about consumers' individual browser targeting needs based on their own particular use cases and business requirements. The distribution of source code and the eventual transpilation to an end user's unique target should be distinct build steps.

The `commonjs` target works relatively nicely, because we can reuse the `next/babel` preset for most of the heavy lifting. Because this is a NextJs plugin, it is absolutely critical that we have the same target. If you can figure out how to replicate the `cjs` Babel target within our `babelrc` to export es modules in a target called `esm` while still consuming the `next/babel` preset, I'd accept a PR for that.

felixmosh:
OK, 
I didn't want to upset you, looked like you didn't get a notification for this comment.

Thanx for the time, I will investigate your suggestion.

isaachinman:
No problem. The relevant preset can be found [here](https://github.com/zeit/next.js/blob/canary/packages/next/build/babel/preset.ts).

felixmosh:
BTW, why do you think that `next-i18next` is transpiled by next's babel config?

isaachinman:
The point is that we need to support targets like IE11. If you'd like to use modern ES and benefit from tree shaking, you can certainly import from the appropriate directory. Is that not working?

NamPNQ:
@isaachinman It doesn't work in case for shared code server and client

PR #411, opened 2019-07-21
export withTranslation as an independent function

### LINKED ISSUE #410: Export `withTranslation` as a function?
In the current implementation, when you want to use `withTranslation`,  you have to import an instance of `NextI18Next` while `WithTranslation` (the _type_) can be directly imported from the library, which means that you inevitably have to import them below:

```javascript
import { withTranslation } from 'path/to/instantiated/nextI18next`
import { WithTranslation } from 'next-i18next'
```

And I suppose it might be more user-friendly if `withTranslation` can also be imported from `next-i18next` as a function.

Export `withTranslation` directly from `react-i18next`.

Hence,

```javascript
import { withTranslation, WithTranslation } from 'next-i18next'
```

without needing to import an instance of NextI18Next or `react-i18next`

I'm totally new around NextI18Next or Next.js, so couldn't figure out a better solution for this yet.

## Additional context

If such change should be made with `withTranslation`, maybe making similar changes with `withRouter`, `Trans`, `Link`, or `i18n` might be related topics but it doesn't look like it's always easy to migrate that way since in some cases they _do_ rely on the instance.

I'll submit my dumbest proposal PR soon, in which I will try not to create a breaking change, i.e., nextI18Next.withTranslation will be left intact, and just export the same function independently.

yuichkun:
## issue
Fixes #410 

This is just a dumb proposal for the issue above.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,6 +1,6 @@
 import 'core-js/stable'
 
-import { withTranslation, useTranslation, Trans } from 'react-i18next'
+import { withTranslation as _withTranslation, useTranslation, Trans } from 'react-i18next'
 import hoistNonReactStatics from 'hoist-non-react-statics'
 
 import createConfig from './config/create-config'
@@ -11,6 +11,8 @@ import { consoleMessage } from './utils'
 import { Link } from './components'
 import { wrapRouter } from './router'
 
+export const withTranslation = (namespace, options) => Component => hoistNonReactStatics(
+  _withTranslation(namespace, options)(Component), Component)
 
 export default class NextI18Next {
 
@@ -28,8 +30,7 @@ export default class NextI18Next {
 
     this.i18n = createI18NextClient(this.config)
     this.appWithTranslation = appWithTranslation.bind(this)
-    this.withTranslation = (namespace, options) => Component => hoistNonReactStatics(
-      withTranslation(namespace, options)(Component), Component)
+    this.withTranslation = withTranslation
 
     const nextI18NextInternals = { config: this.config, i18n: this.i18n }
     this.Link = withInternals(Link, nextI18NextInternals)
--- a/src/index.js
+++ b/src/index.js
@@ -1,7 +1,9 @@
 import { withTranslation, useTranslation, Trans } from 'react-i18next'
 import hoistNonReactStatics from 'hoist-non-react-statics'
+
 import createConfig from './config/create-config'
 import createI18NextClient from './create-i18next-client'
+
 import { appWithTranslation, withInternals } from './hocs'
 import { consoleMessage } from './utils'
 import { Link } from './components'
--- a/src/index.js
+++ b/src/index.js
@@ -1,16 +1,13 @@
-import { withTranslation as _withTranslation, useTranslation, Trans } from 'react-i18next'
+import { withTranslation, useTranslation, Trans } from 'react-i18next'
 import hoistNonReactStatics from 'hoist-non-react-statics'
-
 import createConfig from './config/create-config'
 import createI18NextClient from './create-i18next-client'
-
 import { appWithTranslation, withInternals } from './hocs'
 import { consoleMessage } from './utils'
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -1,6 +1,6 @@
 import 'core-js/stable'
 
-import { withTranslation, useTranslation, Trans } from 'react-i18next'
+import { withTranslation as _withTranslation, useTranslation, Trans } from 'react-i18next'
```
isaachinman: Let's not do anything with underscores like this.

src/index.js
```diff
@@ -11,6 +11,8 @@ import { consoleMessage } from './utils'
 import { Link } from './components'
 import { wrapRouter } from './router'
 
+export const withTranslation = (namespace, options) => Component => hoistNonReactStatics(
```
isaachinman: There's no need to do any hoisting, or anything special at all. `export const withTranslation` can be equal to `this.withTranslation` - it's not dependent on the i18next instance as it will inherit it from context.

types.d.ts
```diff
@@ -49,5 +49,6 @@ declare class NextI18Next {
 export type TFunction = i18next.TFunction;
 export type I18n = i18next.i18n;
 export type WithTranslation = WithTranslation;
+export type withTranslation = typeof withTranslation
```
isaachinman: What is this achieving? Why are you exporting a `const` and a `type` with the same name?

isaachinman:
There's also a conflict to resolve.

src/index.js
yuichkun: OK! I wasn't really sure what the intention of the original hoisting, but if it's not necessary I'll remove this and the underscore dangle as well.

types.d.ts
yuichkun: I don't strongly believe that simply exporting `withTranslation` as a value (function) would also export the type that `react-i18next` exports in the type definition, which looks like below:

```typescript
export function withTranslation(
  ns?: Namespace,
  options?: {
    withRef?: boolean;
  }
): <P extends WithTranslation>(
  component: React.ComponentType<P>,
) => React.ComponentType<Omit<P, keyof WithTranslation>>;
```

src/index.js
```diff
@@ -1,18 +1,13 @@
-import 'core-js/stable'
-
-import { withTranslation as _withTranslation, useTranslation, Trans } from 'react-i18next'
+import { withTranslation, useTranslation, Trans } from 'react-i18next'
```
isaachinman: Please leave the formatting of the imports the same, without removing lines or reordering.

src/index.js
```diff
@@ -30,7 +25,8 @@ export default class NextI18Next {
 
     this.i18n = createI18NextClient(this.config)
     this.appWithTranslation = appWithTranslation.bind(this)
-    this.withTranslation = withTranslation
+    this.withTranslation = (namespace, options) => Component => hoistNonReactStatics(
+      withTranslation(namespace, options)(Component), Component)
```
isaachinman: Why are you hoisting? We _do not need to change the current implementation_.

yuichkun: Wait, I'm confused...
This _is_ the [current implementation](https://github.com/isaachinman/next-i18next/blob/master/src/index.js#L29-L30). I just put it back to the original state.

isaachinman: My mistake! Please disregard.

src/index.js
yuichkun: oops, sorry! An honest mistake.

yuichkun:
Sorry for the messy commits. I can (or you can) squash commits when all is fine.

isaachinman:
No problem, we always squash merge!

types.d.ts
isaachinman: One last thing I forgot - this new type needs to be added to `__tests__/types.test.ts`.

yuichkun: Gotcha!

PR #419, opened 2019-07-24
fix: Do not call router events in SSR

isaachinman:


lavir:
Works for me!

Thank you!

PR #422, opened 2019-07-25
fix: dynamic routing

adascal:
Fixes: https://github.com/zeit/next.js/issues/8092

### DIFF
--- a/src/hocs/app-with-translation.js
+++ b/src/hocs/app-with-translation.js
@@ -36,7 +36,7 @@ export default function (WrappedComponent) {
               router.replace(href, as)
             }
           } else {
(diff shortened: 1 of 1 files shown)


isaachinman:
1. This will only work client side
2. This will only work for a subset of `localeSubpath` options
3. This will result in multiple router events being triggered in a row

adascal:
@isaachinman did you checked original issue? https://github.com/zeit/next.js/issues/8092

I've create and example with dynamic routing issue, and it happen only on client side

In case if I set `localeSubpath` my dynamic routes are transformed from http://localhost:3000/post/5 to http://localhost:3000/post/[id]?id=%5Bid%5D

PR #434, opened 2019-08-06
Fix: appWithTranslation / withTranslation typings

ivansky:
I've got several errors

1. error TS2339: Property `appWithTranslation` does not exist on type `NextI18Next`
2. Cannot invoke an expression whose type lacks a call signature. Type `WithTranslation` has no compatible call signatures. e.g. `withTranslation(i18nextNamespaces),`

`appWithTranslation` should be exported from type `NextI18Next` as it's exported in implementation.
`WithTranslation` type might be as function and as object, it depends on what we are looking for.

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -4,8 +4,7 @@ import * as React from 'react'
 import {
   useTranslation,
   TransProps,
-  withTranslation,
-  WithTranslation
+  withTranslation
 } from 'react-i18next'
 import { LinkProps } from 'next/link'
 import { SingletonRouter } from 'next/router'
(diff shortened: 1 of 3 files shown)


types.d.ts
```diff
@@ -39,8 +38,7 @@ export type UseTranslation = typeof useTranslation
 export type AppWithTranslation = <P extends object>(Component: React.ComponentType<P> | React.ElementType<P>) => any
 export type TFunction = i18next.TFunction
 export type I18n = i18next.i18n
-export type WithTranslation = WithTranslation
-export type withTranslation = typeof withTranslation
```
ivansky: It causes confusion because of shadow names. So maybe we need here renaming in importing and assign it to local exported type and renaming of actual hoc function, e.g. `WithTranslationHoc`

ivansky: @isaachinman What do you think about it?

isaachinman:
@kachkaev Have time for a review?

isaachinman:
Looks good, thanks for getting this bug fix in quickly.

PR #438, opened 2019-08-08
fix: update of i18next and react-i18next version to fix typescript 

dayze:
Fix the following error when using WithTranslation interface 
```Property 't' does not exist on type 'Readonly<Props> & Readonly<{ children?: ReactNode; }>'.```

Typescript user need to be at least at version ``` 17.0.9``` from the README of the react-i18next library
https://github.com/i18next/react-i18next/blob/master/README.md#requirements

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,14 +99,14 @@
     "core-js": "^2",
     "detect-node": "^2.0.4",
     "hoist-non-react-statics": "^3.2.0",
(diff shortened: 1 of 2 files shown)


isaachinman:
Updates to `package.json` must be accompanied by updates to lockfile(s).

isaachinman:
Needs a semantic title. Otherwise ready for merge.

PR #451, opened 2019-08-14
fix: Memory leak regression

### LINKED ISSUE #450: Possible memory leak
## Describe the bug
After starting using next-i18next in our application, we detected a memory leak. After investigating, we saw that it was probably caused by something related to the library. 

Using Chrome Devtools and --inspect flag of node in the example application provided by the repo, we found that a new i18n instance is allocated in memory after each request done to any page (e.g. /) and not freed when the request ends.

In the attached screenshot can be checked that there are 422 i18n instances, one for each request done in that particular test. This instances have a retained size of 10.5MB.

v0.54.0

## Steps to reproduce
We have used provided example application as-is, with its dependencies. To simulate the requests we have used artillery:

`
artillery quick -r 2 -d 10 http://localhost:3000
`

## Expected behaviour
You should not see the memory leak.

## Screenshots
Screenshot 2019-08-14 at 16 38 44

- Device: MBP 2018 13
- Browser: 76.0.3809.100
- Node: v10.16.0

isaachinman:
Fixes #450.

### DIFF
--- a/src/hocs/app-with-translation.tsx
+++ b/src/hocs/app-with-translation.tsx
@@ -34,31 +34,34 @@ export default function (WrappedComponent) {
 
     constructor(props) {
       super(props)
+      if ((process as any).browser) {
 
-      const changeLanguageCallback = (prevLng, newLng) => {
-        const { router } = props
-        const { pathname, asPath, query } = router
-        const routeInfo = { pathname, query }
+        const changeLanguageCallback = (prevLng, newLng) => {
+          const { router } = props
+          const { pathname, asPath, query } = router
+          const routeInfo = { pathname, query }
 
-        if ((process as any).browser && i18n.initializedLanguageOnce && typeof newLng === 'string' && prevLng !== newLng) {
-          const { as, href } = lngPathCorrector(config, { as: asPath, href: routeInfo }, newLng)
-          router.replace(href, as)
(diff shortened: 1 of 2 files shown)


src/hocs/app-with-translation.tsx
```diff
@@ -35,37 +35,40 @@ export default function (WrappedComponent) {
...
-        const { router } = props
-        const { pathname, asPath, query } = router
-        const routeInfo = { pathname, query }
-
-        if ((process as any).browser && i18n.initializedLanguageOnce && typeof newLng === 'string' && prevLng !== newLng) {
-          if (config.localeSubpaths !== localeSubpathOptions.NONE) {
-            const { as, href } = lngPathCorrector(config, { as: asPath, href: routeInfo }, newLng)
-            if (as !== asPath) {
-              router.replace(href, as)
+      if ((process as any).browser) {
```
kachkaev: Next.js itself no longer relies on `process.browser` and uses `typeof window` instead (https://github.com/zeit/next.js/pull/7651). Could `next-i18next` do the same to avoid compatibility issues in future major versions of Next?

isaachinman: Actually, we've got an `isServer` util [here](https://github.com/isaachinman/next-i18next/blob/master/src/config/default-config.ts#L3) which we should really be reusing.

And yes, we can check against `window`.

isaachinman: Will be handled via #454.

PR #458, opened 2019-08-16
chore: Add yarn.lock to npmignore

isaachinman:


kachkaev:
I’m not sure this is recommended, even for libraries.

isaachinman:
https://github.com/yarnpkg/yarn/issues/5406

kachkaev:
Ah right, I thought you were adding `yarn.lock` to `.gitignore` and not `.npmignore`, sorry!

As of `.npmignore`, it's often easier to first exclude everything and then include the stuff you want to see on npm. Something like:

```
*
!build/**
!README.md
```

Similarly, you can configure the whitelist in the `package.json` file: https://twitter.com/dan_abramov/status/679694009254686720

`package.json` itself never gets ignored, AFAIK.

isaachinman:
Thanks. In this case, the whitelist would probably be about the same length as this blacklist. Good to keep in mind, though!

.npmignore
```diff
@@ -4,6 +4,7 @@ examples
 
 # Dependencies
 node_modules
+yarn.lock
```
SalimBensiali, line 7: [This](https://classic.yarnpkg.com/en/docs/yarn-lock/#toc-check-into-source-control) says there is no harm in publishing yarn.lock, but just to keep the package tar balls clean from noisy files I decided to ignore it.

PR #459, opened 2019-08-16
fix: export `WithTranslation` props type for typescript

### LINKED ISSUE #360: Typescript translated component
## Describe the bug
In #220 was a discussion about typescript and `appWithTranslation` method coming from the package. I have a similar image on method `withTranslation`, see screenshot below:

image

`"next-i18next": "^0.44.0"`

## Steps to reproduce
This is my page component which throws the typescript issue
```typescript
import TopMenuBar from '../components/navigation/topMenuBar2'
import React, { Component } from 'react'
import scss from '../styles/index.scss'
import { withTranslation } from '../lib/i18n'

class IndexPage extends Component {

  static async getInitialProps() {
    return {
      namespacesRequired: ['common'],
    }
  }

  render() {
    return (
      <React.Fragment>
        <TopMenuBar title={'Index Page'} />
        <div className={[scss.hero, scss.hero2].join(' ')}>
          <h1 className={scss.title}>Welcome to create-next-app-cli (Create Next.js App building tools)</h1>
          <p className={scss.description}>To get started, edit <code>pages/index.js</code> and save to reload.
          </p>
          <div className={scss.row}>
            <a href="https://nextjs.org/docs/" className={scss.card}>
              <h3>Getting Stasdsdrted &rarr;</h3>
              <p>Learn more about Next.js on official website</p>
            </a>
            <a href="https://github.com/themodernjavascript/create-next-app-cli" className={scss.card}>
              <h3>Create Next App CLI &rarr;</h3>
              <p>Was this tools helpful?</p>
            </a>
          </div>
        </div>
      </React.Fragment>
    )
  }

}

export default withTranslation('common')(IndexPage)
```

## Expected behaviour
Class component should be vaild for the export with `withTranslation`

Anyone Ideas?

Averethel:
Fixes #360 

`WithTranslationHocType` is a new type that's used internally to type `withTranslation` HoC.It has been previously called `WithTranslation`.

That resulted in a name clash since `react-i18next` [exports its own `WithTranslation` interface](https://github.com/i18next/react-i18next/blob/master/src/index.d.ts#L69). This interface is the interface of the pros injected by `withTranslation` HoC (yes that naming is quite counter-intuitive). 

The issue was missed in the tests because the dummy component was not typed.

### DIFF
--- a/__tests__/types.test.tsx
+++ b/__tests__/types.test.tsx
@@ -10,7 +10,7 @@ import NextI18Next, {
   WithTranslation,
 } from '../types'
 
-const DummyComponent = () => <div />
+const DummyComponent: React.FC<{} & WithTranslation> = () => <div />
 
 const emptyConfig = {
   defaultLanguage: null,
@@ -21,4 +21,3 @@ const emptyConfig = {
 const Instance = new NextI18Next(emptyConfig)
 Instance.appWithTranslation(DummyComponent)
 Instance.withTranslation('common')(DummyComponent)
-
--- a/src/index.ts
+++ b/src/index.ts
@@ -9,7 +9,7 @@ import { consoleMessage } from './utils'
 import { Link } from './components';
 import { wrapRouter } from './router';
 
-import { AppWithTranslation, Config, InitConfig, Trans as TransType, Link as LinkType, I18n, UseTranslation, WithTranslation, Router } from '../types'
+import { AppWithTranslation, Config, InitConfig, Trans as TransType, Link as LinkType, I18n, UseTranslation, WithTranslationHocType, Router } from '../types'
 
 export { withTranslation } from 'react-i18next'
 
@@ -20,7 +20,7 @@ export default class NextI18Next {
   readonly i18n: I18n;
   readonly config: Config;
   readonly useTranslation: UseTranslation;
-  readonly withTranslation: WithTranslation;
+  readonly withTranslation: WithTranslationHocType;
   readonly appWithTranslation: AppWithTranslation;
 
   readonly consoleMessage: () => void
--- a/types.d.ts
+++ b/types.d.ts
@@ -4,7 +4,8 @@ import * as React from 'react'
 import {
   useTranslation,
   TransProps,
-  withTranslation
+  withTranslation,
+  WithTranslation
 } from 'react-i18next'
 import { LinkProps } from 'next/link'
 import { SingletonRouter } from 'next/router'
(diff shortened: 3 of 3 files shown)


isaachinman:
Can you please explain your changes here, and the need for two `withTranslation` types?

Averethel:
@isaachinman I updated the description

isaachinman:
@Averethel If you check the history of this file, you will see we've flip-flopped on this `withTranslation` value several times now. Are we absolutely sure this accomplishes everything surrounding `withTranslation`?

Averethel:
@isaachinman Yeah I just browsed and realized what was going on. It depends on what the vision is.
I see `next-i18next` as the only `i18n` related dependency I need in my next project. For that to work (if I'm using typescript), I need to be able to get the `i18n` related types that I need in my project so they need to be re-exported.
Of course, there are alternatives:
- importing them directly from the source, but then I'm in `i18n` dependency hell where I need to have `react-i18n` dependency version matching the one used in `next-i18n`, plus I'm adding it to only get a type
- declaring the prop type myself, but then I need to keep it in sync with what `react-i18next` is declaring and I risk unexpected errors
Of all the 3 re-exporting the types seem like the cleanest and most user-friendly to me.

isaachinman:
> I see next-i18next as the only i18n related dependency I need in my next project

Yes, that's right. All typing should be available via `next-i18next`. We need to re-export.

__tests__/types.test.tsx
```diff
@@ -10,7 +10,7 @@ import NextI18Next, {
   WithTranslation,
 } from '../types'
 
-const DummyComponent = () => <div />
+const DummyComponent: React.FC<{} & WithTranslation> = () => <div />
```
isaachinman: Why is this necessary, and what value does it add?

Averethel: We ensure that the `DummyComponent` has a correct react component type and correct prop type.

The syntax is `React.FC<_type_of_the_props_>`, because `DummyComponent` does not take any props its props type is `{}`. However, we wrap it with `withTransaltion` and this results with certain props being injected by `HoC` (`i18n`, `t`, and `tReady`, which are in `WithTranslation` type). 

For functional components (or stateless components) one could simply avoid typing it (so omit the `FC<Props>` part, however, this is not possible for the class components that need to extend `Component<Props, State>` class.

So long story short, by adding this we ensure that `WithTranslation` is a correct type re-exported from `react-i18next`.

Averethel:
@isaachinman any news?

PR #466, opened 2019-08-21
Add support for arrays in localeSubpaths

rubystephenpoole:


isaachinman:
Hi @rubystephenpoole. Can you please first open an issue when you'd like to see a new feature implemented, so that we can discuss the implications before starting the work?

I think that this approach may cause more problems than it solves, and it could be easier/safer for users to use custom NextJs routes or proxies.

rubystephenpoole:
No problem, I would have done the work regardless as it's a required use-case for our application. Custom routes don't produce identical functionality to the existing i18next middleware, and proxying isn't ideal in our environment.

 I just figured might as well attempt the pr since we have the code in place. If the use case isn't something you'd like to incorporate, no big deal.

isaachinman:
Ah, I see. I think it's a relatively uncommon use case, but it's good to have this work as a point of reference in case the request does come up again. Thanks!

PR #475, opened 2019-08-27
refactor: produce optimized build for nodejs

### LINKED ISSUE #474: No need for polyfilling or transpiling in expressjs middlewares
I just looked at the express middleware and the generated output is huge. I see no need for this because Node 8 and higher are very modern.
This definitly reduce the performance and readability in debugging scenarios. I would propose to use a different compiler settings for node.js only parts.

```js
"use strict";

var _interopRequireDefault = require("@babel/runtime/helpers/interopRequireDefault");

require("core-js/modules/es6.object.define-property");

Object.defineProperty(exports, "__esModule", {
  value: true
});
exports["default"] = _default;

var _objectSpread2 = _interopRequireDefault(require("@babel/runtime/helpers/objectSpread"));

require("core-js/modules/es6.array.for-each");

require("core-js/modules/es6.array.some");

require("core-js/modules/web.dom.iterable");

require("core-js/modules/es6.array.iterator");

require("core-js/modules/es6.object.to-string");

require("core-js/modules/es7.object.values");

require("core-js/modules/es6.string.starts-with");

require("core-js/modules/es6.array.is-array");

var _i18nextExpressMiddleware = _interopRequireDefault(require("i18next-express-middleware"));

var _pathMatch = _interopRequireDefault(require("path-match"));

var _utils = require("../utils");

var route = (0, _pathMatch["default"])();

function _default(nexti18next) {
  var config = nexti18next.config,
      i18n = nexti18next.i18n;
  var allLanguages = config.allLanguages,
      ignoreRoutes = config.ignoreRoutes,
      localeSubpaths = config.localeSubpaths;

  var isI18nRoute = function isI18nRoute(req) {
    return ignoreRoutes.every(function (x) {
      return !req.url.startsWith(x);
    });
  };

  var localeSubpathRoute = route("/:subpath(".concat(Object.values(localeSubpaths).join('|'), ")(.*)"));
  var middleware = [];
  /*
    If not using server side language detection,
    we need to manually set the language for
    each request
  */

  if (!config.serverLanguageDetection) {
    middleware.push(function (req, _res, next) {
      if (isI18nRoute(req)) {
        req.lng = config.defaultLanguage;
      }

      next();
    });
  }
  /*
    This does the bulk of the i18next work
  */


  middleware.push(_i18nextExpressMiddleware["default"].handle(i18n, {
    ignoreRoutes: ignoreRoutes
  }));
  /*
    This does the locale subpath work
  */

  middleware.push(function (req, res, next) {
    if (isI18nRoute(req) && req.i18n && (0, _utils.subpathIsRequired)(config, req.i18n.language)) {
      var lng = (0, _utils.lngFromReq)(req);
      var currentLngSubpath = (0, _utils.subpathFromLng)(config, lng);
      /*
   .......................
...........................................
```

StarpTech:
fix #474


StarpTech:
<details><summary>Before</summary>
<p>

```js
"use strict";

var _interopRequireDefault = require("@babel/runtime/helpers/interopRequireDefault");

require("core-js/modules/es6.object.define-property");

Object.defineProperty(exports, "__esModule", {
  value: true
});
exports["default"] = _default;

require("core-js/modules/es6.object.assign");

require("core-js/modules/es6.array.for-each");

require("core-js/modules/es6.array.some");

require("core-js/modules/web.dom.iterable");

require("core-js/modules/es6.array.iterator");

require("core-js/modules/es6.object.to-string");

require("core-js/modules/es7.object.values");

require("core-js/modules/es6.string.starts-with");

require("core-js/modules/es6.array.is-array");

var _i18nextExpressMiddleware = _interopRequireDefault(require("i18next-express-middleware"));

var _pathMatch = _interopRequireDefault(require("path-match"));

var _utils = require("../utils");

var route = (0, _pathMatch["default"])();

function _default(nexti18next) {
  var config = nexti18next.config,
      i18n = nexti18next.i18n;
  var allLanguages = config.allLanguages,
      ignoreRoutes = config.ignoreRoutes,
      localeSubpaths = config.localeSubpaths;

  var isI18nRoute = function isI18nRoute(req) {
    return ignoreRoutes.every(function (x) {
      return !req.url.startsWith(x);
    });
  };

  var localeSubpathRoute = route("/:subpath(".concat(Object.values(localeSubpaths).join('|'), ")(.*)"));
  var middleware = [];
  /*
    If not using server side language detection,
    we need to manually set the language for
    each request
  */

  if (!config.serverLanguageDetection) {
    middleware.push(function (req, _res, next) {
      if (isI18nRoute(req)) {
        req.lng = config.defaultLanguage;
      }

      next();
    });
  }
  /*
    This does the bulk of the i18next work
  */


  middleware.push(_i18nextExpressMiddleware["default"].handle(i18n, {
    ignoreRoutes: ignoreRoutes
  }));
  /*
    This does the locale subpath work
  */

  middleware.push(function (req, res, next) {
    if (isI18nRoute(req) && req.i18n && (0, _utils.subpathIsRequired)(config, req.i18n.language)) {
      var lng = (0, _utils.lngFromReq)(req);
      var currentLngSubpath = (0, _utils.subpathFromLng)(config, lng);
      /*
        This case will be entered if a subpath
        is required but not present
      */

      if (!(0, _utils.subpathIsPresent)(req.url, currentLngSubpath)) {
        var otherSubpathPresent = allLanguages.some(function (l) {
          return (0, _utils.subpathIsPresent)(req.url, (0, _utils.subpathFromLng)(config, l));
        });

        if (otherSubpathPresent) {
          /*
            If a user has hit a subpath which does not
            match their language, give preference to
            the path, and change user language.
          */
          allLanguages.forEach(function (l) {
            if ((0, _utils.subpathIsPresent)(req.url, (0, _utils.subpathFromLng)(config, l))) {
              req.i18n.changeLanguage(l);
            }
          });
        } else {
          /*
            If a language subpath is required and
            not present, prepend correct subpath
          */
          (0, _utils.redirectWithoutCache)(res, (0, _utils.addSubpath)(req.url, currentLngSubpath));
        }
      }
      /*
        If a locale subpath is present in the URL,
        modify req.url in place so that NextJs will
        render the correct route
      */


      var params = localeSubpathRoute(req.url);

      if (params !== false) {
        var subpath = params.subpath;
        req.query = Object.assign({
          subpath: subpath,
          lng: lng
        }, req.query);
        req.url = (0, _utils.removeSubpath)(req.url, subpath);
      }
    }

    next();
  });
  return middleware;
}
```

</p>
</details>

<details><summary>After</summary>
<p>

```js
"use strict";

Object.defineProperty(exports, "__esModule", {
  value: true
});
exports.default = _default;

var _i18nextExpressMiddleware = _interopRequireDefault(require("i18next-express-middleware"));

var _pathMatch = _interopRequireDefault(require("path-match"));

var _utils = require("../utils");

function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }

const route = (0, _pathMatch.default)();

function _default(nexti18next) {
  const {
    config,
    i18n
  } = nexti18next;
  const {
    allLanguages,
    ignoreRoutes,
    localeSubpaths
  } = config;

  const isI18nRoute = req => ignoreRoutes.every(x => !req.url.startsWith(x));

  const localeSubpathRoute = route(`/:subpath(${Object.values(localeSubpaths).join('|')})(.*)`);
  const middleware = [];
  /*
    If not using server side language detection,
    we need to manually set the language for
    each request
  */

  if (!config.serverLanguageDetection) {
    middleware.push((req, _res, next) => {
      if (isI18nRoute(req)) {
        req.lng = config.defaultLanguage;
      }

      next();
    });
  }
  /*
    This does the bulk of the i18next work
  */


  middleware.push(_i18nextExpressMiddleware.default.handle(i18n, {
    ignoreRoutes
  }));
  /*
    This does the locale subpath work
  */

  middleware.push((req, res, next) => {
    if (isI18nRoute(req) && req.i18n && (0, _utils.subpathIsRequired)(config, req.i18n.language)) {
      const lng = (0, _utils.lngFromReq)(req);
      const currentLngSubpath = (0, _utils.subpathFromLng)(config, lng);
      /*
        This case will be entered if a subpath
        is required but not present
      */

      if (!(0, _utils.subpathIsPresent)(req.url, currentLngSubpath)) {
        const otherSubpathPresent = allLanguages.some(l => (0, _utils.subpathIsPresent)(req.url, (0, _utils.subpathFromLng)(config, l)));

        if (otherSubpathPresent) {
          /*
            If a user has hit a subpath which does not
            match their language, give preference to
            the path, and change user language.
          */
          allLanguages.forEach(l => {
            if ((0, _utils.subpathIsPresent)(req.url, (0, _utils.subpathFromLng)(config, l))) {
              req.i18n.changeLanguage(l);
            }
          });
        } else {
          /*
            If a language subpath is required and
            not present, prepend correct subpath
          */
          (0, _utils.redirectWithoutCache)(res, (0, _utils.addSubpath)(req.url, currentLngSubpath));
        }
      }
      /*
        If a locale subpath is present in the URL,
        modify req.url in place so that NextJs will
        render the correct route
      */


      const params = localeSubpathRoute(req.url);

      if (params !== false) {
        const {
          subpath
        } = params;
        req.query = Object.assign({
          subpath,
          lng
        }, req.query);
        req.url = (0, _utils.removeSubpath)(req.url, subpath);
      }
    }

    next();
  });
  return middleware;
}
```
</p>
</details>

isaachinman:
We'd need something a bit more substantial than passing an ignore flag, in case this changes in the future.

And it'd be great to add a test step that uses [es-check](https://www.google.com/search?q=escheck&oq=escheck&aqs=chrome..69i57j0l5.821j1j0&sourceid=chrome&ie=UTF-8).

StarpTech:
I got your point but I set the target to node 8 everything will be polyfilled and the CI should catch incompatibilities.

isaachinman:
We definitely won't be modifying the actual codebase to influence transpiled output.

StarpTech:
@isaachinman I don't understand you. What's is missing here? This PR will produce much more readable and performant code for node.js. The code is transpiled with preset-env set to node:8. Every incompatibility with new ES features is checked on the CI.

StarpTech:
> We definitely won't be modifying the actual codebase to influence transpiled output.
Only because of the one-time use of the spread operator it's not worth to add the polyfills.

StarpTech:
@isaachinman what has to be changed in order to merge it? Please have a look to my comments above.

isaachinman:
1. We definitely won't be modifying the actual codebase to influence transpiled output
2. We'd need something a bit more substantial than passing an ignore flag

StarpTech:
1. You think it's really worth to add more 30 lines of code just to use the spread operator instead `Object.assign` ?
2. Could you be more concrete? I can't produce next steps from that statement.

isaachinman:
1. Nope. I think it's worth it to prevent accidental regressions in the future. We can't expect all contributors to somehow know they are not allowed to use specific ES features.
2. I'd rather use `include` and `exclude` inside the actual Babel config, and not pass this as a CLI flag.

isaachinman:
Let me also reiterate: this may or may not even improve performance. LOC is not so much of a concern in NodeJs envs, and you might be surprised at what performs well (ES3 vs ES6, etc).

StarpTech:
1. That's why we have a CI which validates all supported Node versions and even when it happens babel will polyfill it so I don't get your point.
2. That's no problem.

> Let me also reiterate: this may or may not even improve performance. LOC is not so much of a concern in NodeJs envs, and you might be surprised at what performs well (ES3 vs ES6, etc).

It's more a question of the need. Why should we polyfill something when node already supports it? Why do you handle node as a browser environment?

StarpTech:
> Let me also reiterate: this may or may not even improve performance. LOC is not so much of a concern in NodeJs envs, and you might be surprised at what performs well (ES3 vs ES6, etc).

I really care on which code my project depends. Such polyfills will polute the global namespace as well and can have negative impact because the native implementation will be overriden isn't ?

isaachinman:
> That's why we have a CI which validates all supported Node versions

What CI are you referring to?

> Why do you handle node as a browser environment?

As stated from the beginning, I'm supportive of this PR! Don't mistake that part. We just need to be absolutely sure that we support the minimum NodeJs version that NextJs officially does.

StarpTech:
> What CI are you referring to?

Oh, the code isn't tested against the supported version. Would you accept a PR to use [circelci workflows](https://circleci.com/blog/introducing-workflows-on-circleci-2-0/) to test against e.g node 8 and node 10 ?

> As stated from the beginning, I'm supportive of this PR! Don't mistake that part. We just need to be absolutely sure that we support the minimum NodeJs version that NextJs officially does.

Due to the fact that this changes very rarely I would go with a fixed build matrix in the circleci config.

isaachinman:
> Would you accept a PR to use circelci workflows to test against e.g node 8 and node 10 ?

Yes, but please open an issue first so that we can discuss implementation.

StarpTech:
@isaachinman flag was added in babel env

isaachinman:
@StarpTech Why is there a new `server-i18n-setup.ts` file?

StarpTech:
@isaachinman because this code is only used server side and we can optimize the code for node as well.

isaachinman:
@StarpTech Why does it need to be optimised? It's eight LOC.

isaachinman:
These additions are not going to make a difference in Node perf, and make the project harder to maintain.

StarpTech:
My intention is to handle all server code in that way. This has no drawback in maintance but on the contrary you get more awareness that this code is for server side only.

isaachinman:
In the world of universal/isomorphic/SSR applications, we often have shared code, which is executed by both browsers and NodeJs. This is why the original setup of this package was to simply transpile to a browser target - because could be sure that Node could also execute it.

The solution you're proposing is to add a new NodeJs target, but even if you shift all server-side-only code to a NodeJs target, we will still have shared code that needs to be able to be executed by both a browser and Node.

This adds complexity to the build process and adds cognitive load to active development.

Mind sharing some perf reports that show a clear improvement? I'm concerned that this is premature optimisation.

StarpTech:
> In the world of universal/isomorphic/SSR applications, we often have shared code, which is executed by both browsers and NodeJs. This is why the original setup of this package was to simply transpile to a browser target - because could be sure that Node could also execute it.

100% agree and modern build tools allow us to negate the drawbacks of that shared approach.

> The solution you're proposing is to add a new NodeJs target, but even if you shift all server-side-only code to a NodeJs target, we will still have shared code that needs to be able to be executed by both a browser and Node.

That's correct. I will never touch shared code. Moving server-side-only files to a specific folder and handle it differently in the build step is no increase in complexity we only need to document it.

> This adds complexity to the build process and adds cognitive load to active development.

I don't know. In practice I will never see how my code is generated. I can develop in typescript and write my tests. I don't understand that argument. btw: With the introduction of typescript you complicated the build/project a lot.

> Mind sharing some perf reports that show a clear improvement? I'm concerned that this is premature optimisation.

It's a preoptimization but with a win win situation. We can easily set the correct target and the code is much more readable when debugging with node_modules.

raphaelbadia:
It would be really helpful not to have this ^2 corejs dependency in the project, I wouldn't get these errors when running next-i81next on storybook

```
 ERROR in ./node_modules/next-i18next/dist/commonjs/hocs/app-with-translation.js
 Module not found: Error: Cannot find module 'core-js/modules/es6.regexp.replace'
@ ./node_modules/next-i18next/dist/commonjs/hocs/app-with-translation.js 38
```

StarpTech:
Closing that issue since the discussion is simply ignored by the maintainer.

Final thoughts: Writing isomorphic javascript is not that easy as it sounds because you still need to differentiate between Node.js and the browser environment.

From the server-side perspective, I don't want to run polyfills when they aren't needed because they pollute the global space, error-stack and can lead to unexpected performance degradations.

isaachinman:
Hey @StarpTech, not ignoring it! Let's reopen it and get it merged this week. I think we are misaligned on the impact on NodeJs runtimes but this is nonetheless an objective improvement.

isaachinman:
So let's put aside perf issues of transpiled JS vs non-transpiled JS.

The issue of pulling polyfills into a NodeJs env is valid, but can you explain how this PR achieves that? Won't our Node process still load `create-i18next-client.ts`, which is transpiled to the `cjs` target?

StarpTech:
@isaachinman ping

StarpTech:
@isaachinman ping the second. Are you still interested?

lifedraft:
🙏

StarpTech:
@isaachinman??? I'd really appreciate to get answer from you. You are the only maintainer.

isaachinman:
> Won't our Node process still load create-i18next-client.ts, which is transpiled to the cjs target?

> Sure, but we will load the optimized Node.js build.

I don't really get it. The cjs target pulls in core-js v2. What is the point of this PR if we inevitably have isomorphic code that needs to run on both the client and the server? That would simply complicate the build process for no gain.

StarpTech:
The point of this PR is to create an optimized build for code which is running only on server side. How do we archive that? By using the nodejs target in Babel instead polyfilling for browsers.

Why is it useful?

- less code
- no patching for features which already exists
- easier to debug locally

Is it clear now?

isaachinman:
It's 10 LOC. I don't get it. Sorry - this probably will not be merged as it complicates the build process without removing the cjs target from NodeJs execution.

StarpTech:
In my initial post you see the difference. It's not just about LOC. 

By the way its really hard to discuss with you a PR. This work was done 30 days ago. You switched your opinions multiple times. This also effects serveral issues opened by me.

Please take your time and start with an open mind. You also like to mixup things. In the last comment you disagreed and complained about my changes instead to suggest code changes.

I will close this issue because you are only maintainer of this project and it seems nobody else is interested in this change.

isaachinman:
Hi @StarpTech - sorry for any frustration caused. Unfortunately by nature, this kind of OSS work is done for free by people whose time is mostly taken up with commercial work. It will, unfortunately, never have anyone's full attention, as it is not backed by any org.

I appreciate your patience and willingness to contribute!

PR #479, opened 2019-08-28
fix: dont call next on redirect

### LINKED ISSUE #478: Middleware isn't aborted on redirect
## Describe the bug
When the request is redirected [here](https://github.com/isaachinman/next-i18next/blob/master/src/middlewares/next-i18next-middleware.ts#L81) the middleware isn't aborted. The `next` callback is always called [here](https://github.com/isaachinman/next-i18next/blob/master/src/middlewares/next-i18next-middleware.ts#L100). This results in double invocation of the express handlers.

## Expected behaviour
Don't call next middlware and abort the response with the redirect call.

## Reproducible
Install the simple example in this repository.

### Use this server code
```
(async () => {
  await app.prepare();
  const server = express();

  server.use(nextI18NextMiddleware(nextI18next));

  server.get('/', (req, res) => {
    console.log('##############');
    handle(req, res);
  });

  server.get('*', (req, res) => {
    handle(req, res);
  });

  await server.listen(port);
  console.log(`> Ready on http://localhost:${port}`); // eslint-disable-line no-console
})();
```
1. Start server with `LOCALE_SUBPATHS=all yarn dev`
2. Go to `http://localhost:3000`
3. You will see the log output two times.

StarpTech:
This will fix #478

### DIFF
--- a/src/middlewares/next-i18next-middleware.ts
+++ b/src/middlewares/next-i18next-middleware.ts
@@ -79,7 +79,7 @@ export default function (nexti18next) {
             not present, prepend correct subpath
           */
           redirectWithoutCache(res, addSubpath(req.url, currentLngSubpath))
-
+          return
         }
         
       }
--- a/src/middlewares/next-i18next-middleware.ts
+++ b/src/middlewares/next-i18next-middleware.ts
@@ -78,8 +78,7 @@ export default function (nexti18next) {
             If a language subpath is required and
             not present, prepend correct subpath
           */
-          redirectWithoutCache(res, addSubpath(req.url, currentLngSubpath))
-          return
+          return redirectWithoutCache(res, addSubpath(req.url, currentLngSubpath))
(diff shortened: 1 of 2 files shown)


StarpTech:
For future improvements I suggest to run an integration test instead mocks and spy. This is really a mess :smile:

src/middlewares/next-i18next-middleware.ts
```diff
@@ -79,7 +79,7 @@ export default function (nexti18next) {
             not present, prepend correct subpath
           */
           redirectWithoutCache(res, addSubpath(req.url, currentLngSubpath))
```
isaachinman: Lets return this function call instead of `undefined`.

isaachinman:
FYI we don't need integration tests to cover this, just a more sensible way of mocking utils. Thank you very much for your help!

PR #484, opened 2019-08-29
feat: target multiple nodejs versions in circleci

### LINKED ISSUE #482: Test against all supported node versions
Circlecli has the capability to test against a matrix with the help of [workflows](https://circleci.com/blog/introducing-workflows-on-circleci-2-0/).
I propose to test the same matrix as next.js does.

[Next.js .circleci/config.yml](https://github.com/zeit/next.js/blob/canary/.circleci/config.yml)

In future we could also improve to test against browser environments instead node only.

StarpTech:
Thi will fix #482

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,46 +1,38 @@
 version: 2.1
 commands:
-  build-test:
+  build-and-test:
     steps:
       - checkout
       - restore_cache:
           name: Restore cache (main)
           keys:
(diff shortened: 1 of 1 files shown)


StarpTech:
@isaachinman please check

isaachinman:
Is there any way to prevent duplication of our steps?

StarpTech:
I research it yesterday a bit but I didn't find a solution until right now :smile:

PR #506, opened 2019-09-11
fix(bug): update lng query param on subpath change

### LINKED ISSUE #504: lng query param is wrong when on secondary language
## Describe the bug
When loading a secondary lang page, the attached `lng` query string param is the **default** one.

``` 
// My config
{
  browserLanguageDetection: false,
  serverLanguageDetection: false,
  defaultLanguage: 'en',
  otherLanguages: ['de'],
  localeSubpaths: {
    de: 'de',
  }
}
```
v1.2.0

## Steps to reproduce
Navigate to `/de/hello-world` page.
Check the `lng` query parameter value, it will `en` instead of **de**.

## Expected behaviour
The `lng` query parameter should reflect a proper lng value.

 - Device: [e.g. MBP 2018 13"]

## Additional context
Go to see the issue
https://pknpq.sse.codesandbox.io/de/about

https://codesandbox.io/s/next-i18next-issue-wrong-query-pknpq

felixmosh:
This PR fixes #504 bug

### DIFF
--- a/src/middlewares/next-i18next-middleware.ts
+++ b/src/middlewares/next-i18next-middleware.ts
@@ -47,7 +47,7 @@ export default function (nexti18next) {
   */
   middleware.push((req: Request, res: Response, next: NextFunction) => {
     if (isI18nRoute(req) && req.i18n) {
-      const currentLng = lngFromReq(req)
+      let currentLng = lngFromReq(req)
       const currentLngSubpath = subpathFromLng(config, currentLng)
       const currentLngRequiresSubpath = subpathIsRequired(config, currentLng)
       const currentLngSubpathIsPresent = subpathIsPresent(req.url, currentLngSubpath)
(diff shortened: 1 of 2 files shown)


src/middlewares/next-i18next-middleware.ts
```diff
@@ -56,7 +56,7 @@ export default function (nexti18next) {
         subpathIsPresent(req.url, subpathFromLng(config, l)))
 
       if (lngFromCurrentSubpath !== undefined && lngFromCurrentSubpath !== currentLng) {
-
+        currentLng = lngFromCurrentSubpath
```
isaachinman: Can we achieve the same thing by just referencing `req.i18n.language` below instead of putting `currentLng` into the query?

felixmosh: I changed the value because the assignment occurs here  https://github.com/isaachinman/next-i18next/blob/e6d70129a43012e0e503f51af28ed47f7bf5adb4/src/middlewares/next-i18next-middleware.ts#L85

isaachinman: Yes I know, I'm asking if it's possible to get the language from `req.i18n` instead.

isaachinman: Not sure we are understanding each other. I can take a look later, this does appear to fix a bug in the short term.

Can you please just move this new line below the `req.i18n.changeLanguage` line. Thanks!

felixmosh: Done

felixmosh:
please tag me whenever you change this code, I would like to learn :]

felixmosh:
When it will be published?

isaachinman:
v1.2.1.

PR #519, opened 2019-10-03
feat: Make the as prop accept all kind of UrlLike or string

### LINKED ISSUE #517: URL object as a prop
The problem is related to the error `'as' type must be 'string', but it is object`

Next.js's Link can accept more than a string for the `as` property, it can be a `UrlLike`

For example: `<Link href={...} as={{ pathname: '/test', query: { page: 1 } }}` with output `<a href="/test?page=1">`.

It would be neat to have it in `next-i18next` as well

The usual `<Link href={...} as={'/test?page=1'}`, but it's a bit more complex to manipulate with dynamic data.

## Additional context
n/a

martpie:
Should fix #517

```tsx
<Link
  href='/somepage'
  as={{ pathname: '/test', hash: 'something', query: { page: 2 } }}
>
  Test
</Link>
```

will output now something like: 

```
http://localhost:3000/en/test?page=2#something
```

### DIFF
--- a/__tests__/utils/lng-path-corrector.test.ts
+++ b/__tests__/utils/lng-path-corrector.test.ts
@@ -35,7 +35,7 @@ describe('lngPathCorrector utility function', () => {
   it('throws if currentRoute.as is not undefined or a string', () => {
     currentRoute.as = 10
     expect(() => lngPathCorrector(config, currentRoute, 'en')).toThrowError(
-      '\'as\' type must be \'string\', but it is number',
+      '\'as\' type must be \'string\' or an \'object\', but it is number',
     )
   })
 
--- a/src/utils/lng-path-corrector.ts
+++ b/src/utils/lng-path-corrector.ts
@@ -1,4 +1,4 @@
-import { format as formatUrl, parse as parseUrl } from 'url'
+import { format as formatUrl, parse as parseUrl, UrlObject } from 'url'
 import { LinkProps } from 'next/link'
 
 import { Config } from '../../types'
@@ -8,6 +8,10 @@ import subpathFromLng from './subpath-from-lng'
 
 type As = LinkProps['as']
 type Href = LinkProps['href']
+interface Route {
+  href: Href;
+  as: As;
+}
 
 const parseAs = (originalAs, href): As => {
   const asType = typeof originalAs
@@ -24,7 +28,7 @@ const parseAs = (originalAs, href): As => {
   return as
 }
 
-const parseHref = (originalHref): Href => {
+const parseHref = (originalHref): UrlObject => {
   const hrefType = typeof originalHref
   let href
 
@@ -40,7 +44,7 @@ const parseHref = (originalHref): Href => {
   return href
 }
 
-export default (config: Config, currentRoute, currentLanguage): LinkProps => {
+export default (config: Config, currentRoute: Route, currentLanguage: string): LinkProps => {
   const { allLanguages, localeSubpaths } = config
   const { as: originalAs, href: originalHref } = currentRoute
 
@@ -63,7 +67,7 @@ export default (config: Config, currentRoute, currentLanguage): LinkProps => {
   */
   Object.values(localeSubpaths).forEach((subpath) => {
     if (typeof as === 'object') {
-      if (subpathIsPresent(as, subpath)) {
+      if (subpathIsPresent(as.pathname, subpath)) {
         as.pathname = removeSubpath(as.pathname, subpath)
       }
     } else if (typeof as === 'string') {
@@ -87,8 +91,12 @@ export default (config: Config, currentRoute, currentLanguage): LinkProps => {
       as = `/${subpath}${currentAs}`.replace(/\/$/, '')
     }
 
+    if (typeof href.query !== 'object') throw new Error(`'href.query' must be an object`)
+    if (!href.query) href.query = {}
+
+
     href.query.lng = currentLanguage
-    href.query.subpath = subpath
+    href.query.subpath = subpath || ''
   }
 
   return { as, href }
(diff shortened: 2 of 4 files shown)


martpie:
I have an additional question: would you accept a PR to enable `strict: true` in TypeScript? I will happily work on making it work, I already saw a couple of potential bugs covered by `strictNullChecks` etc

isaachinman:
> enable strict: true in TypeScript?

Yes.

martpie:
I'll do it in a separate pr then :)

martpie:
Do you need anything else from me @isaachinman ?

__tests__/utils/lng-path-corrector.test.ts
```diff
@@ -35,7 +35,7 @@ describe('lngPathCorrector utility function', () => {
   it('throws if currentRoute.as is not undefined or a string', () => {
```
isaachinman: Needs to be updated.

src/utils/lng-path-corrector.ts
```diff
@@ -1,26 +1,34 @@
...
+}
+
+const parseAs = (originalAs, href): As => {
   const asType = typeof originalAs
   let as
 
   if (asType === 'undefined') {
     as = formatUrl(href, { unicode: true })
-  } else if (asType === 'string') {
+  } else if (asType === 'string' || (asType === 'object' && originalAs !== null)) {
```
isaachinman: Is checking the type against `object` safe here? Many, many things are `objects` in JavaScript.

src/utils/lng-path-corrector.ts
```diff
@@ -1,26 +1,34 @@
...
   } else {
-    throw new Error(`'as' type must be 'string', but it is ${asType}`)
+    throw new Error(`'as' type must be 'string' or an 'object', but it is ${asType}`)
   }
 
   return as
 }
 
-const parseHref = (originalHref) => {
+const parseHref = (originalHref): UrlObject => {
```
isaachinman: Is this typing correct? Can't it also be a string?

src/utils/lng-path-corrector.ts
```diff
@@ -58,19 +66,37 @@ export default (config: Config, currentRoute, currentLanguage) => {
     Strip any/all subpaths from the `as` value
   */
   Object.values(localeSubpaths).forEach((subpath) => {
-    if (subpathIsPresent(as, subpath)) {
-      as = removeSubpath(as, subpath)
+    if (typeof as === 'object') {
```
isaachinman: Shouldn't we also check that `as.pathname` is a string?

src/utils/lng-path-corrector.ts
```diff
@@ -58,19 +66,37 @@ export default (config: Config, currentRoute, currentLanguage) => {
...
 
   if (subpathIsRequired(config, currentLanguage)) {
     const basePath = `${href.protocol}//${href.host}`
-    const currentAs = as.replace(basePath, '')
+    let currentAs = ''
+
     const subpath = subpathFromLng(config, currentLanguage)
 
-    as = `/${subpath}${currentAs}`.replace(/\/$/, '')
+    if (typeof as === 'object') {
```
isaachinman: Same feedback here for `as.pathname`.

src/utils/lng-path-corrector.ts
```diff
@@ -58,19 +66,37 @@ export default (config: Config, currentRoute, currentLanguage) => {
...
+      as = `/${subpath}${currentAs}`.replace(/\/$/, '')
+    }
+
+    if (typeof href.query !== 'object') throw new Error(`'href.query' must be an object`)
+    if (!href.query) href.query = {}
+
+
     href.query.lng = currentLanguage
-    href.query.subpath = subpath
+    href.query.subpath = subpath || ''
```
isaachinman: Can you explain why this is necessary?

src/utils/lng-path-corrector.ts
```diff
@@ -58,19 +66,37 @@ export default (config: Config, currentRoute, currentLanguage) => {
...
+    if (typeof as === 'object') {
+      currentAs = as.pathname.replace(basePath, '')
+      as.pathname = `/${subpath}${currentAs}`.replace(/\/$/, '')
+    } else if (typeof as === 'string') {
+      currentAs = as.replace(basePath, '')
+      as = `/${subpath}${currentAs}`.replace(/\/$/, '')
+    }
+
+    if (typeof href.query !== 'object') throw new Error(`'href.query' must be an object`)
+    if (!href.query) href.query = {}
```
isaachinman: Besides the formatting, which needs to be fixed, this looks pretty dangerous to me. A falsy check isn't good enough - let's be clear about our assertions here.

Also, you're throwing if `href.query` isn't an `object` the line before this. Can you explain the logic here?

martpie:
Sorry for the delay, I will try to address some of your comments and reply to some others asap

> These changes make me a little nervous as they increase complexity and seem to be relying on some unclear/unsafe type assertions.

Regarding the complexity, maybe I can try to simplify things a little bit and be a bit less imperative/sequential.

Regarding the unsafe type assertions, you're right and I will try to improve these. I based myself a lot on TypeScript type checking + Next.js's typings, but there may be runtime issues indeed.

isaachinman:
Hi @martpie - do you want to proceed with this one?

martpie:
By proceeding, do you mean taking it over? or discarding it? 😄

isaachinman:
@martpie Sure, I can take it over if you don't want to continue. Just let me know.

martpie:
Dropping it, it's a bit outdated and I will restart a branch from scratch, addressing some PR comments. I will also check how `next` implemented their link component as there may be some interesting things to get there.

PR #520, opened 2019-10-03
chore: Add directly react, react-dom, next-i18next and update next

wootsbot:
- Add directly react, react-dom, next-i18next and update next, i18next to have an updated example.

I think we can have a more updated example and I felt free to update the dependencies and eliminate obsolete things

### DIFF
--- a/examples/simple/package.json
+++ b/examples/simple/package.json
@@ -14,10 +14,10 @@
   "devDependencies": {},
   "dependencies": {
     "express": "^4.16.4",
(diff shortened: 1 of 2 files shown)


isaachinman:
Feel free to update deps, but the deps with `link:` must remain that way.

Also, I don't think we should have `i18next` installed as a dep in the example dir - should resolve that from `next-i18next`.

isaachinman:
Closing due to inactivity.

PR #522, opened 2019-10-07
feat: Add option to configure client locale path

itsfadnis:
Currently the client locale path (server route that delivers a translation file) is assumed to be the same as the localePath option.

This PR provides allows us specify a custom path for it via the clientLocalePath option, which is optional.

For example I have a project with the following folder structure:
```
root
  - src // Folder where the next app resides
    - i18n.js
    - server.js
    - static
      - locales
        - en
        - fr
  - package.json
  - next.config.json
```

And this is where I initialize my `NextI18Next` instance:
```
// src/i18n.js

  defaultLanguage: 'en',
  otherLanguages: ['fr'],
  localePath: 'src/static/locales',
  // Optional
  clientLocalePath: 'static/locales'
})
```

Without the `clientLocalePath` option, the backend route that the client side would hit would be `/src/static/locales/{{ns}}/{{lng}}.json`

With the `clientLocalePath` option, we can add a custom path which in this case would be `/static/locales/{{ns}}/{{lng}}.json`.

### DIFF
--- a/__tests__/config/create-config.test.ts
+++ b/__tests__/config/create-config.test.ts
@@ -189,8 +189,8 @@ describe('create configuration in non-production environment', () => {
 
       expect(config.ns).toEqual(['universal'])
 
-      expect(config.backend.loadPath).toEqual('/static/translations/{{ns}}/{{lng}}.json')
-      expect(config.backend.addPath).toEqual('/static/translations/{{ns}}/{{lng}}.missing.json')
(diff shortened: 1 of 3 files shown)


isaachinman:
This adds unnecessary scope to the API.

You can simply flip over `process.browser` or `typeof window`.

dobrud:
Thanks a lot @isaachinman. I got it working with

```
  defaultLanguage: 'en',
  otherLanguages: ['fr'],
  localePath: typeof window !== undefined ? 'src/static/locales' : 'static/locales',
})
```

Issue #523, opened 2019-10-08
Add support for "public" directory


urkopineda:
## Is your feature request related to a problem? Please describe.
From `nextjs` 9.0.6 and above the `static` directory is moved to `public/static`. Is not mandatory to move it, but the `static` directory will be deprecated in the future. [More information here.](https://github.com/zeit/next.js/blob/master/errors/static-dir-deprecated.md)

Moving the `static` folder to `public/static` throws the next error:
```
***/app/node_modules/next-i18next/dist/commonjs/config/create-config.js:54
        throw new Error("Default namespace not found at ".concat(defaultNSPath));
        ^

Error: Default namespace not found at ***/app/static/locales/es/common.json
    at _default (***/app/node_modules/next-i18next/dist/commonjs/config/create-config.js:54:15)
    at new NextI18Next (***/app/node_modules/next-i18next/dist/commonjs/index.js:52:46)
    at Object.<anonymous> (***/app/i18n.js:12:18)
    at Module._compile (internal/modules/cjs/loader.js:936:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:947:10)
    at Module.load (internal/modules/cjs/loader.js:790:32)
    at Function.Module._load (internal/modules/cjs/loader.js:703:12)
    at Module.require (internal/modules/cjs/loader.js:830:19)
    at require (internal/modules/cjs/helpers.js:68:18)
    at Object.<anonymous> (***/app/server.js:14:14)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

## Describe the solution you'd like
The path `***/app/static/locales/es/common.json` should be **`***/app/public/static/locales/es/common.json`**

## Describe alternatives you've considered
For an easy fix, we could keep as temp solution the locales at the static path.

## Additional context
...

isaachinman:
PRs welcome.

wootsbot:
@isaachinman I want to take this issue.

kachkaev:
This worked for me:

1. Move `locales` to `/path/to/project/public/locales/{LOCALE}/{NAMESPACE}.json`

2. In `/path/to/project/src/i18n.ts` :
    ```ts
    export const nextI18next = new NextI18next({
      // ...
      localePath: typeof window === "undefined" ? "public/locales" : "locales",
    });
    ```

[...]

PR #525, opened 2019-10-10
feat: Add public directory support

wootsbot:
- Add public directory support.
- Add console warning about the deprecation of static directory.

close: #523 

### notes
Add support to the example for the tests to be successful

### DIFF
--- a/examples/simple/i18n.js
+++ b/examples/simple/i18n.js
@@ -20,4 +20,5 @@ const localeSubpathVariations = {
 module.exports = new NextI18Next({
   otherLanguages: ['de'],
   localeSubpaths: localeSubpathVariations[localeSubpaths],
+  localePath: typeof window === "undefined" ? "public/locales" : "static/locales",
 })
--- a/src/config/create-config.ts
+++ b/src/config/create-config.ts
@@ -29,14 +29,25 @@ export default (userConfig) => {
   } = combinedConfig
 
   if (isServer()) {
-
     const fs = eval("require('fs')")
     const path = require('path')
+    const hasStaticDir = fs.existsSync(path.join(process.cwd(), 'static'))
+    const hasPublicDir = fs.existsSync(path.join(process.cwd(), 'public'))
+
+    let staticLegacy = 'static/locales'
+
+    if (hasStaticDir) {
+      console.warn('The static directory has been deprecated in favor of the public directory.')
+    }
+
+    if(hasPublicDir) {
+      staticLegacy = localePath
+    }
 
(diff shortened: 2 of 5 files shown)


examples/simple/i18n.js
```diff
@@ -20,4 +20,5 @@ const localeSubpathVariations = {
 module.exports = new NextI18Next({
   otherLanguages: ['de'],
   localeSubpaths: localeSubpathVariations[localeSubpaths],
+  localePath: typeof window === "undefined" ? "public/locales" : "static/locales",
```
isaachinman: This behaviour should be built in.

src/config/create-config.ts
```diff
@@ -29,14 +29,25 @@ export default (userConfig) => {
   } = combinedConfig
 
   if (isServer()) {
-
     const fs = eval("require('fs')")
     const path = require('path')
+    const hasStaticDir = fs.existsSync(path.join(process.cwd(), 'static'))
```
isaachinman: We should be checking if there are namespace files here, not just if the dir exists.

src/config/create-config.ts
```diff
@@ -29,14 +29,25 @@ export default (userConfig) => {
...
+    const hasPublicDir = fs.existsSync(path.join(process.cwd(), 'public'))
+
+    let staticLegacy = 'static/locales'
```
isaachinman: This won't work. We need to respect the `localePath` in case a user has provided a non-default value.

examples/simple/i18n.js
wootsbot: it is a question or affirmation?

src/config/create-config.ts
wootsbot: Could it be on the same line or adding another validation?

src/config/create-config.ts
wootsbot: Correct, so we just add the warning and the `localePath` change?

examples/simple/i18n.js
isaachinman: An affirmation - it doesn't make sense to require all users to perform the same check.

src/config/create-config.ts
isaachinman: Sure, it could be another `const` called `hasDefaultNamespace` or something.

src/config/create-config.ts
isaachinman: Not sure exactly what you mean.

isaachinman:
Closing in favour of #569.

PR #541, opened 2019-10-28
fix: Do not ignore routes at i18nextMiddleware level

### LINKED ISSUE #374: react-i18next:: i18n.languages were undefined or empty undefined
## Describe the bug
I get the the following error:

```console
[ wait ]  compiling ...
react-i18next:: i18n.languages were undefined or empty undefined
```

which appears only on the server (I mean in the output of the terminal, it doesn’t appear on the browser console).

"next": "^8.1.0",
"i18next": "^17.0.6",
"next-i18next": "^0.44.0",

## Steps to reproduce

I followed every step described in the readme except `keySeparator` (because I have keys like `Status.ONGOING`)

Below is my configuration:

```js
import NextI18Next from 'next-i18next'

const instance = new NextI18Next({
  defaultLanguage: 'en',
  otherLanguages: ['ru'],
  keySeparator: false
})

export default instance
```

## Expected behaviour

This doesn’t look like a critical error, though I’m wondering the cause.

## Screenshots

- 

 - Device: MacBook Pro (Retina, 13-inch, Late 2013)
 - Browser: Chrome Canary Version 77.0.3838.0 (Official Build) canary (64-bit)

## Additional context

- I’m working on a monorepo (yarn workspaces)
- I use TypeScript

isaachinman:
Fixes #374.

### DIFF
--- a/__tests__/middlewares/next-i18next-middleware.test.ts
+++ b/__tests__/middlewares/next-i18next-middleware.test.ts
@@ -53,11 +53,7 @@ describe('next-18next middleware', () => {
   it('sets up i18nextMiddleware handle on setup', () => {
     callAllMiddleware()
 
-    expect(i18nextMiddleware.handle)
-      .toBeCalledWith('i18n',
-        expect.objectContaining({
-          ignoreRoutes: expect.arrayContaining(['/_next/', '/static/']),
-        }))
+    expect(i18nextMiddleware.handle).toBeCalledWith('i18n')
   })
 
   it(`does not call any next-i18next middleware if localeSubpaths is "${JSON.stringify(localeSubpathVariations.NONE)}"`, () => {
@@ -77,6 +73,7 @@ describe('next-18next middleware', () => {
       callAllMiddleware()
 
       expect(next).toBeCalled()
+      expect(req.i18n).toBeDefined()
     })
 
     it('adds lng to query parameters and removes from url for i18next processing', () => {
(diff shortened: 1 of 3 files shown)


eric-burel:
Hi @isaachinman , I however encounter the same issue with a static build (`next export`). Fallback languages is used (en) before user language is quicly loaded (fr). Any fix so that the client wait for the language to be loaded?

isaachinman:
@eric-burel Static exports are not supported as first class by this package. Of course the user language isn't loaded during SSR - the content is static and headers/cookies aren't being checked.

eric-burel:
Thanks for the very fast answer. I guess there is no better solution indeed in static mode. I could maybe do smth so that when no language is loaded yet, we display nothing instead of fallbacking to the i18n token name for better UX (text will seem to appear but that's better than flashing the wrong text). I understand that it's outside the scope of this project though.

isaachinman:
You can configure the fallback behaviour via `react-i18next` options, if I recall correctly.

There isn't (and won't ever be) a great "truly static" solution until/unless you adhere strictly to locale subpaths. If you're interested in hearing about experimental work therein, feel free to email me.

That being said, serverless support should come this year.

Good luck!

eric-burel:
Thanks! We will probably work more thoroughly on i18n as we plan to move Vulcan framework from Meteor to Next, out-of-the-box i18n is one of our critical features. Right now I am on a project where it's more secondary so I am happy with current version of this lib. It's great despite the complexity of setting up i18n.

eric-burel:
According to this doc you can't prevent fallbacking to key, only prevent fallbacking to another language. It sounds weird but that's how it seems to work, worst case scenario is that user will see the i18n key.

PR #552, opened 2019-11-08
fix: Setting stackTraceLimit properly in createConsoleLog

### LINKED ISSUE #550: Error.stackTraceLimit not being reset if strictMode is off
## Describe the bug
createConsoleLog function in src/utils/console-message.ts sets Error.stackTraceLimit and returns before it can be set back to previous value if strictMode is false.

This breaks our build process.

v2.0.0

## Expected behaviour
Error.stackTraceLimit is not being set before we check for strictMode value

jakubjanousek:
fixes #550

### DIFF
--- a/src/utils/console-message.ts
+++ b/src/utils/console-message.ts
@@ -51,7 +51,7 @@ export default function createConsoleLog(messageType, message) {
 
   /* Temporarily set the stacktrace to 0 or errorStackTraceLimit,
      in order to only display a message */
-  (Error as any).errorStackTraceLimit = errorStackTraceLimit
+  Error.stackTraceLimit = errorStackTraceLimit
 
   // Make room for new message
   console.log()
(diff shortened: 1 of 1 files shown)


src/utils/console-message.ts
```diff
@@ -52,7 +51,7 @@ export default function createConsoleLog(messageType, message) {
 
   /* Temporarily set the stacktrace to 0 or errorStackTraceLimit,
      in order to only display a message */
-  (Error as any).errorStackTraceLimit = errorStackTraceLimit
+  Error.stackTraceLimit = errorStackTraceLimit
```
isaachinman: Why was this changed?

jakubjanousek: I have removed 2 lines
39: set Error.stackTraceLimit, but did it before check for strictMode
55: didn't do anything as far as I know, since there is no property errorStackTraceLimit on Error (hence there needed to be `as any`)

I have replaced them by line 54, which sets Error.stackTraceLimit at the correct place

PR #569, opened 2019-11-22
feat: Public dir support

### LINKED ISSUE #523: Add support for "public" directory

nathanfriemel:
Took a pass at adding public folder support, looking for feedback:

- Changed default config localePath to public/static/locales
- Updated create config to look for combinedConfig localePath and if not found try to fall back to static/locales and print a deprecation console warning in the server config; client config attempts to remove `public/` from backend paths
- Updated tests

close #523


nathanfriemel:
Didn't realize there was already a PR for this, feel free to close this one.

isaachinman:
Hi @nathanfriemel, thanks for doing this work. I haven't heard from the author of #525 in awhile, so we can proceed with this PR if you're keen.

Can you please remove all the function return typings? I appreciate you want to fix TS warnings, but that should be handled in a separate PR as it just makes it harder to review this work. Once that's sorted I will review. Thanks!

nathanfriemel:
Return types are gone.

examples/simple/package.json
```diff
@@ -14,10 +14,6 @@
   "devDependencies": {},
   "dependencies": {
     "express": "^4.16.4",
-    "i18next": "^14.0.1",
-    "next": "^9.0.3",
-    "next-i18next": "link:../../",
-    "react": "link:../../node_modules/react",
-    "react-dom": "link:../../node_modules/react-dom"
+    "next-i18next": "link:../../"
```
isaachinman: What's going on here?

package.json
```diff
@@ -68,8 +68,8 @@
     "@types/express": "^4.16.1",
     "@types/jest": "^24.0.16",
     "@types/jest-environment-puppeteer": "^4.0.0",
-    "@types/react": "^16.8.4",
-    "@types/react-dom": "^16.8.2",
+    "@types/react": "^16.9.11",
```
isaachinman: Are these @types changes necessary? If not, let's revert that please.

package.json
```diff
@@ -115,6 +115,7 @@
   },
   "peerDependencies": {
     "next": ">= 7.0.0",
-    "react": ">= 16.8.0"
+    "react": ">= 16.8.0",
+    "react-dom": ">= 16.8.0"
```
isaachinman: Why was this added?

src/config/create-config.ts
```diff
@@ -32,36 +33,47 @@ export default (userConfig) => {
...
     // Validate defaultNS
     // https://github.com/isaachinman/next-i18next/issues/358
     if (process.env.NODE_ENV !== 'production' && typeof combinedConfig.defaultNS === 'string') {
       const defaultNSPath = path.join(process.cwd(), `${localePath}/${defaultLanguage}/${combinedConfig.defaultNS}.${localeExtension}`)
       const defaultNSExists = fs.existsSync(defaultNSPath)
       if (!defaultNSExists) {
-        throw new Error(`Default namespace not found at ${defaultNSPath}`)
+        // if defaultNS doesn't exist, try to fall back to the deprecated static folder
+        // https://github.com/isaachinman/next-i18next/issues/523
+        if (fs.existsSync(path.join(process.cwd(), `${STATIC_LOCALE_PATH}/${defaultLanguage}/${combinedConfig.defaultNS}.${localeExtension}`))) {
```
isaachinman: Can you move this logic out to a variable? Something like `const staticDirExists`.

src/config/create-config.ts
```diff
@@ -32,36 +33,47 @@ export default (userConfig) => {
...
+          console.warn('Deprecation Warning - falling back to /static folder, deprecated in next@9.1.*')
```
isaachinman: We've got our own logging util in `src/utils/console-message.ts`. Please use that where possible.

src/config/create-config.ts
```diff
@@ -32,36 +33,47 @@ export default (userConfig) => {
...
     if (!combinedConfig.ns) {
       const getAllNamespaces = p => fs.readdirSync(p).map(file => file.replace(`.${localeExtension}`, ''))
-      combinedConfig.ns = getAllNamespaces(path.join(process.cwd(), `${localePath}/${defaultLanguage}`))
+      combinedConfig.ns = getAllNamespaces(path.join(process.cwd(), `${serverLocalePath}/${defaultLanguage}`))
     }
 
   } else {
 
+    // remove public/ prefix from client site config
+    const clientLocalePath = localePath.replace(/^public\//, '')
```
isaachinman: I wonder how safe this is. We should probably check via something like `localePath.startsWith('/public/')`.

examples/simple/package.json
nathanfriemel: Upgrading next to version 9.0.6+ was causing the tests to throw [this error](https://reactjs.org/docs/error-decoder.html/?invariant=321).  Seemed related to [this bug](https://github.com/zeit/next.js/issues/9022) in next, a work around was mentioned [here](https://github.com/zeit/next.js/issues/9022#issuecomment-552136320).  Wanted to test these changes against next 9.1, probably not important, could revert back to 9.0.3

isaachinman: Ah yeah, that's what I figured. I tried to upgrade myself, but ran into that bug.

I'm not understanding how this runs at all without `react` and `react-dom` as deps.

nathanfriemel:
Changing the default to `public/static/locales` would probably simplify things.  I'm pretty sure the Next blog suggests just moving the static folder into the public folder so that you don't have to update import paths across your app.  I can make that change and probably undo a lot of the test changes in the process.

isaachinman:
Sounds good. Thanks for being flexible!

src/config/create-config.ts
nathanfriemel: Will switch over.

src/config/create-config.ts
nathanfriemel: Should be a no-op if the pattern doesn't match, but couldn't hurt to check startsWith first.

src/config/create-config.ts
nathanfriemel: Will do.

package.json
nathanfriemel: Reverting

package.json
nathanfriemel: Reverting

isaachinman:
@nathanfriemel Just let me know when this is ready for review again.

nathanfriemel:
@isaachinman I think it is ready to look at again.

__tests__/config/test-helpers.ts
```diff
@@ -15,24 +15,24 @@ const userConfig = {
   defaultNS: 'universal',
   fallbackLng: 'it',
   otherLanguages: ['fr', 'it'],
-  localePath: 'static/translations',
+  localePath: 'public/translations',
```
isaachinman: Any reason why you didn't make this `public/static/translations`?

nathanfriemel: Just missed it, will update it and the tests.

nathanfriemel:
@isaachinman README updated and updated test-helpers localePath.

__tests__/config/test-helpers.ts
```diff
@@ -15,24 +15,24 @@ const userConfig = {
...
-  localePath: 'static/translations',
+  localePath: 'public/static/translations',
   localeStructure: '{{ns}}/{{lng}}',
   localeSubpaths: localeSubpathVariations.FOREIGN,
 }
 
 const userConfigClientSide = {
   ...userConfig,
   backend: {
-    loadPath: '/static/translations/{{ns}}/{{lng}}.json',
```
isaachinman: Why was this change made?

__tests__/config/test-helpers.ts
```diff
@@ -15,24 +15,24 @@ const userConfig = {
...
+    addPath: '/translations/{{ns}}/{{lng}}.missing.json',
   },
 }
 
 const userConfigServerSide = {
   ...userConfig,
   backend: {
-    loadPath: '/home/user/static/translations/{{ns}}/{{lng}}.json',
-    addPath: '/home/user/static/translations/{{ns}}/{{lng}}.missing.json',
+    loadPath: '/home/user/public/translations/{{ns}}/{{lng}}.json',
```
isaachinman: Again wondering why this isn't `public/static`?

src/utils/console-message.ts
```diff
@@ -31,9 +31,9 @@ function logMessage(messageType, message) {
  * @param {String} message
  * @param {Object} options
  */
-export default function createConsoleLog(messageType, message) {
+export default function createConsoleLog(messageType, message, config = this.config) {
```
isaachinman: Why was this change made? I suppose it's because we're trying to reference `this.config` before it exists?

yarn.lock
```diff
@@ -1900,7 +1900,7 @@ async-sema@3.0.0:
   resolved "https://registry.yarnpkg.com/async-sema/-/async-sema-3.0.0.tgz#9e22d6783f0ab66a1cf330e21a905e39b3b3a975"
```
isaachinman: Why is there a diff on the lockfile?

__tests__/config/test-helpers.ts
nathanfriemel: It should be public/static, also the create-config.test should be changed to public/static on lines 115-116.

Looks like process.env.NODE_ENV is always 'production' in the create-config.test even with `NODE_ENV=test jest` so the test never runs the `Validate defaultNS` block and serverLocalePath is just whatever the test-helpers userConfig.localePath is set to.  I'm not sure why NODE_ENV is production atm and interested to see if you see the same thing.

src/utils/console-message.ts
nathanfriemel: Using consoleMessage in create-config was throwing `Cannot read property 'config' of undefined`

yarn.lock
nathanfriemel: Now that all the package.json changes are reverted there shouldn't be any diff, I'll revert it to master.

isaachinman: Cool, thanks!

__tests__/config/test-helpers.ts
isaachinman: Yeah, it looks like it's due to [this](https://github.com/isaachinman/next-i18next/blob/e48f7e9558c341d3f1fd476e56acd5f0433cad52/__tests__/config/create-config.test.ts#L249) hacky setup.

We should instead capture the original value in a var, and reassign before the end of the test:

```jsx
const originalNodeEnv = process.env.NODE_ENV
process.env.NODE_ENV = 'production'

...perform test

process.env.NODE_ENV = originalNodeEnv
```

Modifying globals in tests is almost always problematic!

isaachinman: cc @capellini

isaachinman:
@nathanfriemel Need to resolve a couple conflicts unfortunately. But this PR is my top priority at the moment, so I won't merge anything except any potential urgent bug fixes until we get it in.

__tests__/config/test-helpers.ts
capellini: Hey @isaachinman - do you need something from me?

isaachinman: Just tagged you to notify you that we'll most likely need to update these tests. If @nathanfriemel is happy to do the work, I don't think we need anything!

capellini: Ok.  I'm here to assist, if needed. Just let me know.

isaachinman: Thanks!

nathanfriemel: Resolved the conflicts, want me to go ahead and change static to public/static?

isaachinman:
@nathanfriemel Let me know when this ready for re-review. Thanks!

__tests__/config/test-helpers.ts
nathanfriemel: Went ahead and changed to public/static

nathanfriemel:
@isaachinman Resolved conflicts, ready to look at.

src/config/create-config.ts
```diff
@@ -33,36 +34,53 @@ export const createConfig = (userConfig) => {
...
       const defaultNSExists = fs.existsSync(defaultNSPath)
       if (!defaultNSExists) {
-        throw new Error(`Default namespace not found at ${defaultNSPath}`)
+        // if defaultNS doesn't exist, try to fall back to the deprecated static folder
+        // https://github.com/isaachinman/next-i18next/issues/523
+        const staticDirPath = path.join(process.cwd(), STATIC_LOCALE_PATH, defaultFile)
+        const staticDirExists = fs.existsSync(staticDirPath)
+        if (staticDirExists) {
+          consoleMessage('warn', 'Falling back to /static folder, deprecated in next@9.1.*', combinedConfig)
+          serverLocalePath = STATIC_LOCALE_PATH
```
isaachinman: I am just now realising that this fallback will only take place in non-production modes. Wouldn't that effectively break all currently-deployed apps where people rely on the default value of `localePath` being the `static` dir?

nathanfriemel: Yeah you're right, will look at this again in the morning.

nathanfriemel: Sorry for the break, holidays and work and what not.  What was the original reason for the `process.env.NODE_ENV !== 'production'` here?  Was it to prevent the error throw in production?  I looked at the ticket in the comment and the PR for it and just want to figure out the original intent before changing anything more.

isaachinman: No problem. Yes that's correct - this `if` block is/was solely for non-production validation of the default namespace.

nathanfriemel: Moved the production check to just before the error throw, hopefully that sorts things.

isaachinman:
Thank you @nathanfriemel for all your work on this. I merged the PR and cleaned up a few things, and have released under v3.0.0. There should be no breaking changes, just wanted to be on the safe side.

PR #575, opened 2019-11-27
fix error if pathname is not valid in subpathIsPresent

iherem:
when use <Link href="#"> got error `Cannot read property 'length' of null` in subpath-is-present.ts file
So i return false if pathname is not valid.

### DIFF
--- a/src/utils/subpath-is-present.ts
+++ b/src/utils/subpath-is-present.ts
@@ -6,6 +6,7 @@ export default (url: string, subpath: string) => {
   }
 
   const { pathname } = parseUrl(url)
+  if (!pathname) return false
   return (
     (pathname.length === subpath.length + 1 && pathname === `/${subpath}`) ||
     (pathname.startsWith(`/${subpath}/`))


isaachinman:
Sorry, what is this PR for? The file you've modified is an internal until. It's not meant for end users.

iherem:
@isaachinman Thanks for reply, When I use <Link href = "#">, I get the error that says above even though the previous version didn't error.

isaachinman:
Why are you setting `href` to just a hash?

Trying to understand if this is a genuine use case.

iherem:
my use case is `<Link href={someCondition ? '/home' : '#'}>` 
I trying to disable user to change page if someCondition equal to false.
or you have any idea for this case.

isaachinman:
This seems like a more generic NextJs question of how to disable links. Can you pass `null` instead of `"#"`?

Tokenyet:
I met this issue for some reason, but I don't know where It comes. I changed the Link from [Mui's Link](https://github.com/mui-org/material-ui/blob/f7a2c421083023a9b3da84be358bfeca562e9563/examples/nextjs/src/Link.js#L6), and there is whole error in my project, I have no idea how to fix this, and where's the wrong 😕

```
<Link href={`/article/${data.article.id}`}>
    {article.name}
</Link>
```

This is the example of using the modified `Link` by Mui, but gonna go wrong with changing to `Link` provided by my `i18n`.

isaachinman:
Fixed by #602.

PR #581, opened 2019-12-04
feat: Add useRouter and withRouter

### LINKED ISSUE #316: Provide HOC withRouter to be feature identical with default nextjs method.
Hello,
the current approach to access the router inside a component isn't ideal:

```jsx
import { Router } from '~/i18n';
// in the render function
Router.pathname
```
This has several reasons:

- We can't render isomorphic without to conditonally check browser or SSR.
- Differs from `nextjs` [approach](https://github.com/zeit/next.js/#using-a-higher-order-component)
- Not compatible with [hooks](https://reactjs.org/docs/hooks-reference.html#conditionally-firing-an-effect) because only the component state can be watched for changes.

The api should look as follow:

```jsx
import { withRouter } from '~/i18n'
withRouter(MyComponent)
```

Thank you!

martpie:
- [x] `useRouter`
- [x] `withRouter`
- [x] Update example
- [x] Update README
- [x] ~Eventually update bundle size limits (I hit the error locally)~
- [x] Add tests

Fix #316

Before updating the README, maybe @isaachinman you can tell me if the rest looks good to you :)

### DIFF
--- a/src/hocs/index.ts
+++ b/src/hocs/index.ts
@@ -1,2 +1,3 @@
 export { appWithTranslation } from './app-with-translation'
 export { withInternals } from './with-internals'
+export { createWithRouter } from './with-router'
--- a/src/hocs/with-router.tsx
+++ b/src/hocs/with-router.tsx
@@ -1,0 +1,31 @@
+import React from 'react'
+import { WithRouterProps, ExcludeRouterProps } from 'next/dist/client/with-router'
+import { NextComponentType, NextPageContext } from 'next-server/dist/lib/utils'
+
+import { UseRouterHook } from '../../types'
+
+export const createWithRouter = (useRouter: UseRouterHook) => {
+  const withRouter = <
+    P extends WithRouterProps,
+    C = NextPageContext
+  >(
+    ComposedComponent: NextComponentType<C, any, P>
+  ): React.ComponentType<ExcludeRouterProps<P>> => {
+    function WithRouterWrapper(props: any) {
+      return <ComposedComponent router={useRouter()} {...props} />
+    }
+
+    WithRouterWrapper.getInitialProps = ComposedComponent.getInitialProps
+    // This is needed to allow checking for custom getInitialProps in _app
+    ;(WithRouterWrapper as any).origGetInitialProps = (ComposedComponent as any).origGetInitialProps
+    if (process.env.NODE_ENV !== 'production') {
+      const name =
+        ComposedComponent.displayName || ComposedComponent.name || 'Unknown'
+      WithRouterWrapper.displayName = `withRouter(${name})`
+    }
+
+    return WithRouterWrapper
+  }
+
+  return withRouter
+}
--- a/src/hooks/use-router.tsx
+++ b/src/hooks/use-router.tsx
@@ -1,0 +1,37 @@
+import { useEffect, useState } from 'react'
+import { SingletonRouter } from 'next/router'
+import { wrapRouter } from '../router'
+import { NextI18NextInternals, UseRouterHook } from '../../types'
+
+
+/**
+ * create a hook that need to be reactive when the Router is updated (with
+ * asPath in particular)
+ */
+export const createUseRouter = (internals: NextI18NextInternals): UseRouterHook => {
+  const useRouter = (): SingletonRouter => {
+    const [router, setRouter] = useState(wrapRouter(internals))
+
+    const updateRouter = () => {
+      setRouter(wrapRouter(internals))
+    }
+
+    useEffect(() => {
+      // TODO should routeChangeStart and routeChangeError be included, it all
+      // depends when asPath changes
+      // router.events.on('routeChangeStart', updateRouter)
+      router.events.on('routeChangeComplete', updateRouter)
+      // router.events.on('routeChangeError', updateRouter)
+
+      return () => {
+        // router.events.off('routeChangeStart', updateRouter)
+        router.events.off('routeChangeComplete', updateRouter)
+        // router.events.off('routeChangeError', updateRouter)
+      }
+    }, [])
+
+    return router
+  }
+
+  return useRouter
+}
--- a/.eslintrc
+++ b/.eslintrc
@@ -21,7 +21,6 @@
     "@typescript-eslint/explicit-function-return-type": 0,
     "@typescript-eslint/explicit-member-accessibility": 0,
     "@typescript-eslint/no-object-literal-type-assertion": 0,
-    "@typescript-eslint/explicit-function-return-type": 0,
     "padded-blocks": 0,
     "global-require": 0,
     "function-paren-newline": 0,
--- a/.eslintrc
+++ b/.eslintrc
@@ -21,6 +21,7 @@
     "@typescript-eslint/explicit-function-return-type": 0,
     "@typescript-eslint/explicit-member-accessibility": 0,
     "@typescript-eslint/no-object-literal-type-assertion": 0,
+    "@typescript-eslint/explicit-function-return-type": 0,
     "padded-blocks": 0,
     "global-require": 0,
     "function-paren-newline": 0,
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ It's recommended to export this `NextI18Next` instance from a single file in you
 
 After creating and exporting your `NextI18Next` instance, you need to take the following steps to get things working:
 
-1. Create an `_app.js` file inside your `pages` directory, and wrap it with the `NextI18Next.appWithTranslation` higher order component (HOC). You can see this approach in the [examples/simple/pages/_app.js](./examples/simple/pages/_app.js). 
+1. Create an `_app.js` file inside your `pages` directory, and wrap it with the `NextI18Next.appWithTranslation` higher order component (HOC). You can see this approach in the [examples/simple/pages/_app.js](./examples/simple/pages/_app.js).
 2. Create a `server.js` file inside your root directory, initialise an [express](https://www.npmjs.com/package/express) server, and use the `nextI18NextMiddleware` middleware with your `nextI18Next` instance passed in. You can see this approach in the [examples/simple/server.js](./examples/simple/server.js).
 3. Update the scripts in `package.json` to:
 ```
@@ -215,6 +215,20 @@ server.get('/products/:id', (req, res) => {
 
 /* Third, add catch-all GET for non-custom routes */
 server.get('*', (req, res) => handle(req, res))
+```
+
+## Access the Router object
+
+This library exposes a modified version on Next.js's `Router` object that can be exposed in various forms from the `nextI18next` object.
+
+Their use is similar to Next.js's [`Router`](https://nextjs.org/docs/api-reference/next/router#router-api), [`useRouter`](https://nextjs.org/docs/api-reference/next/router#userouter) and [`withRouter`](https://nextjs.org/docs/api-reference/next/router#withrouter).
+
+```jsx
+export const {
+  Router,
+  useRouter,
+  withRouter
+} = NextI18NextInstance
 ```
 
 ## Accessing the Current Language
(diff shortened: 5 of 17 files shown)


martpie:
ps: I could not test the if the URL is correctly changed with `/de` `/en` on the simple example as there is no custom server

examples/simple/pages/index.js
```diff
@@ -21,9 +21,15 @@ const Homepage = ({ t }) => (
           <button
             type='button'
           >
-            {t('to-second-page')}
+            {t('to-second-page')} (Link)
```
isaachinman: Any content should go into the actual translation string. I know it's a small thing, but we should do our best to promote best practices.

examples/simple/pages/index.js
```diff
@@ -21,9 +21,15 @@ const Homepage = ({ t }) => (
...
           </button>
         </Link>
+        <button
+          type='button'
+          onClick={() => { router.push('/second-page') }}
+        >
+          {t('to-second-page')} (onClick)
```
isaachinman: Might as well make this a new key.

examples/simple/pages/index.js
```diff
@@ -21,9 +21,15 @@ const Homepage = ({ t }) => (
           <button
             type='button'
           >
-            {t('to-second-page')}
+            {t('to-second-page')} (Link)
           </button>
         </Link>
+        <button
+          type='button'
+          onClick={() => { router.push('/second-page') }}
```
isaachinman: The `onClick` can just return the func call:

```jsx
onClick={() => router.push('/second-page')}
```

examples/simple/pages/index.js
```diff
@@ -38,4 +44,4 @@ Homepage.propTypes = {
   t: PropTypes.func.isRequired,
 }
 
-export default withTranslation('common')(Homepage)
+export default withRouter(withTranslation('common')(Homepage))
```
isaachinman: Curious here, does the order of these HOCs matter? Also, we might want to multiline format this.

Remember that (somewhat unfortunately) a lot of people will just be copy/pasting this example. It makes me wonder if it'd be better to split this `withRouter` functionality into a separate component, so that we can clearly divide the piece of code that is demonstrating `withRouter`.

examples/simple/pages/second-page.js
```diff
@@ -1,33 +1,43 @@
...
-const SecondPage = ({ t }) => (
-  <React.Fragment>
-    <main>
-      <Header title={t('h1')} />
-      <Link href='/'>
+const SecondPage = () => {
+  const router = useRouter()
+  const { t } = useTranslation(['second-page'])
+
+  const backToHome = () => {
```
isaachinman: No point declaring this if it's only used once - we should maintain consistency with the other page.

examples/simple/pages/second-page.js
```diff
@@ -1,33 +1,43 @@
...
+
+  return (
+    <React.Fragment>
+      <main>
+        <Header title={t('h1')} />
+        <Link href='/'>
+          <button
+            type='button'
+          >
+            {t('back-to-home')} (Link)
```
isaachinman: Same comments as the index page about translation content.

src/hocs/index.ts
```diff
@@ -1,2 +1,3 @@
 export { default as appWithTranslation } from './app-with-translation'
 export { default as withInternals } from './with-internals'
+export { default as createWithRouter } from './with-router'
```
isaachinman: If you rebase against the latest master, you'll see that we've moved everything to named exports, as discoverability and intellisense is much better. Please follow that pattern. Thanks!

src/hocs/with-router.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react'
+import { WithRouterProps, ExcludeRouterProps } from 'next/dist/client/with-router'
```
isaachinman: Is this safe/correct? Should we really be importing directly from a `/dist` path? That seems like an implementation detail which is subject to change.

src/hocs/with-router.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { UseRouterHook } from '../../types'
+
+const createWithRouter = (useRouter: UseRouterHook) => {
+  const withRouter = <
+    P extends WithRouterProps,
+    C = NextPageContext
+  >(
+    ComposedComponent: NextComponentType<C, any, P>
+  ): React.ComponentType<ExcludeRouterProps<P>> => {
+    function WithRouterWrapper(props: any) {
```
isaachinman: Are you writing this as a `function` component to give it its own scope?

src/hocs/with-router.tsx
```diff
@@ -0,0 +1,35 @@
...
+      return <ComposedComponent router={useRouter()} {...props} />
+    }
+
+    WithRouterWrapper.getInitialProps = ComposedComponent.getInitialProps
+    // This is needed to allow checking for custom getInitialProps in _app
```
isaachinman: Please write comments in this format:

```
/*
  Comment goes here
*/
```

src/hocs/with-router.tsx
```diff
@@ -0,0 +1,35 @@
...
+    ;(WithRouterWrapper as any).origGetInitialProps = (ComposedComponent as any).origGetInitialProps
```
isaachinman: I don't understand what `origGetInitialProps` is doing here, as it is not referenced anywhere else. Is this a NextJs internal thing?

src/hocs/with-router.tsx
```diff
@@ -0,0 +1,35 @@
...
+      WithRouterWrapper.displayName = `withRouter(${name})`
+    }
+
+    return WithRouterWrapper
+  }
+
+  return withRouter
+}
+
+
```
isaachinman: Too many empty lines. We really need to set up prettier!

src/hooks/use-router.tsx
```diff
@@ -0,0 +1,39 @@
+import { useEffect, useState } from 'react'
```
isaachinman: Please follow the comment format convention in this file as well.

isaachinman: Does this file need to be `.tsx`? I don't see any JSX.

src/hooks/use-router.tsx
```diff
@@ -0,0 +1,39 @@
...
+const createUseRouter = (internals: NextI18NextInternals): UseRouterHook => {
+  const useRouter = (): SingletonRouter => {
+    const [router, setRouter] = useState(wrapRouter(internals))
+
+    const updateRouter = () => {
+      setRouter(wrapRouter(internals))
+    }
+
+    useEffect(() => {
+      // TODO should routeChangeStart and routeChangeError be included, it all
```
isaachinman: What is blocking this?

src/hooks/use-router.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { NextI18NextInternals, UseRouterHook } from '../../types'
+
+
+/**
+ * create a hook that need to be reactive when the Router is updated (with
+ * asPath in particular)
+ */
+const createUseRouter = (internals: NextI18NextInternals): UseRouterHook => {
+  const useRouter = (): SingletonRouter => {
+    const [router, setRouter] = useState(wrapRouter(internals))
```
isaachinman: Do we definitely need state for this?

examples/simple/pages/index.js
martpie: If an HoC is well written, the order should not matter, for typings as well. It's a bit tricky to get right, but I adapted here the `withRouter` from `next`, so I am quite confident with it.

Regarding splitting, I can do that yes

src/hooks/use-router.tsx
martpie: Probably not indeed

src/hooks/use-router.tsx
martpie: The question is: do we need to trigger the `useRouter` hook on `routeChangeStart` and `routeChangeError`, I would guess no, `routeChangeComplete` should be enough, but I am not a 100% sure, maybe @timneutkens could answer this question 🙏 😇

src/hooks/use-router.tsx
martpie: The router needing to be reactive, I'd say yes indeed

src/hocs/with-router.tsx
martpie: Upstream issue: https://github.com/zeit/next.js/issues/9424 

I need to submit a patch to Next.js, but I think we can live with this in the meantime. I'll add a comment.

src/hocs/with-router.tsx
martpie: Well, it's usually how HOCs are made: it's a function returning a function

src/hocs/with-router.tsx
martpie: The initial implementation can be seen here: https://github.com/zeit/next.js/blob/canary/packages/next/client/with-router.tsx

I literally stole it and adapted the imports to use `next-i18next`'s `useRouter`

src/hooks/use-router.tsx
martpie: What is wrong exactly? ESLint does not show any error

martpie:
> In general, I am somewhat unfamiliar with react-i18next usage with hooks, as I haven't had a chance to use them in a real project.

We are more talking here of Next.js usage of hooks ;) `react-i18next` has not much to do with this PR

examples/simple/pages/index.js
isaachinman: I agree, which is why I was asking. Thanks for clarify. If you've based this on NextJs source code, it would be nice to have a link in a comment.

isaachinman: Another thought: is it definitely not possible to just reuse the NextJs source? At every step of the way with this package, I have tried to avoid taking on any burden of maintenance if at all possible.

src/hocs/with-router.tsx
isaachinman: Interesting. Thanks.

src/hocs/with-router.tsx
isaachinman: Okay, understood. But I don't see `origGetInitialProps` being used anywhere. What part have I missed?

src/hooks/use-router.tsx
isaachinman: Ah yes, sorry about that. I'm not actually sure if it's even possible to set prettier to check comment formatting. I recently changed all commenting in this repo to follow this format:

```
/*
  Comment goes here
*/
```

src/hooks/use-router.tsx
isaachinman: Is this also coming from NextJs source somewhere? If so, same points/comments as above.

isaachinman:
@martpie Sorry yes of course, that's what I meant. Let's see if we can get anyone knowledgeable/well-placed from the community to act as a second pair of eyes here.

src/hooks/use-router.tsx
martpie: I simplified the implementation, should be more straightforward now. I can still add comments if you feel it needs some.

isaachinman:
@martpie Worryingly, the e2e tests had failed on Node 8. It didn't look related to the changes here, so I reran the workflow and it passed. However, this is the first time I've _ever_ seen our test suite behave in a flaky way. Any idea what might be causing that?

martpie:
Mmmh, I cannot see the tests results (CircleCI gives me a 404) 🤔

martpie:
I still need to write tests, I'll try to do that this week-end

isaachinman:
Have got another example of this new flakiness on master [here](https://circleci.com/gh/isaachinman/next-i18next/1262?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link). Not sure what has caused this yet, but it looks unrelated to the work here.

martpie:
At some point, you just have to get it done, here we are.

Regarding the documentation/README, you'll know better than me how to update it properly :)

isaachinman:
Thanks! I can admin merge if appropriate. I will aim to review this soon!

isaachinman:
@martpie There are a few things I'd like to change here but I am happy to take over the work. What do you think about merging this onto the `use-router-and-with-router` branch of this repo to facilitate that?

martpie:
Sure! How could I change the target branch?

PR #592, opened 2019-12-19
chore: Add sponsor notice

isaachinman:


kachkaev:
`npm`no longer recommends using `postinstall` script for sponsorship information. There is a new [`funding` field](https://github.com/npm/rfcs/blob/latest/accepted/0017-add-funding-support.md) in `package.json`, which should be used instead.

Post-install scripts are disabled by some developers (including myself) because they are a can of worms security-wise. Related discussion: https://github.com/yarnpkg/yarn/issues/7761#issuecomment-565493023

kachkaev:
Example migration from a postinstall script: https://github.com/styled-components/styled-components/pull/2931

isaachinman:
Yup, I've just remembered a blog post I saw a few weeks ago and am already in the process of updating.

PR #607, opened 2020-01-07
chore: update i18next to v19.0.3

ratson:
Update to latest i18next

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -9,7 +9,7 @@ import {
 } from 'react-i18next'
 import { LinkProps } from 'next/link'
 import { SingletonRouter } from 'next/router'
-import i18next from 'i18next'
+import * as i18next from 'i18next'
 
 export type InitConfig = {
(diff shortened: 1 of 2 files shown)


types.d.ts
```diff
@@ -9,7 +9,7 @@ import {
 } from 'react-i18next'
 import { LinkProps } from 'next/link'
 import { SingletonRouter } from 'next/router'
-import i18next from 'i18next'
+import * as i18next from 'i18next'
```
isaachinman: This shouldn't be necessary with `esModuleInterop` set to `true`. Am I misunderstanding something?

ratson: I got error without it

isaachinman: We'll also need lockfile changes. I'll handle this upgrade separately. Thanks for bringing it to my attention.

isaachinman:
Fixed by #608.

PR #616, opened 2020-01-16
feat: Add initPromise to NextI18Next class

### LINKED ISSUE #564: Awaiting for initialisation of next-i18next instance
Hi @isaachinman 👋 

Referring to the next issue https://github.com/isaachinman/next-i18next/issues/374#issuecomment-529196313 specifically this comment:

> Finally took some time to dig into this. The issue of awaiting the initialisation of the i18next instance is a very trivial one, and I can provide an easy, non-breaking way for users to do that if they want to. However, I think that that race condition is causing a very low percentage of the error reports we are seeing here.

Would you mind on sharing the approach of awaiting the initialisation of `next-i18next` instance, please ?

We're having some issues with it and we haven't found a way to await it or initialise it synchronously 

Thanks 🙏

isaachinman:
Adds an `initPromise` property to the NextI18Next class which can then be awaited if desired:

```jsx
await nextI18next.initPromise
```

Fixes #564.

### DIFF
--- a/__tests__/create-i18next-client.test.ts
+++ b/__tests__/create-i18next-client.test.ts
@@ -5,7 +5,7 @@ import createI18nextClient from '../src/create-i18next-client'
 const i18nextMiddleware = require('i18next-express-middleware')
 
 jest.mock('i18next', () => ({
-  init: jest.fn(),
+  init: jest.fn(() => new Promise(resolve => setTimeout(() => resolve(), 0))),
   use: jest.fn(),
 }))
 
@@ -24,6 +24,15 @@ describe('initializing i18n', () => {
     (i18next.init as jest.Mock).mockClear();
     (i18next.use as jest.Mock).mockClear()
     i18nextMiddleware.LanguageDetector.mockClear()
+  })
+
+  it('should return both the i18n object and an initPromise', async () => {
+    const { i18n, initPromise } = createI18nextClient({
+      use: [],
+      customDetectors: [],
+    })
+    expect(typeof i18n.isInitialized).toBe('boolean')
+    expect(typeof initPromise.then).toBe('function')
   })
 
   it('should not initialize i18n if i18n is already initialized', () => {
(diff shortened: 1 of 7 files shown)


toinelin:
Hi ! Does the `initPromise` addition was added to the last release ? If not, when will it be available ? Thank you ! :)

isaachinman:
Hi @toinelin, I've just published it with v4.1.0. Enjoy!

totalolage:
> Hi @toinelin, I've just published it with v4.1.0. Enjoy!

How can `initPromise` be accessed in version 8? I see it in the package source code but can't find a way to get to it that isn't insanely hacky.

isaachinman:
@totalolage In what use case are you needing to await `initPromise`?

totalolage:
@isaachinman 
For [i18next-hmr](https://github.com/felixmosh/i18next-hmr/tree/master/examples), it requires passing the `i18n` instance for initialisation.

isaachinman:
@totalolage There is a global `i18n` instance exported directly:

```tsx
import { i18n } from 'next-i18next'
```

I am mostly unaware of the `i18next-hmr` package – perhaps @felixmosh can let us know if a `next-i18next@v8` example is available?

felixmosh:
I've tested my lib with `next-i18next@v8` with the `i18n` global instance... it works on the server side, but won't work on client side, since `i18next` doesn't uses `backend` on client side in order the fetch jsons anymore. 
`i18next-hmr` needs backend config for client side in order to work.

isaachinman:
@felixmosh If you're using `getServerSideProps` or `getStaticProps`, wouldn't HMR just work "automatically" via the normal NextJs process?

felixmosh:
It won't works, I guess that it somehow related to the fact that next doesn't reload the page when json files are change.

totalolage:
@felixmosh I was able to get it to work by waiting for `i18n` to not be `null` in an ugly `setTimeout` loop. I would like to be able to just await the internal promise though.

Issue #626, opened 2020-01-26
Aggressive caching makes development hard


wasd171:
## Describe the bug
As far as I could understand from reading other issues like https://github.com/isaachinman/next-i18next/issues/447, https://github.com/isaachinman/next-i18next/issues/369, https://github.com/isaachinman/next-i18next/issues/282, and https://github.com/isaachinman/next-i18next/issues/169 this package loads the translations from the filesystem when the Next.js server is started and keeps them in memory. While this is a good strategy for production it makes development painful – one needs to restart the whole server in order to see an updated translated content.

This could be partially handled by returning `namespacesRequired: []` during the development, in this case the server returns the page with the missing content, but the browser fetches updated translations client-side. Unfortunately the `defaultNS` from the config is always cached, so one needs something like `defaultNS: 'empty'` to work around this behaviour.

## Occurs in next-i18next version
Commit `484450d`

## Steps to reproduce
```
git clone https://github.com/isaachinman/next-i18next.git
cd next-i18next
yarn
yarn build
cd examples/simple
yarn
yarn dev
```

+ Open http://localhost:3000/
+ Now assuming that your system/browser locale is not German, edit the `description` in the `public/static/locales/en/footer.json`
+ Refresh the page – no content has changed
+ Edit the `Homepage.getInitialProps` in `pages/index.js` to return `namespacesRequired: ['common']`
+ Refresh the page – the footer's content would get updated and during the short window between the initial load and your browser recreating the page on the client side you will see a `"description"` instead of any string that you have specified under the `description` key in the `public/static/locales/en/footer.json`

## Expected behaviour
I would expect the `next-i18next` not to attempt to cache the translations while Next.js is running in a development mode and always load the fresh version from the filesystem.

## OS (please complete the following information)
 - Device: MacBook Pro (Retina, 13-inch, Early 2015), macOS 10.14.6 (18G1012)
 - Browser: Chrome 79.0.3945.130, Safari 13.0.4 (14608.4.9.1.4), Firefox 72.0.2

isaachinman:
As you've seen in other issues, it is quite easy to set up a watch/restart on your locale assets. Building some sort of sophisticated/opinionated hot reload within this package is not a priority at this time. Keep in mind some users consume their assets via network/cdn.

isaachinman:
If you'd like to put forward a proof of concept PR, feel free.

iDuuck:
[...]

PR #630, opened 2020-01-29
fix: Add a simple cache invalidation in dev mode

wasd171:
This pull request addresses https://github.com/isaachinman/next-i18next/issues/626

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,13 @@ const SomeButton = () => (
 )
 ```
 
+### 6. Performance considerations
+
+In order to provide a better developer experience this package always attempts to load the fresh translations
+if your enviromental variable `NODE_ENV` does not equal to `production`. This allows to always see the current
+version of the translated content. However it also means that if you are not running with `NODE_ENV=production` in
+the production environment you might encounter performance issues.
+
 ## Custom Routing
 
 Custom routing can be achieved via the `app.render` method:
--- a/src/hocs/app-with-translation.tsx
+++ b/src/hocs/app-with-translation.tsx
@@ -3,6 +3,7 @@ import { withRouter } from 'next/router'
 
 import hoistNonReactStatics from 'hoist-non-react-statics'
 import { I18nextProvider, withSSR } from 'react-i18next'
+import {i18n as I18Next} from 'i18next'
 
 import { lngFromReq, lngPathCorrector, lngsToLoad } from '../utils'
 import { NextStaticProvider } from '../components'
@@ -28,6 +29,10 @@ export const appWithTranslation = function (WrappedComponent) {
     namespaces
       .filter(ns => !i18n.hasResourceBundle(lng, ns))
       .map(ns => i18n.reloadResources(lng, ns)),
+  )
+
+  const forceLoadNamespaces = (lng: string, namespaces: string[], i18nextInstance: I18Next) => Promise.all(
+    namespaces.map(ns => i18nextInstance.reloadResources(lng, ns))
   )
 
   class AppWithTranslation extends React.Component<Props> {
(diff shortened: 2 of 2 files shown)


isaachinman:
Ah, I wasn't expecting actual imperative use of `i18nextInstance.reloadResources`. I thought we were talking about cache invalidation and a hard browser reload?

wasd171:
I meant that `i18n.services.resourceStore.data` acts as a certain cache and needs to be force updated on every network request while running in dev mode

isaachinman:
So, a few thoughts: I'd like this to be a config-level option, eg `config.reloadResourcesDev` which is disabled by default. Also, I'd like the footprint of this option to be as small as possible. I think we can achieve this with a one liner in [the middleware](https://github.com/isaachinman/next-i18next/blob/master/src/middlewares/next-i18next-middleware.ts#L49), as `reloadResources` called without any args will reload all resources.

Finally, we'd need a couple tests to cover this. Thanks!

wasd171:
I think that we can't put it in the middleware level. Imagine the following case: the developer
+ loads page A
+ navigates to page B
+ updates content of the page A
+ navigates back to page A
If we have it in the middleware, a force reload of the page will be needed, while having it in the `App.getInitialProps` would properly reload it automatically.

isaachinman:
Those sort of considerations are why I am very hesitant to support this functionality in the first place. It needlessly complicates something that should probably just live in userland via `nodemon` anyways. This package never intended to support "live editing". Perhaps you could put together an add-on package and maintain that separately?

wasd171:
I tend to disagree. This is a not an attempt to create a sophisticated hot reload solution but rather an attempt to provide an experience expected by default – page reload or navigation show the updated content. By moving it to the middleware level you on purpose hurt the developer experience and I cannot see a good reason why. What do you mean exactly when talking about a small footprint? Is it the number of lines (can be reduced but that would probably damage the readability of the code a bit) or the performance overhead (in this case we should not load all translations and namespaces – plus this also leads to the situations where it is possible for the developer to introduce bugs invisible on the dev but appearing on the prod)? I see a possibility to remove developers' confusion in 6 lines of code if we inline the `forceLoadNamespaces` function and 1 line if we really want to push it to the minimum.

Unfortunately this can't go to an add-on, I think that `next-i18next` does not support a plugin system of its own (correct me on this if I'm wrong) and the Next.js setup (`req.i18n` vs `this.i18n`) is too specific to be one of the `i18next` plugins.

isaachinman:
> page reload or navigation show the updated content

Those are two separate things in a SPA. A "live reload" setup within a SPA that does not perform hard document requests is "sophisticated" in my book. I'd rather not have this feature touch the `app-with-translation` HOC, as the cognitive complexity therein is already too high.

To put it simply: if your use case requires reloading of translations from the filesystem in SPA navigation, you should probably just roll it yourself. The only real world use case I can see for that type of functionality is live editing.

wasd171:
However you are already doing reloading of some translations in SPA navigation! Let's check the following case:
+ `examples/simple` is running
+ you go to the `index.js` page
+ update the `second-page.json`
+ navigate to the `second-page.js`
+ the SPA would load an updated `second-page.json`

Given that this behaviour is already present in the `app-with-translation` I do not see a reason to not update it to handle the case of the dev environment

blake-mealey:
I definitely agree with @wasd171 - I have just started using this project and I found it very frustrating having to restart the server every time. I started using nodemon as a workaround but it's still a ton of overhead to see a simple change when Next.js has such a great dev experience otherwise.

blake-mealey:
I can see the argument for making it config-level since not all users use NODE_ENV=production (although that seems to be pretty standard). It should be called out in the docs early on though so that it's easy for devs to find

isaachinman:
I believe it'd be possible to write this as a one-liner via `i18n. reloadResources` in the HOC.

The goal of this package has never been to main a "great" dev experience on par with NextJs. I'm the sole maintainer and rely on contributions by the community, but am also very wary of scope creep. Using nodemon to restart your server when you change localisation content in development is a perfectly reasonable solution.

PR #646, opened 2020-03-03
docs: explicit the need for `getInitialProps` for stateless app

eric-burel:
It's a trap for new users. It is not immediate [in Next official doc for custom app](https://nextjs.org/docs/advanced-features/custom-app) that we need to add `getInitialProps` for i18n, all the more that it disable automatic static optimization (it frightens beginners to uncomment the code :) )


isaachinman:
Please add a semantic title. Otherwise looks good - thanks!

eric-burel:
Done, thank you

Issue #20, opened 2018-12-01
Setting <html lang="..."> by default


kachkaev:
I'm wondering if there is any way to implicitly set the `lang` attribute to the html tag. This is a good practice, which improves SEO and a11y.

Currently, my workaround is roughly this:

```js
// ./pages/_document.js
import Document, { Head, Main, NextScript } from 'next/document'

export default class MyDocument extends Document {
  render() {
    return (
      <html lang={this.props.__NEXT_DATA__.props.initialLanguage}>
        <Head />
        <body>
          <Main />
          <NextScript />
        </body>
      </html>
    )
  }
}
```

If there is no way to inject `lang` automatically, shall it be mentioned in the example or plugin docs?

isaachinman:
Hi @kachkaev, I don't think that's a "workaround" - it seems like a reasonable solution to me.

In the NextJs paradigm, if you need to access attributes on the `html` tag directly, creating a `_document.js` component is the only way to do so.

How did you expect this to work otherwise?

It might be good practice, but I don't think it belongs in the example as it adds unnecessary complexity - a `_document.js` file is not necessary to get this plugin working.

We can definitely link to this issue from the README, though.

kachkaev:
> How did you expect this to work otherwise?

I'm not sure, to be honest. Ideally, I would expect `next-i18next` to just add `lang` attribute via some API magic 😅 If you think that this is impossible, I'm happy to look at my "workaround" as at a "reasonable solution".

Perhaps, mentioning the importance of `<html lang="..." />` in README is worth it. Feel free to close this issue whenever you feel comfortable. 😉

isaachinman:
Added with 16610ca.

isaachinman:
@stephanemombuleau Please provide a reproducible example (repository). Once you provide that, I can attempt to help you.

This is most likely user error as the approach above works just if you add it to the [simple example](https://github.com/isaachinman/next-i18next/tree/master/examples/simple), and serialisation of `__NEXT_DATA__` is how this whole package works.

stephanemombuleau:
It's all good now! Turns out I declared a namespace but forgot to create the files. I don't know if that's the wanted behavior though but I'll let you be the judge of that! Thanks anyway :)

felixmosh:
This is my solution, use the `lng` from request object, and pass it as a prop to Document component.

```
export default class MyDocument extends Document<IDocumentProps> {
  public static async getInitialProps(ctx) {
[...]

PR #667, opened 2020-03-30
feat: Change <html>'s lang attribute when changeLanguage is called

martpie:
Following https://github.com/isaachinman/next-i18next/issues/20#issuecomment-605944899

One thing that I'm a bit hesitant with is that the `lang` attribute will be set on language change, but won't be set by default.

Maybe the documentation should be more opinionated about how to set `lang` server-side for the first render.


src/hocs/app-with-translation.tsx
```diff
@@ -44,6 +44,10 @@ export const appWithTranslation = function (WrappedComponent) {
           if (i18n.initializedLanguageOnce && typeof newLng === 'string' && prevLng !== newLng) {
             const { as, href } = lngPathCorrector(config, { as: asPath, href: routeInfo }, newLng)
             router.replace(href, as, { shallow: config.shallowRender })
+
+            // Update <html>'s lang attribute
+            const html = document.querySelector('html')
+            if (html) html.setAttribute('lang', newLng)
```
isaachinman: Could we put this into an actual block, please?

examples/simple/__tests__/e2e/basic.test.js
```diff
@@ -50,6 +55,9 @@ describe('example project', () => {
       await expect(page).toClick('button', { text: 'Wechseln Locale' })
 
       await expect(page).toMatch('A simple example')
+
+      const langAttribute = await page.evaluate('document.querySelector("html").getAttribute("lang")')
+      await expect(langAttribute).toBe(testLanguage)
```
isaachinman: This is great, but it seems like something which should be covered by unit tests as well. Is that possible?

src/hocs/app-with-translation.tsx
martpie: sure

examples/simple/__tests__/e2e/basic.test.js
martpie: I am not sure how this could be unit-tested as it's a DOM modification 🤔

isaachinman: Ah, you'd mock/assert on the `document` methods. It can be a bit tricky with these methods, though. Let me know if you have an issue?

martpie: So testing it is one thing, but what is the point if Puppeteer tests pass? Am I missing a usecase or something?

isaachinman: Is your question about why we would want to cover a feature with two sets of tests?

In general I'd prefer this functionality to be covered by unit tests, not e2e tests, but if it's in both, that's just an added bonus.

martpie: Well, it sounds a little bit like you wanted to test

```
 it(`should display h1 in ${testLanguageSecondary} locale`, async () => {
  await expect(page).toMatch('Ein einfaches Beispiel')
})
```

as a unit-test too.

Anyway, I can see what I can do, this feels like a dupe test to me

martpie: Maybe I can do it with `jsdom` 🤔

isaachinman: Ah no, I just meant a unit test asserting that `html.setAttribute` was called one time with the correct arguments. Indeed, this is a use case for `jsdom`.

martpie: Ah wait, I may have misunderstood 🤔 

So 

```
describe('changeHtmlLangAttribute', () => {
  beforeEach(() => {
    document.body.innerHTML = '<div>Hello World</div>'
  })

  it('should change <html> lang attribute based on its first parameter', () => {
    changeHtmlLangAttribute('de')
    expect(document.querySelector('html').getAttribute('lang')).toBe('de')

    changeHtmlLangAttribute('fr')
    expect(document.querySelector('html').getAttribute('lang')).toBe('fr')
  })
})
```

is not needed? 😅

martpie: I think the build just failed, but for a puppeteer reason, re-running it should fix it

isaachinman: I'm rerunning the workflow now. Asserting on `document.querySelector('html')` is indeed better if you can figure that out with `jsdom`. Otherwise asserting on `setAttribute` is OK.

martpie: 👀 ✅

martpie:
> I hate to suggest it, but perhaps we should put this functionality behind a config bool with a default value of false.

should I add that?

isaachinman:
Really not quite sure. It'd be nice if we could set the `lang` automatically for users during SSR. While it's technically a bug to not update `lang` after client side changes, I doubt it's affecting very much. The initial doc is what will impact crawlers and search engines, etc.

isaachinman:
I will be releasing `next-i18next@beta` very shortly, wherein NextJs handles the `lang` attribute internally, thus making this PR obsolete. Thanks @martpie!

martpie:
Purr-fect! 👌

Issue #580, opened 2019-12-02
Remove core-js dependency


schneck:
The core-js dependency collides with a more recent core-js 3 version, if present.
In addition, it looks like the dependency was not needed any more. I can't find any usage in the whole project except in the package file

isaachinman:
As a package, we have no control what dependencies a user may or may not install.

The dep is being used [here](https://github.com/isaachinman/next-i18next/blob/88033719baf46dd7bc0ca02f0cc35b0e98bce8af/.babelrc#L26).

bitworking:
I think a library should not include any polyfills. Because you don't know which devices are targeted.

isaachinman:
We target the same that NextJs does out of the box, because this is a NextJs plugin.

bitworking:
NextJs has "@babel/runtime-corejs2" as dependency which is not a problem.

isaachinman:
@bitworking If you can put forth a PR that removes polyfills but retains IE11 support, it will be merged immediately.

bitworking:
I want to put a PR but the Pre-commit-hook doesn't let me commit since the latest update. I get following errors:

node_modules/i18next-express-middleware/index.d.ts(9,7): error TS2687: All declarations of 'i18n' must have identical modifiers.
node_modules/i18next-express-middleware/index.d.ts(9,7): error TS2717: Subsequent property declarations must have the same type.  Property 'i18n' must be of type 'i18n & { options: Config; }', but here has type 'i18n'.
types.d.ts(65,7): error TS2687: All declarations of 'i18n' must have identical modifiers.

Can it be that you don't run those checks by yourself?

isaachinman:
@bitworking If I clone this repository, pull latest `master`, and do `yarn && yarn lint && yarn test`, everything passes.

No, I do not commit or push with `--no-verify`. Moreover, we have CI that runs on each commit/PR.

Please provide more details.

bitworking:
@isaachinman ok thanks. I'll try again. Perhaps I have another TypeScript version locally..

PR #669, opened 2020-03-30
chore: Drop Babel to compile the package with tsc

martpie:
**This is an exploration** more than a hard wish to see this merged.

Background story: we're making a lot of use of `next-i18next` at my company. We are creating a new boilerplate to start our projects, and this package is part of it.

We try to not pollute bundles too much, and there I saw this:

Screen Shot 2020-03-30 at 12 55 47

Related to https://github.com/isaachinman/next-i18next/issues/152 and https://github.com/isaachinman/next-i18next/issues/580

So I had a look, and saw you were building the package with Babel, which was including polyfills for you.

It felt a bit weird for me, as the packaged should already be transpiled to IE 11 code without adding polyfills by itself.

As some parts of the packages are TypeScript, I tried to figure out what could be the downside of dropping Babel and the TypeScript Compiler instead, while targeting ES3 for the default export (could be ES5 too I guess).

Here's a first draft. Everything seems to be working except one test that broke (related to Classes).

Now my question is: should I keep investing time in this direction to remove the need for `core-js@2` (Next.js comes with its own polyfill system now too, since 9.3)? Or is it something you are not considering/feel is not the right approach?

I may be missing some part of the story here.

Anyway, if this is something you're not interested into, I'll just close the PR.


isaachinman:
> It felt a bit weird for me, as the packaged should already be transpiled to IE 11 code without adding polyfills by itself.

Not sure what you mean by this. We can transpile to an ECMAScript target, but there are specific language features which vary in support - that's the point of the polyfills in this case.

In general I do not hold strong opinions about the build process. The only requirement is that `next-i18next` fully maintains the same level of browsers support that NextJs does, and last I checked this included IE11.

martpie:
> but there are specific language features which vary in support

do you have a concrete example? 🤔

isaachinman:
@martpie It's been awhile since I worked in this space, but I would suggest you just clone the repo, remove `corejs` and `useBuiltIns`, run the example app in prod mode, and see what breaks in IE11.

As a side note, the `es` target is specifically provided for people that want to opt out of this build process and would rather consume "current NodeJs" code and/or transpile on their own.

martpie:
> I would suggest you just clone the repo, remove corejs and useBuiltIns, run the example app in prod mode, and see what breaks in IE11.

sounds reasonable, I will try that!

PR #695, opened 2020-05-02
fix: i18next-http-backend/cjs

isaachinman:


adrai:
@isaachinman there's a chance with version 1.0.12 of i18next-http-backend it's not necessary anymore to import with /cjs...
Is there a fast way I can test this? Or you just tell me if it works or not? ;-)

isaachinman:
@adrai I think you'd need to test on the Vercel platform, as that's where we ran into issues.

isaachinman:
https://github.com/isaachinman/next-i18next/issues/274#issuecomment-622948743

adrai:
ok, no luck...
the build on Vercel worked:
image
but not at runtime:
image
:-(

PR #698, opened 2020-05-03
fix: Modified the example app and the instructions as per issue 687

### LINKED ISSUE #687: Incorrect setup instructions
## Describe the bug
In step 3 you suggest to setup i18n config using `import` type of modules which does not work by default in express server.  
```
import NextI18Next from 'next-i18next'

  defaultLanguage: 'en',
  otherLanguages: ['de']
})

export default NextI18NextInstance
```
Your own simple example uses `require` type of modules but as you mention in the comments we can't just copy paste it.  Even if we switch on the experimental feature for node.js to support `import` modules, the example still breaks as your `server.js`  file uses old `require` modules.  So you need a bit more hassle with the example and the instructions.

I would suggest to change the step 3 to:
```

  defaultLanguage: 'en',
  otherLanguages: ['de'],
})
```

Also I would appreciate if you add in your setup instruction a bit of reasoning information. It is not completely clear why you decided to go with the custom express server instead of loading translations in `getInitialProps` of `_app.js` file.
Is it only for the custom routing feature?

slava-lu:
fixes #687 
Modified the example app and the instructions so that they can be used without modification in any environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ If translations are not found in `config.localePath` or `public/static/locales` 
 
 ### 3. Project setup
 
-The export of `next-i18next` is a class constructor, into which you pass your config options. The resulting class has all the methods you will need to translate your app:
+The default export of `next-i18next` is a class constructor, into which you pass your config options. The resulting class has all the methods you will need to translate your app:
 
 ```jsx
 const NextI18Next = require('next-i18next').default
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ const NextI18Next = require('next-i18next').default
 
 module.exports = new NextI18Next({
   defaultLanguage: 'en',
-  otherLanguages: ['de']  
+  otherLanguages: ['de']
 })
 ```
 
--- a/README.md
+++ b/README.md
@@ -48,23 +48,15 @@ If translations are not found in `config.localePath` or `public/static/locales` 
 
 ### 3. Project setup
 
-The default export of `next-i18next` is a class constructor, into which you pass your config options. The resulting class has all the methods you will need to translate your app:
+The export of `next-i18next` is a class constructor, into which you pass your config options. The resulting class has all the methods you will need to translate your app:
 
 ```jsx
-import NextI18Next from 'next-i18next'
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -48,23 +48,15 @@ If translations are not found in `config.localePath` or `public/static/locales`
 
 ### 3. Project setup
 
-The default export of `next-i18next` is a class constructor, into which you pass your config options. The resulting class has all the methods you will need to translate your app:
+The export of `next-i18next` is a class constructor, into which you pass your config options. The resulting class has all the methods you will need to translate your app:
```
isaachinman, line 51: It is incorrect to remove this. We _are_ talking about the default export (as opposed to a named export).

examples/simple/i18n.js
```diff
@@ -21,3 +21,21 @@ module.exports = new NextI18Next({
   otherLanguages: ['de'],
   localeSubpaths: localeSubpathVariations[localeSubpaths],
 })
+
+
+/*
+  Uncomment the code below to run the example in your local environment
```
isaachinman, line 27: If we have this in the README, we don't need to duplicate it here.

README.md
slava-lu: Ok

examples/simple/i18n.js
slava-lu: I changed it but my idea was that you can actually copy paste the code from example app. This is how next.js examples work.

README.md
```diff
@@ -51,20 +51,12 @@ If translations are not found in `config.localePath` or `public/static/locales`
...
 
 ```jsx
-import NextI18Next from 'next-i18next'
+const NextI18Next = require('next-i18next').default
 
-const NextI18NextInstance = new NextI18Next({
+module.exports = new NextI18Next({
   defaultLanguage: 'en',
-  otherLanguages: ['de']
+  otherLanguages: ['de']
```
isaachinman, line 58: Just a small thing - we don't need or want whitespace here.

slava-lu: done

PR #707, opened 2020-05-10
chore: update dependencies with core-js to 3

shunkakinoki:
Hi, thanks for this awesome plugin!

I realized the plugin was causing outdated core-js related errors, 

so upgraded the dependencies with `yarn upgrade --latest` and core-js to 3. 

also some minor upgrades refactorings regarding the tests to pass


isaachinman:
Hi, what's the idea here? I see numerous changes that are not necessarily related. The `core-js` dep was pinned to v2 intentionally, see #414.

shunkakinoki:
sorry, accidentally updated all outdated packages too.
closing the PR,
but any plans for updating to v3?

isaachinman:
We'll update it as soon as it makes sense to. Feel free to check out the output of the simple example via webpack bundle analyzer.

PR #719, opened 2020-05-18
feat: allow config.otherLanguages to be empty

### LINKED ISSUE #606: Allow config.otherLanguages to be empty
Currently when I create a new instance of NextI18Next with the `otherLanguages` option not set (or set to an empty array), I get the following error in my console:

    To properly initialise a next-i18next instance you must provide one or more locale codes in config.otherLanguages.

The app requires exactly one locale to start with – additional locales will not be rolled out until a few months after launch. Nevertheless, I still have to deliver a fully internationalised app so my client can add the translations later on, as and when required.

Is there any way around this restriction?

mweibel:
fixes #606.

There are others who have requested it plus i18next does work without such a check - why a library/plugin like next-i18next would add such a "validation" without a real need doesn't make much sense to me. 
Anyhow, simple to fix. Thanks for making next-i18next 🙂

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -32,9 +32,6 @@ export default class NextI18Next {
     this.consoleMessage = consoleMessage.bind(this)
 
     /* Validation */
-    if (this.config.otherLanguages.length <= 0) {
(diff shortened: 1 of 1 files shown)


mweibel:
🎉 thanks for the fast merge!

PR #749, opened 2020-07-07
feat: add assetPrefix option

### LINKED ISSUE #748: Add optional assetPrefix when used via CDN
In my setup we use a CDN and NextJS's option `assetPrefix`. We have several different nextjs apps all running on the same origin and we add an assetPrefix so our CDN knows which origin to route the requests back to. We strip the assetPrefix when we statically serve the resource. 

We'd like to do the same with next-i18next and provide an asset prefix that affects the clientSide loadPaths only. 

Add a new option called `assetPrefix` that gets prepended to the client here: https://github.com/isaachinman/next-i18next/blob/master/src/config/create-config.ts#L88, but not the server fs backend as that a disk path.

I tried changing `localePath` but that affected the server fs backend as well. 

## Additional context
I'd be happy to open a PR if you feel this is the right solution. Thanks, and I really appreciate this package!

nsmith7989:
Add option for assetPrefix. See #748

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,6 +244,7 @@ MyPage.getInitialProps = async({ req }) => {
 | `use` (for plugins) | `[]`  |
 | `customDetectors` | `[]`  |
 | `shallowRender` | `false`  |
+| `assetPrefix` | `''`  |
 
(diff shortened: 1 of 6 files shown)


kriss145:
Hi @nsmith7989 It is merged to production version? We need it option to solve our problem.

PR #771, opened 2020-07-28
chore: Upgrade i18next dependencies

### LINKED ISSUE #692: Firefox warning next-i18next cookie not secure
I'm getting this warning in Firefox 76 with next-i18next 4.3.0:

`
Cookie “next-i18next” will be soon rejected because it has the “sameSite” attribute set to “none” or an invalid value, without the “secure” attribute. To know more about the “sameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Cookies
`

### LINKED ISSUE #757: Cookie options get overwritten
## Describe the bug
Hi, I set some cookie options inside the `detection` object. However, when loading the page, even though the options are set to the cookie (e.g. `cookieSecure true, cookieSameSite strict`, I could see that in my browser console flickering up for a second), the cookie instantly gets overwritten with another cookie that doesn't have this options set (`cookieSecure false, cookieSameSite none`) 

5.0.0-beta.4

## Steps to reproduce
Take the next-i18next vercel project, change the detection options and check the browser's cookie tab.

## Expected behaviour
The cookie shouldn't get overwritten with the "standard" cookie any time the page loads or the language changes

 - Device: [e.g. MBP 2018 13"]

isaachinman:
Fixes #692. Fixes #757. 

Breaking change.


martpie:
Looking good 👌

PR #799, opened 2020-08-17
fix: I18nContext exported as type

### LINKED ISSUE #789: typescript I18nContext
## Describe the bug
typescript  I18nContext type is not working correctly I got this error

import { I18nContext } from "next-i18next";

export const useActiveLang = () => {
  const {
    i18n: { language: activeLang },
  } = useContext(I18nContext);
  return { activeLang };
};

image

"next-i18next": "^6.0.1",

https://github.com/isaachinman/next-i18next#accessing-the-current-language 

Here is provided code from the documentation I guess it has to be working without any issue but same error

image

image

seanplwong:
I18nContext is exported as type instead of
value, typescript would complain when it is
used as value.

This commit:
- manually creates the type for context and export it
  as upstream `react-i18next` also has incorrect types
- export `withTranslation` as it is exported in index
- add a demo page for I18nContext in example

Fix #789

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
     "i18next-http-middleware": "^3.0.2",
     "path-match": "^1.2.4",
     "prop-types": "^15.6.2",
-    "react-i18next": "^11.7.0",
+    "react-i18next": "^11.7.3",
     "url": "^0.11.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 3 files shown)


isaachinman:
If you believe the upstream `react-i18next` type is incorrect, let's just fix this there. cc @jamuhl.

seanplwong:
I could create another pull request for them, should be easy

didn't have time yesterday night for this, having trouble getting things run in windows machine 😂 , so was only thinking about getting a quick fix.

isaachinman:
Typically the i18next team is very good about fixing things upstream, so let's wait a day or two and see if we can just take the simple route here.

jamuhl:
@seanplwong @isaachinman just create a PR for fixing types

seanplwong:
@isaachinman can u check again?

isaachinman (approved):

PR #835, opened 2020-09-20
feat: nextjs bump version

Darkweak:


isaachinman:
Hi there, `next-i18next@beta` will be released very soon, which will only support Next v10 and higher, making this PR obsolete. Thanks @Darkweak!

Issue #761, opened 2020-07-20
Add option to specify clientLocalePath in config to support approach "One Domain — Multiple Next.js Apps"


DavydenkovM:
## Is your feature request related to a problem? Please describe.
It looks like currently there's no way to provide i18n support for "Multiple Next.js Apps" hosted on one server 
(behind a single reverse proxy).

I have the following next.js apps:

```
next_js_1
  static / locales / en / common.json ...

next_js_2
  static / locales / en / common.json ...
```

Everything is fine on server-side, but I can't find a way to specify locales prefix, so that my Nginx server can return correct locales for each app (for request `/static/locales/en/common.json`). Is there any way to change client request path from `/static/locales/en/common.json` to `APP_PREFIX/static/locales/en/common.json`?

I've tried to change `localePath` or `localeStructure` config options, but then SSR is broken because there's no such directories in FS.

https://github.com/isaachinman/next-i18next/blob/master/src/config/create-config.ts#L73
https://github.com/isaachinman/next-i18next/blob/master/src/config/create-config.ts#L101

## Describe the solution you'd like
Could you please add additional props `clientLocalePath` && `serverLocalePath` in `create-config.ts` to provide more flexibility for `next-i18next` config

Thanks in advance for your attention!

isaachinman:
```jsx
const localePath = typeof window === 'undefined' ? serverPath : clientPath
```

DavydenkovM:
Thanks!

PR #845, opened 2020-10-06
feat: allow to use different locale paths for server and client

SvetoslavKuzmanov:
This is a small PR which allows user to specify different path for client and server scenarios. This allows for more flexible folder structure of the nextjs app, as sometimes just removing the `/public/` portion of the path isn't enough. For example having the nextjs app under `src` folder, as result of using custom server, this way the clientLocalePath would be `./public/static/locales` and the serverLocalePath `./src/public/static/locales`.

### DIFF
--- a/src/config/create-config.ts
+++ b/src/config/create-config.ts
@@ -34,11 +34,19 @@ export const createConfig = (userConfig) => {
     localeStructure,
   } = combinedConfig
 
+  let {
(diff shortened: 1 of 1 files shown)


isaachinman:
https://github.com/isaachinman/next-i18next/issues/761#issuecomment-661305289

Issue #510, opened 2019-09-17
Is cimode supported?


keethu24:
Does the library support `cimode`? If yes, how to configure it with `next-i18next` properly?

isaachinman:
The `config` object is passed directly into the i18next constructor. So yes, anything i18next supports is supported here.

andfk:
I've tried with version `6.0.1` and using `cimode` tries to get the translations files so it throws an error. Something somewhere is not handling such option correctly

isaachinman:
@andfk Feel free to debug and let me know what you think the problem is.

PR #851, opened 2020-10-09
Handles "cimode" correctly

andfk:
As reported before #510 "cimode" is not getting handled correctly, so i've made some small changes for that to work. 

What does "cimode"? It just returns the key requested to the "t" function as explained [here](https://www.i18next.com/overview/api#changelanguage)

Thanks for this library :)


isaachinman:
Looks like you'll need to bump the bundlesize limit.

andfk:
Yeah @isaachinman i see. Now is changed.

src/utils/is-cimode.ts
```diff
@@ -0,0 +1,3 @@
+// Detects if in CI mode due language used
+// https://www.i18next.com/overview/api#changelanguage
+export const isCIMode = (language: string) => language === 'cimode'
```
isaachinman, line 3: Needs a newline at end of file.

src/utils/is-cimode.ts
```diff
@@ -0,0 +1,3 @@
+// Detects if in CI mode due language used
```
isaachinman, line 1: Can we get some basic test coverage for this util?

src/config/create-config.ts
```diff
@@ -34,6 +34,11 @@ export const createConfig = (userConfig) => {
     localeStructure,
   } = combinedConfig
 
+
+  if (isCIMode(defaultLanguage)) {
+    return combinedConfig
```
isaachinman, line 39: Can you explain why this bailout is necessary?

isaachinman (changes requested):

src/config/create-config.ts
andfk: Yes, this is main the reason and why this PR exists. If you set "cimode" without this condition, the server side, will try to resolve "cimode" namespace files, which is wrong

src/utils/is-cimode.ts
andfk: Super simple test added anyway i wonder i should i add more test cases like in `create-config` or asset if the client really returns the keys when in cimode

src/config/create-config.ts
isaachinman: Well, if your aim is to remove the filesystem checks, you should do that directly, instead of bailing out of the entire procedure.

andfk: I already tried in a past commit to by pass the server & client setup, anyway im not a fan of that nested flow condition (even codeclimate fails). Whats your main concern? "cimode" has an specific target where we just want to get the translation key, nothing more so i think no "extra config" is required. Im wrong? Please let me know.

isaachinman (approved):

src/config/create-config.ts
isaachinman: Sure, we can proceed with this. What you've said makes sense.

andfk:
thanks @isaachinman

PR #861, opened 2020-10-21
fix: handling of backend options if using with other backends

### LINKED ISSUE #854: Default namespace not found at "/public/static/locales/en/common.json" when using Locize backend
## Describe the bug
You must provide a clear and concise description of what the bug is.

When backend (Locize) is provided, it is expected to find the namespace from the provided backend. However, its trying to search inside the /public/static folder.

You must provide the npm module version, git tag, or even better the explict commit causing the issue.

"next-i18next": 6.0.3

## Steps to reproduce
You must provide a reproducible repository, or copy/paste blocks of code which are sufficient to reproduce the bug.

1. npx create-next-app
2. Add i18n.js with LocizeBackend
```
const LocizeBackend = require('i18next-locize-backend');

  use: [
    LocizeBackend
  ],
  backend: {
    projectId: PROJECT_ID
    apiKey: API_KEY
    version: process.env.LOCIZE_VERSION || 'latest',
  },
})
```
3. Update _app.js
```
import '../styles/globals.css'

import { appWithTranslation} from '../i18n'

function MyApp({ Component, pageProps }) {
  return <Component {...pageProps} />
}

export default appWithTranslation(MyApp)
```

## Expected behaviour
You must provide a clear and concise description of what you expected to happen.

It should import the namespace and translations from the backend

## Screenshots
If applicable, add screenshots or a GIF to help explain your problem.
image

 - Device: [e.g. MBP 2018 13"]

## Additional context

Tried to follow https://stackoverflow.com/questions/55994799/how-to-integrate-next-i18next-nextjs-locize, @adrai's https://github.com/isaachinman/next-i18next/pull/705, but not working.

adrai:
fixes #854

### DIFF
--- a/src/config/create-config.ts
+++ b/src/config/create-config.ts
@@ -38,13 +38,6 @@ export const createConfig = (userConfig) => {
     localeStructure,
   } = combinedConfig
 
-  if (isServer()) {
-    /*
-      On Server side preload (languages)
-    */
-    combinedConfig.preload = allLanguages
-  }
-
   /** 
    * Skips translation file resolution while in cimode
    * https://github.com/isaachinman/next-i18next/pull/851#discussion_r503113620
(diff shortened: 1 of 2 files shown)


src/config/create-config.ts
```diff
@@ -44,7 +42,8 @@ export const createConfig = (userConfig) => {
     return combinedConfig
   }
 
-  if (isServer()) {
+  const hasCustomBackend = combinedConfig.use && combinedConfig.use.find((b) => b.type === 'backend')
```
isaachinman, line 45: Why check `combinedConfig` instead of `userConfig`?

adrai: 🤷‍♂️ does it make any difference?
Do you want me to change it?

isaachinman: If the `defaultConfig` ever contained `type === 'backend'`, this condition would be true 100% of the time. Also keep in mind that we don't deep merge the backend object until [the very end](https://github.com/isaachinman/next-i18next/blob/1d352541c543e0f08fc7bda724bf1f2c24ec0683/src/config/create-config.ts#L4).

isaachinman:
Thanks @adrai. It occurs to me that the `combinedConfig.preload = allLanguages` line might still be important for custom backends. We rely on all locales/translations being available from init to be able to handle SSR. What do you think?

adrai:
Do you want me to move that line after line 27?

adrai:
@isaachinman What's the problem with the node-multi-build ? I can't see the details.

adrai:
> @isaachinman What's the problem with the node-multi-build ? I can't see the details.

found it

isaachinman:
@adrai It's actual test failures. If you run `yarn test` locally, you'll see it.

adrai:
> @adrai It's actual test failures. If you run `yarn test` locally, you'll see it.

had to clean node_modules, idkw...
now it's working... push is coming in next minutes

adrai:
@isaachinman ok for you? btw: have you seen the change regarding supportedLngs?

src/config/create-config.ts
```diff
@@ -34,7 +38,12 @@ export const createConfig = (userConfig) => {
     localeStructure,
   } = combinedConfig
 
-
+  if (isServer()) {
```
isaachinman, line 41: Sorry, last nitpick, can this go inside the existing `isServer` block, and have a nested `!hasCustomBackend`? Seems a bit more readable/maintainable to me.

adrai: The poblem is, the else block should be executed too

adrai: or not?

adrai: maybe not

adrai:
Now, codeclimate has an issue: https://codeclimate.com/github/isaachinman/next-i18next/pull/861 any idea?

src/config/create-config.ts
isaachinman: Sorry, I don't get what you mean about the `else` block? We should just wrap the code that we _don't_ want to run inside of the `isServer` block inside a secondary `if (!hasCustomBackend)` condition.

adrai: already done this

adrai:
@isaachinman did you have a chance to look at it?

isaachinman:
@adrai CodeClimate approved. Hopefully I will have time for a final review by end of week. Thanks so much!

dkleber89:
Would be nice when this bugfix get merged soon ... Thanks a lot guys

isaachinman (approved):

PR #862, opened 2020-10-21
fix: allow fallbackLng to be set to false

sverps:
Setting fallback language to defaultLanguage in production should only occur if the developer didn't specifically set it to `false`, since that would be a valid config for i18next.

This fix will make it possible to avoid loading any fallback translation files, which could speed up page load.

### DIFF
--- a/src/config/create-config.ts
+++ b/src/config/create-config.ts
@@ -118,7 +118,7 @@ export const createConfig = (userConfig) => {
   /*
     Set fallback language to defaultLanguage in production
   */
-  if (!userConfig.fallbackLng) {
+  if (userConfig.fallbackLng === undefined) {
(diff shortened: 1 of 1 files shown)


src/config/create-config.ts
```diff
@@ -118,7 +118,7 @@ export const createConfig = (userConfig) => {
   /*
     Set fallback language to defaultLanguage in production
   */
-  if (!userConfig.fallbackLng) {
+  if (userConfig.fallbackLng === undefined) {
```
isaachinman, line 121: Can we do something like `typeof !== "boolean"`?

isaachinman (changes requested):

src/config/create-config.ts
sverps: I made a commit with the requested change

isaachinman (approved):

PR #892, opened 2020-11-28
feat: Add setting of language via pageProps

bravetheheat:
Added a feature to allow the defaultLanguage for a specific page to be set through a "lang" prop. This can be passed through getInitialProps.

I understand that this is not the direction that next. js is looking to go towards, but this (or a similar mechanism) is necessary for the localization of URLs, which still play a role in SEO.

Let me know if you need any changes to the implementation or the formatting.

### DIFF
--- a/src/hocs/app-with-translation.tsx
+++ b/src/hocs/app-with-translation.tsx
@@ -17,6 +17,7 @@ interface Props {
 interface WrappedComponentProps {
   pageProps: {
     namespacesRequired?: string[];
+    lang?: string;
   };
 }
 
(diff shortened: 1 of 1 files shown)


isaachinman:
Hi @bravetheheat. The future of `next-i18next` will involve fully handing over the responsibility of handling routing/language to NextJs. Thus, this kind of feature request belongs on the NextJs repo, not here. Hope that makes sense. Thanks!

bravetheheat:
@isaachinman Thanks for the response! Does that mean that you will be moving away from i18next-browser-languageDetector and i18next-http-middleware?

isaachinman:
@bravetheheat Correct, NextJs as a framework will handle all detection and routing.

PR #908, opened 2021-01-08
chore: Adding correct types

TKone7:
- Remove unused types
- Define new types

### DIFF
--- a/examples/simple/next-i18next.config.js
+++ b/examples/simple/next-i18next.config.js
@@ -1,7 +1,7 @@
+const { i18n } = require('./next.config')
 const path = require('path')
 
 module.exports = {
-  defaultLanguage: 'en',
-  otherLanguages: ['de'],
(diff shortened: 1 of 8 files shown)


isaachinman:
@TKone7 Sorry, I think there was duplication of work here. Can you rebase against latest?

TKone7:
Did this. I am a bit confused though. I thought we have to export lower case `appWithTranslation` and not `AppWithTranslation`. In order to allow importing like `import { appWithTranslation } from 'next-i18next'`

isaachinman:
@TKone7 Ah yes, you're completely correct. We need to change [this line](https://github.com/isaachinman/next-i18next/blob/09af10d94eb3f961cf7c7be2f1e9195c56d892ad/types.d.ts#L73), as `appWithTranslation` is now a named export from the root module, and doesn't come off the `NextI18Next` class.

PR #926, opened 2021-01-29
fix: Remove server side code from client bundle

felixmosh:
This PR allowing tree-shaking to work on client / server bundles using [`process.browser`](https://github.com/vercel/next.js/blob/58ea3bb47966ba7ae0f12a9267b54f79a5b4c9fc/test/unit/next-babel-loader.unit.test.js#L124) which is an env variable that gets replaced by next build.

Before, (client bundle)
image

After, (client bundle)
next-i18next-after-fix

In the `after` image you can see that `createConfig` function is much smaller & optimized.

### DIFF
--- a/src/config/createConfig.ts
+++ b/src/config/createConfig.ts
@@ -1,5 +1,5 @@
 import { defaultConfig } from './defaultConfig'
-import { consoleMessage, isServer } from '../utils'
+import { consoleMessage } from '../utils'
 import { Config } from '../../types'
 
 const deepMergeObjects = ['backend', 'detection']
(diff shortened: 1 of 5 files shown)


isaachinman:
Hey @felixmosh, thanks for this! I am pretty sure the Vercel team recommends using `typeof window !== 'undefined'` for these sorts of things. Do you have any link(s) to documentation or discussions about this?

felixmosh:
You are correct, apparently it considered [`legacy`](https://github.com/vercel/next.js/pull/7651).

Any way, the code most be inlined, in order to enable the optimization, I will change my PR.

felixmosh:
Tried it, it doesn't works, probably because next-babel plugin doesn't run on node_modules :|

isaachinman:
If `process.browser` works, then I am happy to proceed with it. I'm wondering how this relates to webpack v5, though?

felixmosh:
Tested it with `Webpack 5` it works as well :]

isaachinman (approved):

PR #922, opened 2021-01-27
fix: TS issues for v8-beta

franzmoro:
- fixes #907 by exposing `appWithTranslation` type
- changes `appWithTranslation` type signature: param is a React Component, not a React Component instance
~~- adds a Next.js example to typescript~~
- fixes the `run-example:prod` script by adding the `build` step before it

PR #928, opened 2021-01-30
feat: add typescript example

franzmoro:
This is a leftover from the [original PR](https://github.com/isaachinman/next-i18next/pull/922) addressing missing TS types. It adds a typescript example.

Used default `tsconfig.json` created when running `yarn dev` straight after having created the Next.js app with `npx create-next-app`. I could be up to the dev(s) to modify this config by themselves.


isaachinman:
My main concern with adding other examples has always been maintenance. However, I think TS is so mainstream now, that I would actually consider just making the main example TS, as the source code itself isn't too different.

What are your thoughts?

franzmoro:
The code of the 2 examples is virtually the same, yes.
I agree also on TS being mainstream, but still a lot of teams don't use it. I'm not sure if having just the TS example becomes confusing, as JS is supported of course.

I'm happy to keep both, but I'm not too opinionated about it - just trying to help. LMK.

examples/with-typescript/components/Footer.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import { version } from 'next-i18next/package.json'
+import { useTranslation } from 'next-i18next'
+
+export const Footer: React.FC<{}> = () => {
```
isaachinman, line 5: The generic argument is unnecessary – you can just write `React.FC =`.

examples/with-typescript/components/Header.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import Head from 'next/head'
+
+interface Props {
```
isaachinman, line 4: Let's use `type` over `interface ` in all cases.

examples/with-typescript/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "v10",
```
isaachinman, line 2: Name should be something like `next-i18next/examples/with-typescript`.

examples/with-typescript/pages/index.tsx
```diff
@@ -0,0 +1,48 @@
+import Link from 'next/link'
+import { useRouter } from 'next/router'
+
+import { useTranslation } from 'next-i18next'
+import { serverSideTranslations } from 'next-i18next/serverSideTranslations'
+
+import { Header } from '../components/Header'
+import { Footer } from '../components/Footer'
+
+const Homepage = () => {
```
isaachinman, line 10: Needs `React.FC`.

examples/with-typescript/pages/second-page.tsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import Link from 'next/link'
+import { useTranslation } from 'next-i18next'
+import { serverSideTranslations } from 'next-i18next/serverSideTranslations'
+
+import { Header } from '../components/Header'
+import { Footer } from '../components/Footer'
+
+const SecondPage: React.FC<{}> = () => {
```
isaachinman, line 9: Again, unnecessary generic arg can be dropped.

isaachinman (changes requested):

examples/with-typescript/pages/_app.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+import { appWithTranslation } from 'next-i18next'
```
sethreidnz, line 2: Why is this called appWithTranslation? When I try to import this in my TS project like this I get:

```
'"../../node_modules/next-i18next/types"' has no exported member named 'appWithTranslation'. Did you mean 'AppWithTranslation'?
```

sethreidnz: It looks like the original example has a file `i18n.js` that it imports this function from. But in this example that file doesn't exist... and this function is being imported from `next-i18next` how does that work?

sethreidnz: Ahh I get it now, this is part of a new beta version.

examples/with-typescript/components/Header.tsx
```diff
@@ -0,0 +1,34 @@
...
+  title: string;
+}
+
+export const Header: React.FC<Props> = ({ title }) => (
```
danielrakoczyas, line 8: HeaderProps

PR #930, opened 2021-01-31
feat: Disable FS backend when user provides one

filipesmedeiros:
One thing to note is that it seemed necessary to make the changes to the way we handle the passing of the config to the client.
Next cannot serialize stuff like functions, so we cannot return everything and assume it's serializable, especially the `use` array and the `backend` prop (with the props inside).
The way this was done now requires the user to pass the same config to `appWithTranslation` that they pass to `serverSideTranslations`. This is kind of repetitive and limiting (same client side config for the whole app). We could allow the user the to pass a function that decides what config to use based on the current page, but that seems troublesome.

Let me know what you think!

### DIFF
--- a/src/createClient/node.ts
+++ b/src/createClient/node.ts
@@ -9,8 +9,9 @@ export default (config: InternalConfig): CreateClientReturn => {
 
   if (!instance.isInitialized) {
     const hasCustomBackend = config?.use?.some((b) => b.type === 'backend')
-    if (!hasCustomBackend)
+    if (!hasCustomBackend) {
       instance.use(i18nextFSBackend)
+    }
 
     config.use.forEach(x => instance.use(x))
     initPromise = instance.init(config)
--- a/src/createClient/node.ts
+++ b/src/createClient/node.ts
@@ -1,9 +1,9 @@
 import i18n from 'i18next'
 import i18nextFSBackend from 'i18next-fs-backend/cjs'
 
-import { Config, CreateClientReturn, InitPromise } from '../../types'
+import { InternalConfig, CreateClientReturn, InitPromise } from '../../types'
 
-export default (config: Config): CreateClientReturn => {
+export default (config: InternalConfig): CreateClientReturn => {
   const instance = i18n.createInstance(config)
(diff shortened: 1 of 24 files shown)


LeonardDrs:
Tried this modification on my app, it works really well.
Dunno if you've seen my edit to this [comment](https://github.com/isaachinman/next-i18next/issues/931#issuecomment-772686669) but I didn't had to pass the configuration myself to `serverSideTranslations` and the `sprintf-postProcessor` worked just fine in SSG/SSR

filipesmedeiros:
Your comment, doesn't it mean you gotta have `next-i18next.config.js` inside `src`? Or was it outside?

LeonardDrs:
Nop, `next-i18next.config.js` is right beside `package.json` and `next.config.js` on the root of the project

filipesmedeiros:
I will rebase this is the branch with the new changes @isaachinman just merged

isaachinman:
Sounds good, thanks!

src/createClient/node.ts
```diff
@@ -8,7 +8,9 @@ export default (config: InternalConfig): CreateClientReturn => {
   let initPromise: InitPromise
 
   if (!instance.isInitialized) {
-    instance.use(i18nextFSBackend)
+    const hasCustomBackend = config?.use?.some((b) => b.type === 'backend')
+    if (!hasCustomBackend)
```
isaachinman, line 12: Let's use curly braces here, please.

types.d.ts
```diff
@@ -23,6 +23,7 @@ export type UserConfig = {
   use?: any[]
 } & InitOptions
 
+
```
isaachinman, line 26: Shouldn't be necessary.

isaachinman (changes requested):

src/createClient/node.ts
filipesmedeiros: Sure thing!

types.d.ts
filipesmedeiros: My bad, removed

isaachinman (approved):

PR #937, opened 2021-02-09
feat: Add an ESM entry point to allow better bundle optimizations

felixmosh:
tree-shaking & module concatenation

### DIFF
--- a/babel.config.json
+++ b/babel.config.json
@@ -59,7 +59,7 @@
           {
             "modules": false,
             "targets": {
-              "browsers": ["last 2 versions", "ie > 11"]
+              "browsers": ["last 2 versions", "ie >= 11"]
             }
           }
         ],
--- a/babel.config.json
+++ b/babel.config.json
@@ -51,6 +51,43 @@
         ]
       ]
     },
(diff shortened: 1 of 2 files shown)


isaachinman:
Sorry, how does this differ from the `es` target?

felixmosh:
It only preserve the import / export syntax without rest of the "modern" syntax like it is in es version

isaachinman:
Then it should target IE11, as that is NextJs' target.

isaachinman:
Cool. My main concern here is if the addition of the `module` entry in `package.json` will cause some build systems to start grabbing that entrypoint instead of the `cjs` target.

I imagine we will need to release this on a major?

felixmosh:
This is the main idea, webpack (or other bundlers) should use the `module` entry, because it is more bundle friendly

isaachinman:
Well, I think we'll need to release it on the next major. Can you change the target and merge into the beta branch?

felixmosh:
It is based on the beta version ;)

isaachinman (approved):

PR #944, opened 2021-02-14
docs: missing dep in install section

YuriGor:
Looks like it's not working if `i18next` is not installed, as I found comparing my not working app created using this README and "simple" example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ A live demo is [available here](http://next-i18next.com/). This demo app is the 
 yarn add next-i18next
 ```
 
-You need to also have `react` and `next` installed.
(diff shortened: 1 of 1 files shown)


isaachinman:
Hi @YuriGor – `i18next` is [declared as a dependency of `next-i18next` directly](https://github.com/isaachinman/next-i18next/blob/master/package.json#L105), so therefore this is not needed. Thanks!

YuriGor:
Yes, I saw there is dep in package.json but when I installed "next-i18next" - it didn't work until I manually installed also "i18next" as in example project. So maybe this is an issue, not docs? I'd expect if lib hasd some dep - it should be handled by npm automatically...

Issue #945, opened 2021-02-15
SSR unfriendly approach demonstrated in example


YuriGor:
## Describe the bug
In example app i18next instance imported
https://github.com/isaachinman/next-i18next/blob/8b6d63740c4acac8f6707cec0c8a74d59fca7ae7/examples/simple/pages/index.js#L2
This way `i18n.language` works on client side only, but has `undefined` value on server side.
So if you for example want to render current language in SSR page - there will be an error because of client/server value inconsistency.

## Occurs in next-i18next version
v7.0.1

## Steps to reproduce
In this line:
https://github.com/isaachinman/next-i18next/blob/8b6d63740c4acac8f6707cec0c8a74d59fca7ae7/examples/simple/pages/index.js#L15
try to render `{i18n.language}` and check logs for error

## How to fix
To make it work as expected we should not import `i18next` but take it from component props as suggested here:
https://react.i18next.com/latest/withtranslation-hoc

```js
import React from 'react';
import { withTranslation } from 'react-i18next';

function MyComponent({ t, i18n }) {
  return <p>{t('my translated text')}</p>
}

export default withTranslation()(MyComponent);
```

isaachinman:
Sure, makes sense. Feel free to open a PR with these changes.

isaachinman:
Hi there, I have just released `next-i18next@8.0.0`, which uses the internationalised routing built into NextJs v10+. This new version of `next-i18next` is quite different, thus making this PR no longer relevant. Thanks!

PR #946, opened 2021-02-15
docs: example app fixed for SSR #945

YuriGor:


YuriGor:
- Sorry for my German, google translate used.
- Feel free to adjust app design changes

isaachinman:
Hey, @YuriGor – can we just convert one of the existing components, without adding all this other UI/changes? I'd like to keep the simple example as simple as possible.

I think we could achieve an example by using the [Footer](https://github.com/isaachinman/next-i18next/blob/master/examples/simple/components/Footer.js), no?

YuriGor:
Most minimal change would be just this moving `i18n` from import to props and nothing more.
image
Maybe you would prefer to have only this change? 
This change already will save time for somebody, who will start development using example and will try to render current language, so we don't have to explicitly demonstrate how to render current lang on SSR page.

isaachinman:
Yep, seems fine to me!

YuriGor:
Ok, then it will be easier to close this PR and create new one.

PR #952, opened 2021-02-19
feat: added customClientLocalePath property

hochitom:
I'm currently working on a project where we're using next & next-i18next within a nx monorepo. and in this case we needed a custom path property to load the translation files in the client.

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,6 +202,7 @@ const { i18n: { language } } = useContext(I18nContext)
 | `otherLanguages` (required) | `[]`  |
 | `localeExtension` | `'json'`  |
 | `localePath` (required) | `'/public/static/locales'`  |
+| `customClientLocalePath` | `'/someOtherFolder/locales'`  |
 | `localeStructure` | `'{{lng}}/{{ns}}'`  |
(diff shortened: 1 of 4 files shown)


isaachinman:
Hey @hochitom – not sure I understand the purpose of this PR. You can just modify the `localePath` config option itself to achieve this behaviour, that's what it's there for.

hochitom:
yeah I know, my description was maybe a bit unclear. when working with nx monorepos, the destination of the locales is for example under `apps/somename/public/static/translations/en/common.json` which I need to configure with `localePath`. When doing this the client tries to load the files also from `apps/somename/public/static/translations/en/common.json` which in this case need to be just `/static/translations/en/common.json`. that's why we added this additional parameter to get nx + next + next-i18next working.

isaachinman:
Still not sure I understand. It sounds like you just need to do:

```
localePath = typeof window === 'undefined' ? serverPath : clientPath
```

Is that the behaviour you are hoping for?

hochitom:
you are totally right -> the new parameter is not necessary.

PR #954, opened 2021-02-21
refactor: Use serialize-javascript

### LINKED ISSUE #931: [8.0.0-beta.4] Error when using postProcessors
Hi, first of all, thanks for this plugin, yours and the community's efforts :heart:  

## Describe the bug
I'm unable to use an i18next postprocessor - here [sprintf-postProcessor](https://github.com/i18next/i18next-sprintf-postProcessor)-, similar to [this comment](https://github.com/isaachinman/next-i18next/issues/869#issuecomment-762595861), any `use` config parameter seems to trigger the following error when trying to start the next application (using `next` command):
> SerializableError: Error serializing `._nextI18Next.userConfig.use[0].process` returned from `getStaticProps` in "/".
> Reason: `function` cannot be serialized as JSON. Please only return JSON serializable data types.

next-18next version: 8.0.0-beta.4 - c3dcef1f2c3510fa6eb2fd5c3fad42ab717964f6

## Steps to reproduce
- Checkout c3dcef1f2c3510fa6eb2fd5c3fad42ab717964f6 in this repository
- Apply "Reproducible patch" bellow
- Run `yarn run-example`

<details>
  <summary>Reproducible patch from c3dcef1f2c3510fa6eb2fd5c3fad42ab717964f6</summary>
  
```diff
diff --git a/examples/simple/next-i18next.config.js b/examples/simple/next-i18next.config.js
new file mode 100644
index 0000000..5366457
--- /dev/null
+++ b/examples/simple/next-i18next.config.js
@@ -0,0 +1,6 @@
+const sprintf = require("i18next-sprintf-postprocessor");
+
+module.exports = {
+  postProcess: "sprintf",
+  use: [sprintf],
+};
diff --git a/examples/simple/package.json b/examples/simple/package.json
index be9e52c..83c434f 100644
--- a/examples/simple/package.json
+++ b/examples/simple/package.json
@@ -10,6 +10,7 @@
   },
   "dependencies": {
     "next": "^10.0.0",
+    "i18next-sprintf-postprocessor": "latest",
     "next-i18next": "link:../../"
   },
   "devDependencies": {
```
  
</details>

## Expected behaviour
There should be no errors and I should be able to translate using any backend/plugin

isaachinman:
Uses [serialize-javascript](https://www.npmjs.com/package/serialize-javascript) to stringify a user's i18next config, which may contain primitives (like functions) that NextJs will refuse to serialise out of the box.

The config gets serialised in `serverSideTranslations`, and then reparsed in `appWithTranslation` via `Function()`.

This allows us to maintain a `next-i18next.config.js` file, without requiring the user to import it themselves, or pass it as arguments into any of `next-i18next`'s functions.

Fixes #931.

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,7 +97,8 @@
     "i18next": "^19.8.4",
     "i18next-fs-backend": "^1.0.7",
     "i18next-http-backend": "^1.0.21",
-    "react-i18next": "^11.7.3"
+    "react-i18next": "^11.7.3",
+    "serialize-javascript": "^5.0.1"
   },
   "peerDependencies": {
     "next": ">= 9.5.0",
--- a/src/appWithTranslation.tsx
+++ b/src/appWithTranslation.tsx
@@ -24,11 +24,13 @@ export const appWithTranslation = <P extends Record<string, unknown>>(WrappedCom
         userConfig,
       } = props.pageProps._nextI18Next
 
+      const parsedUserConfig = Function(`'use strict';return(${userConfig})`)()
+
       locale = initialLocale;
   
       ({ i18n } = createClient({
         ...createConfig({
-          ...userConfig,
+          ...parsedUserConfig,
           lng: initialLocale,
         }),
         lng: initialLocale,
--- a/src/serverSideTranslations.ts
+++ b/src/serverSideTranslations.ts
@@ -1,4 +1,5 @@
 import fs from 'fs'
+import serialize from 'serialize-javascript'
 import path from 'path'
 
 import { createConfig } from './config/createConfig'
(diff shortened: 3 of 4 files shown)


filipesmedeiros:
Hey! I don't love "hammering" the Function call, but hey, if it's safe and it works, I'm all for it. My only question is: how do we know for sure that the config is a function and not an object, for example?

isaachinman:
@filipesmedeiros Not sure I catch your meaning – the config _is_ an object. It also happens to contain functions, etc.

filipesmedeiros:
Doesn't `Function` only eval functions?

PR #974, opened 2021-02-24
fix: Add check for type of window

felixmosh:
This is needed when working in env that is outside of Next.js (like storybook)

### DIFF
--- a/src/config/createConfig.test.ts
+++ b/src/config/createConfig.test.ts
@@ -1,3 +1,6 @@
+/**
+ * @jest-environment node
+ */
 import fs from 'fs'
 import path from 'path'
 
--- a/src/config/createConfig.ts
+++ b/src/config/createConfig.ts
@@ -41,7 +41,7 @@ export const createConfig = (userConfig: UserConfig): InternalConfig => {
     combinedConfig.fallbackLng = combinedConfig.defaultLocale
   }
 
-  if (!process.browser) {
+  if (!process.browser && typeof window === 'undefined') {
     combinedConfig.preload = locales
 
     const hasCustomBackend = userConfig?.use?.some((b) => b.type === 'backend')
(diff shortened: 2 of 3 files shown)


isaachinman:
Looks like tests need to be updated. Also, can you confirm that we still retain the bundle stripping optimisation?

felixmosh:
I've checked the bundle shrinking, this is due to the fact that it becomes `if(false && X)` or `if(true && X)`

isaachinman:
To confirm, you mean that adding the `typeof window === 'undefined'` does not affect the performance gains achieved by #926?

felixmosh:
No affect of the tree-shaking :]

isaachinman:
Great! If you don't have time to fix the tests, just let me know and I'll take care of it.

felixmosh:
I've fixed the tests that are related to createConfig,

The `appWithTranslation.test` that fails is part of "weird" test, it mixes server & client in the assertions, I will happy for your help with it :]

isaachinman:
We don't want to use `@jest-environment node` here, because this code executes in _both_ client and server envs.

I can take care of this, but you'll need to [follow these steps](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork).

isaachinman:
Actually – because this change set is so simple, I think it'll be easier for me to just open another PR. Thanks @felixmosh!

felixmosh:
> We don't want to use `@jest-environment node` here, because this code executes in _both_ client and server envs.
> 
> I can take care of this, but you'll need to [follow these steps](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork).

This is true, so, the only option is to split this test into 2 separate files, server & client.

PR #976, opened 2021-02-24
refactor: Add typeof window check to createConfig

isaachinman:


isaachinman:
@felixmosh Mind a quick review? The tests make sound assertions, but could be cleaned up in the future in terms of client vs server. However, I think this is fine for now.

felixmosh (approved):
Looks OK, thanx!

PR #986, opened 2021-02-24
refactor: Add globalI18n

### LINKED ISSUE #983: A way to get the t-function outside of components
Maybe I'm missing something, but it does not currently seem possible to translate things outside of components/hooks without passing down the `t` function as an argument manually which is quite a hassle. For instance, we have various functions that open toasts on errors and those have to be translated.  

In previous versions of the library we could get the global instance via the module that initialised next-i18next.

Perhaps something like this:

https://github.com/vinissimus/next-translate#gett

isaachinman:
Fixes #983.

### DIFF
--- a/src/appWithTranslation.tsx
+++ b/src/appWithTranslation.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useMemo } from 'react'
 import hoistNonReactStatics from 'hoist-non-react-statics'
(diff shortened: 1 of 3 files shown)


tonytamps:
Absolutely fantastic work @isaachinman - I've been very casually waiting for this feature and I'm so grateful that you're such an active maintainer. It doesn't go unnoticed and it's very appreciated.

PR #1000, opened 2021-02-26
fix: prevent loading translations on each request

adrai:
With the current implementation i18next is initialized on each request on server side.
This triggers a backend load each time.
When using "external" backend plugins via `.use()` this will generate unnecessary load and traffic, i.e. when using i18next-http-backend or i18next-locize-backend.

This PR tries to imitate the "old" behaviour done by i18next-http-middleware, like here: https://github.com/i18next/i18next-http-middleware/blob/master/lib/index.js#L39
where the i18next instance was cloned on each request, this way no unnecessary backend fetch was done.

### DIFF
--- a/src/createClient/node.ts
+++ b/src/createClient/node.ts
@@ -6,7 +6,9 @@ import { InternalConfig, CreateClientReturn, InitPromise, I18n } from '../types'
 let instance: I18n
 
 export default (config: InternalConfig): CreateClientReturn => {
-  if (!instance) instance = i18n.createInstance(config)
+  if (!instance) {
+    instance = i18n.createInstance(config)
+  }
   let initPromise: InitPromise
 
   if (!instance.isInitialized) {
--- a/src/createClient/node.ts
+++ b/src/createClient/node.ts
@@ -1,29 +1,14 @@
 import i18n from 'i18next'
 import i18nextFSBackend from 'i18next-fs-backend'
 
-import { InternalConfig, CreateClientReturn, InitPromise } from '../types'
+import { InternalConfig, CreateClientReturn, InitPromise, I18n } from '../types'
 
-let instance
+let instance: I18n
 
 export default (config: InternalConfig): CreateClientReturn => {
+  if (!instance) instance = i18n.createInstance(config)
   let initPromise: InitPromise
-  if (instance) {
-    // using cloneInstance for subsequent calls,
-    // prevents to load the translations again via backend plugin,
-    // like in i18next-http-middleware
-    let i18nextClone
-    initPromise = new Promise((resolve, reject) => {
-      i18nextClone = instance.cloneInstance({
-        ...config,
-        initImmediate: false,
-      }, (err, t) => err ? reject(err) : resolve(t))
-    })
-    return {
-      i18n: i18nextClone,
-      initPromise,
-    }
-  }
-  instance = i18n.createInstance(config)
+
   if (!instance.isInitialized) {
     const hasCustomBackend = config?.use?.some((b) => b.type === 'backend')
     if (!hasCustomBackend) {
--- a/src/createClient/node.ts
+++ b/src/createClient/node.ts
@@ -3,10 +3,27 @@ import i18nextFSBackend from 'i18next-fs-backend/cjs'
 
 import { InternalConfig, CreateClientReturn, InitPromise } from '../../types'
 
-export default (config: InternalConfig): CreateClientReturn => {
-  const instance = i18n.createInstance(config)
-  let initPromise: InitPromise
+let instance
 
+export default (config: InternalConfig): CreateClientReturn => {
+  let initPromise: InitPromise
+  if (instance) {
+    // using cloneInstance for subsequent calls,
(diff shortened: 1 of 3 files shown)


adrai:
btw: to see visibly what I mean by "backend load", set debug to true and look for: `i18next::backendConnector: loaded namespace...`
image

with and without the PR changes

isaachinman:
Hey @adrai – while creating a global singleton might reduce API calls in some cases, please be aware that the NextJs build process (SSG), SSR, and ISR, all entail building pages on separate threads and lambdas, so there is no guarantee that such a global singleton would even be reused.

A better approach would probably be a local filesystem cache. Are you able to cache Locize assets in the fs with some sort of cache TTL?

adrai:
Yes, I'm aware this optimization is only for serverFULL environments.
For serverless environments it's anyway suggested to bundle the translations with the app (possibility 2: https://github.com/locize/next-i18next-locize#possibility-2-bundle-translations-with-app)

isaachinman:
@adrai My point is that even if you ran a NextJs app on a central server, this optimisation won't affect your build process or ISR, and you'll still end up making tens or hundreds of API calls per build/ISR. I think the only render process this optimisation would help would be actual runtime SSR (which should be the _least_ used option), but even then I'm not completely sure how the NextJs server manages threads.

> Because of this we suggest to download the translations in your CI/CD pipeline (via cli or via api) and package them with your serverless function.

My point was, what do you think about writing a filesystem cache plugin to take care of this for users? Or modifying `i18next-locize-backend` to simply have a `fsCache` option?

I have written similar such fs caches:

https://gist.github.com/isaachinman/ba6dfc5e4dcd0d749bedbb1b68b33927

adrai:
ok, will need to think how to further optimize this...
This PR just reflects the same behaviour like i18next-http-middleware

isaachinman:
In my opinion, any of the remote-service i18next backends should support a cache layer. You could try to detect if a fs is available, and primarily do fs caching, and then fallback to in-memory caching if there is no fs available.

If you'd like, I might be able to write an agnostic memoisation package that caches in fs if available, else mem.

adrai:
This is not so easy... because these plugins are isomorphic and works in browser, Node.js and Deno environment ;-)
but maybe there's a possibility with the i18next-chained-backend... will check...

isaachinman:
Node vs browser would be easy to detect, but I don't have any experience developing for Deno, or what it involves to request filesystem access with it. Let me know if I can help!

adrai:
hmmmm as soon as something required in next-i18next.config.js requires('fs') I get "Module not found: Can't resolve 'fs'"

adrai:
What do you think about something like this?
a cache layer in next-i18next:
image

adrai:
Any idea how to proceed with this?

src/createClient/node.ts
```diff
@@ -3,10 +3,27 @@ import i18nextFSBackend from 'i18next-fs-backend/cjs'
...
 import { InternalConfig, CreateClientReturn, InitPromise } from '../../types'
 
+let instance
+
 export default (config: InternalConfig): CreateClientReturn => {
-  const instance = i18n.createInstance(config)
   let initPromise: InitPromise
-
+  if (instance) {
+    // using cloneInstance for subsequent calls,
```
isaachinman, line 11: Why do we want to `cloneInstance` instead of just returning the same instance?

adrai: Originally a new instance was created, I don't know the reason.
That's why I used cloneInstance.

isaachinman: I'd say, if we are trying to prevent duplicate instances for users who run a central server as opposed to serverless, then let's just return the same instance. Shouldn't be any need to clone.

adrai: ok, changed the code accordingly.

src/createClient/node.ts
```diff
@@ -1,10 +1,12 @@
...
 import i18nextFSBackend from 'i18next-fs-backend'
 
-import { InternalConfig, CreateClientReturn, InitPromise } from '../types'
+import { InternalConfig, CreateClientReturn, InitPromise, I18n } from '../types'
+
+let instance: I18n
 
 export default (config: InternalConfig): CreateClientReturn => {
-  const instance = i18n.createInstance(config)
+  if (!instance) instance = i18n.createInstance(config)
```
isaachinman, line 9:
```suggestion
  if (!instance) {
    instance = i18n.createInstance(config)
  }
```

adrai: done

isaachinman (approved):

adrai:
awesome, thank you ❤️

isaachinman:
Thanks for your help!

adrai:
fyi: I've added an additional [fs-caching example](https://github.com/locize/next-i18next-locize#optional-server-side-caching-to-filesystem) in a [separate branch](https://github.com/locize/next-i18next-locize/tree/local-caching)

isaachinman:
Yeah, that looks very sensible. Thanks @adrai!

PR #1008, opened 2021-03-02
refactor: Add @types/i18next-fs-backend as dependency

isaachinman:


LeonardDrs:
I kinda disagree. Types dependencies should be in devDep, peerDependencies and tagged as optional in `peerDependenciesMeta` since pure Javascript project won't care for it.
I'm not 100% certain of this point of view though, what are your thoughts?

Issue #1026, opened 2021-03-05
Type error: Type 'i18n' is not assignable to type 'null'.


cubi76041:
## Describe the bug
**yarn build** is throwing **Type error: Type 'i18n' is not assignable to type 'null'.**

## Occurs in next-i18next version
8.0.6

## Steps to reproduce
Contents of **tsconfig.json**

```json
{
  "compilerOptions": {
    "target": "es5",
    "lib": ["dom", "dom.iterable", "esnext"],
    "allowJs": true,
    "skipLibCheck": true,
    "strict": true,
    "forceConsistentCasingInFileNames": true,
    "noEmit": true,
    "esModuleInterop": true,
    "module": "esnext",
    "moduleResolution": "node",
    "resolveJsonModule": true,
    "isolatedModules": true,
    "jsx": "preserve",
    "baseUrl": ".",
    "noImplicitAny": false,
  },
  "exclude": [
    "node_modules",
    "dist",
    ".next",
    "out",
    "next.config.js",
    ".babelrc",
    "bundles",
    "coverage"
  ],
  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", "**/jest.setup.ts"]
}
```

Contents of **next-i18next.config.js**:
```ts
module.exports = {
  i18n: {
    locales: ['en', 'fr'],
    defaultLocale: 'en',
  },
  debug: true,
  localePath: path.resolve("./public/static/locales"),
};
```
Contents of **next.config.js**:

```js
const { i18n } = require("./next-i18next.config");

const nextConfig = {
  productionBrowserSourceMaps: true,
  analyzeServer: ["server", "both"].includes(process.env.BUNDLE_ANALYZE),
  analyzeBrowser: ["browser", "both"].includes(process.env.BUNDLE_ANALYZE),
  bundleAnalyzerConfig: {
    server: {
      analyzerMode: "static",
      reportFilename: "./bundles/server.html",
    },
    browser: {
      analyzerMode: "static",
      reportFilename: "./bundles/client.html",
    },
  },
  i18n,
  async rewrites() {
    return [
      {
        source: "/",
        destination: "/home",
      },
    ];
  },
};
```

Contents of **pages/_app.tsx**:

```ts
class WebApp extends App<AppWithStore> {
  render() {
    const { Component, pageProps, store } = this.props;
    return (
      <Provider store={store}>
        <Head>
          <title>{pageProps?.title || DEFAULT_TITLE}</title>
          <meta
[...]

PR #1027, opened 2021-03-05
fix: generate types using tsc and stop shipping the /src directory

stevejhiggs:
By using typescripts declaration mode we can generate d.ts files for all of the relevant types and stop shipping the /src directory

fix for https://github.com/isaachinman/next-i18next/issues/1026

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,5 +1,6 @@
 # Untranspiled source and examples
 examples
+src
 
 # Dependencies
 node_modules
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   },
   "main": "dist/commonjs/index.js",
   "module": "dist/esm/index.js",
-  "types": "./types.d.ts",
+  "types": "dist/types/types.d.ts",
   "license": "MIT",
   "engines": {
     "node": ">=10"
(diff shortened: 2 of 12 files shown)


isaachinman:
Hey @stevejhiggs, looks good! Mind fixing up the conflict?

stevejhiggs:
Done! I had a few issues with a duplicate import of i18n that seems to have been introduced after the last PR merge. I've fixed the duplicate by renaming it in the same way as in ./appwithTranslation

isaachinman:
Looks good. Can we delete [this file](https://github.com/isaachinman/next-i18next/blob/master/serverSideTranslations.d.ts) as well?

stevejhiggs:
I had left that alone as I was unsure as to why both the serverSideTranslations.js + d.ts existed in the root directory. The export from /src/index.ts would seem to cover all the import use cases.

Happy to get rid of both if you do not think they are needed

daniel-stockman-lark:
@stevejhiggs @isaachinman `serverSideTranslations.d.ts` is necessary to support `import { serverSideTranslations } from 'next-i18next/serverSideTranslations'`, so it should be updated to point to the new types output location `dist/types/serverSideTranslations`. (the package.json `types` definition doesn't cover this explicit subpath, only `main`)

```patch
-export { serverSideTranslations } from './src/serverSideTranslations'
+export { serverSideTranslations } from './dist/types/serverSideTranslations'
```

stevejhiggs:
@daniel-stockman-lark ah I see. My mistake, I misread the docs and thought that entry point was not used. I've made the change you suggested.

isaachinman:
Yeah that's correct – `serverSideTranslations` needs to be a separate entry point as it consumes NodeJs modules like `fs` and needs to be stripped from client bundles.

isaachinman (approved):

isaachinman:
@stevejhiggs Happy for this to go in?

stevejhiggs:
Yep, all happy on my side

RA80533:
What were the thoughts on placing these files in the root directory as opposed to a build directory, etc.? Furthermore, were these files meant to be tracked in Git? Conventionally, source code is tracked in Git while the build files are not while the opposite is true for npm publication.

isaachinman:
@RA80533 I don't take your point. None of the files in this PR are build files. You can see which files we ship via npm [here](https://www.npmjs.com/package/next-i18next), on the "Explore" tab.

RA80533:
This change concerns moving the folder that contains the types to the dist folder, yes?

RA80533:
serverSideTranslations.js and serverSideTranslations.d.ts are at the root directory and in dist.

isaachinman:
That is because `serverSideTranslations` is imported from the root:

```tsx
import { serverSideTranslations } from 'next-i18next/serverSideTranslations'
```

Open to any suggestions that support all ECMAScript targets.

PR #1057, opened 2021-03-11
refactor: Some tweaks to the example demo

pgrodrigues:


isaachinman:
Hi @pgrodrigues – can you explain what this PR achieves?

pgrodrigues:
> Hi @pgrodrigues – can you explain what this PR achieves?

Some cleanup to the css, there's no need to have same rules defining different values in the same class.

The title is just a minor thing, probably not important for bigger apps, but for smaller websites where title (and eventually meta description) play a role in SEO. An example can be simple but showcase different possibilities.

isaachinman (approved):

PR #1064, opened 2021-03-14
fix: better support for shallow route changes

### LINKED ISSUE #1023: Shallow routing does not work
## Describe the bug
When switching the location with shallow routing option (using the Link or Router) it is not working (language does not switch).

Cloned from master branch (8.0.6).

## Steps to reproduce
Simply add `shallow={true}` to the [Link component in the example app](https://github.com/isaachinman/next-i18next/blob/master/examples/simple/pages/index.js#L20)

## Expected behaviour
Change of language should also work with shallow routing

 - Browser: Chrome 88.0.4324.190

## Additional context
I can explain why this happens. With the current implementation of next-i18next the i18n instance is always completely replaced. The problem is that the `t` function of `useTranslation` is [cached](https://github.com/i18next/react-i18next/blob/master/src/useTranslation.js#L45). So, if the components do not get recreated but just rerendered (as in shallow routing) the `t` is the old cached one even if the `I18nextProvider` has already a new i18n instance.

### LINKED ISSUE #1059: Re-renders of memoized components on shallow route changes
## Describe the bug
As we discussed in #983, components that utilise useTranslation are always re-rendered on shallow route changes, regardless of if they are memoized or not. This is caused by the fact that the i18n instance is re-created every time the App is rendered.

The problem has been exacerbated in the last two weeks: it now re-renders the memoized component **twice** on every shallow route change rather than once (compare 8.0.1 and 8.1.1). I suspect this is because of this MR being merged in react-i18next: https://github.com/i18next/react-i18next/pull/1273 - which is also mentioned in  #1023.
8.1.1

## Steps to reproduce
Example repo: https://github.com/skrivanos/next-i18next-rerender-poc
https://next-i18next-rerender-poc.vercel.app

## Expected behaviour
The memoized component should not re-render.

skrivanos:
* Memoize the i18n client and do not re-create it unless the route or
  locale has changed (fix #1059).

* Let Next's router locale take precedence over initialLocale.

### DIFF
--- a/src/appWithTranslation.tsx
+++ b/src/appWithTranslation.tsx
@@ -46,7 +46,8 @@ export const appWithTranslation = (
         throw new Error('appWithTranslation was called without config.i18n')
       }
 
-      return createClient({
+
+      const instance = createClient({
         ...createConfig({
           ...userConfig,
           lng: locale,
@@ -54,11 +55,11 @@ export const appWithTranslation = (
         lng: locale,
         resources: initialI18nStore,
       }).i18n
-    }, [_nextI18Next, locale])
 
-    useMemo(() => {
-      globalI18n = i18n
-    }, [i18n])
+      globalI18n = instance
+
+      return instance
+    }, [_nextI18Next, locale])
 
     return i18n !== null ? (
       <I18nextProvider
--- a/src/appWithTranslation.tsx
+++ b/src/appWithTranslation.tsx
@@ -22,12 +22,17 @@ export const appWithTranslation = (
   configOverride: UserConfig | null = null,
 ) => {
   const AppWithTranslation = (props: AppProps) => {
-    let i18n: I18NextClient | null = null
-    let locale = null
+    const { _nextI18Next } = props.pageProps
+    const { locale } = props.router
 
-    if (props?.pageProps?._nextI18Next) {
-      let { userConfig } = props.pageProps._nextI18Next
-      const { initialI18nStore, initialLocale } = props.pageProps._nextI18Next
+    // Memoize the instance and only re-initialize when either:
+    // 1. The route changes (non-shallowly)
+    // 2. Router locale changes
+    const i18n: I18NextClient | null = useMemo(() => {
+      if (!locale || !_nextI18Next) return null
+
+      let { userConfig } = _nextI18Next
+      const { initialI18nStore } = _nextI18Next
 
       if (userConfig === null && configOverride === null) {
         throw new Error('appWithTranslation was called without a next-i18next config')
@@ -41,17 +46,15 @@ export const appWithTranslation = (
         throw new Error('appWithTranslation was called without config.i18n')
       }
 
-      locale = initialLocale;
-
-      ({ i18n } = createClient({
+      return createClient({
         ...createConfig({
           ...userConfig,
-          lng: initialLocale,
+          lng: locale,
         }),
-        lng: initialLocale,
+        lng: locale,
         resources: initialI18nStore,
-      }))
-    }
+      }).i18n
+    }, [_nextI18Next, locale])
 
     useMemo(() => {
       globalI18n = i18n
--- a/src/appWithTranslation.client.test.tsx
+++ b/src/appWithTranslation.client.test.tsx
@@ -2,8 +2,10 @@ import React from 'react'
 import fs from 'fs'
 import { screen, render } from '@testing-library/react'
 import { I18nextProvider } from 'react-i18next'
+import createClient from './createClient'
 
 import { appWithTranslation } from './appWithTranslation'
+import { NextRouter } from 'next/router'
 
 jest.mock('fs', () => ({
   existsSync: jest.fn(),
@@ -19,15 +21,16 @@ jest.mock('react-i18next', () => ({
   __esmodule: true,
 }))
 
+jest.mock('./createClient', () => jest.fn())
 
 const DummyApp = appWithTranslation(() => (
   <div>Hello world</div>
 ))
 
-const props = {
+const createProps = (locale = 'en', router: Partial<NextRouter> = {}) => ({
   pageProps: {
     _nextI18Next: {
-      initialLocale: 'en',
+      initialLocale: locale,
       userConfig: {
         i18n: {
           defaultLocale: 'en',
@@ -36,9 +39,15 @@ const props = {
       },
     },
   } as any,
-} as any
+  router: {
+    locale: locale,
+    route: '/',
+    ...router,
+  },
+} as any)
 
-const renderComponent = () =>
+const defaultRenderProps = createProps()
+const renderComponent = (props = defaultRenderProps) =>
   render(
     <DummyApp
       {...props}
@@ -50,6 +59,8 @@ describe('appWithTranslation', () => {
     (fs.existsSync as jest.Mock).mockReturnValue(true);
     (fs.readdirSync as jest.Mock).mockReturnValue([]);
     (I18nextProvider as jest.Mock).mockImplementation(DummyI18nextProvider)
+    const actualCreateClient = jest.requireActual('./createClient');
+    (createClient as jest.Mock).mockImplementation(actualCreateClient)
   })
   afterEach(jest.resetAllMocks)
 
@@ -69,6 +80,7 @@ describe('appWithTranslation', () => {
       },
     } as any)
     const customProps = {
+      ...createProps(),
       pageProps: {
         _nextI18Next: {
           initialLocale: 'en',
@@ -91,6 +103,7 @@ describe('appWithTranslation', () => {
       <div>Hello world</div>
     ))
     const customProps = {
+      ...createProps(),
       pageProps: {
         _nextI18Next: {
           initialLocale: 'en',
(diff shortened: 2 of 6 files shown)


skrivanos:
Reckon examples/simple/pages/index.ts should be changed to use shallow routing for the change locale button?

examples/simple/package.json
```diff
@@ -10,7 +10,7 @@
   },
   "dependencies": {
     "next": "10.0.5",
-    "next-i18next": "link:../../"
+    "next-i18next": "link:../.."
```
isaachinman, line 13: Why was this change necessary?

package.json
```diff
@@ -108,7 +108,7 @@
     "hoist-non-react-statics": "^3.2.0",
     "i18next": "^19.8.4",
     "i18next-fs-backend": "^1.0.7",
-    "react-i18next": "^11.8.8"
+    "react-i18next": "^11.8.10"
```
isaachinman, line 111: Are you intending to pin to `11.8.10`? Is there something new we need between these two versions? Usually I let dependabot handle updates.

src/appWithTranslation.tsx
```diff
@@ -22,12 +23,19 @@ export const appWithTranslation = (
   configOverride: UserConfig | null = null,
 ) => {
   const AppWithTranslation = (props: AppProps) => {
-    let i18n: I18NextClient | null = null
-    let locale = null
+    const router = useRouter()
+    const { _nextI18Next } = props.pageProps
+    const initialLocale = _nextI18Next?.initialLocale || null
+    const locale = router?.locale || initialLocale
```
isaachinman, line 29: In what case would `router?.locale` ever be falsy?

isaachinman (commented):
In general, these type of changes make me a little nervous, as it seems like even though it might fix shallow routing, it may also cause other issues.

Would something [like this](https://github.com/isaachinman/next-i18next/issues/1023#issuecomment-791981835) work as well? I'm not sure I'm keen on memoising the entire client.

Let me know your thoughts!

skrivanos:
> In general, these type of changes make me a little nervous, as it seems like even though it might fix shallow routing, it may also cause other issues.
> 
> Would something [like this](https://github.com/isaachinman/next-i18next/issues/1023#issuecomment-791981835) work as well? I'm not sure I'm keen on memoising the entire client.
> 
> Let me know your thoughts!

I can understand that. But no, that fix alone would not help (and it's also baked into this PR). One of the problems is that creating a new instance causes two re-renders of memoized components on every shallow route change, leading to extremely poor performance.

examples/simple/package.json
skrivanos: Not sure. Yarn did it automatically after a yarn upgrade so figured the syntax might've changed in later versions (I used the latest one). I could revert this if you want.

package.json
skrivanos: Not pin it, but a fix was introduced that was needed: https://github.com/i18next/react-i18next/pull/1273

src/appWithTranslation.tsx
skrivanos: It's null on the server (at least when running the dev server) despite what the types say.

skrivanos: Ok, it turns out that there's a router instance in `AppProps` as well that's set to an instance of `ServerRouter` on the server, so I think that should be used instead of `useRouter`. It also contains the current locale so there will be no reason to get the initialLocale from `_nextI18Next` anymore.

However, the tests aren't too happy about this change so please confirm if this sounds good before I fix them.

examples/simple/package.json
isaachinman: Yeah, let's revert unless there's an intentional reason.

src/appWithTranslation.tsx
isaachinman: Are you quite sure that `useRouter()` doesn't return a locale when server side? That doesn't sound correct to me, but I might be wrong.

Remind me, what is the problem with just using `_nextI18Next?.initialLocale`?

isaachinman:
Sorry for a delayed response. Happy to keep this PR moving and get it merged in.

src/appWithTranslation.tsx
skrivanos: `useRouter` does not seem to return a router instance at all when using it in _app.tsx. Not sure if this is a Next bug or not, however, the router is always available through the props so should be fine to use that.

`_nextI18Next?.initialLocale` will refer to the old locale because getServerSideProps does not re-run on shallow route changes.

skrivanos:
@isaachinman Noticed a slight hiccup when using this branch in our real application. 

1. Load page 1
2. Route to page 2 (non-shallowly)
3. Route back to page 1 (non-shallowly)
4. Switch locale (shallowly)
5. The page will show the wrong language because new i18n instance won't have the resource bundle for the second language yet.

Not quite sure why yet, just flagging so this does not get accidentally merged. I will look for a solution and modify the e2e tests to cover this case.

PS. Do you prefer that I amend my changes or create new commits?

isaachinman:
@skrivanos Do you have an example use case for switching the locale _and_ wanting shallow routing at the same time? That seems like somewhat of an edge case to me.

skrivanos:
> @skrivanos Do you have an example use case for switching the locale _and_ wanting shallow routing at the same time? That seems like somewhat of an edge case to me.

@isaachinman No, I don't and I agree that it's an edge case. Although it might be considered unexpected that it doesn't work, but could perhaps just mention in the README.

src/appWithTranslation.tsx
```diff
@@ -41,21 +46,17 @@ export const appWithTranslation = (
...
         resources: initialI18nStore,
-      }))
+      }).i18n
+    }, [_nextI18Next, locale])
 
-      useMemo(() => {
-        globalI18n = i18n
-      }, [i18n])
-    }
+    globalI18n = i18n
```
isaachinman, line 59: Should this assignment be inside `useMemo`?

isaachinman:
Cool, then I agree it should be handled in a separate PR (if at all).

src/appWithTranslation.tsx
skrivanos: I can change it if you want me to, it won't really make any difference in this case. Generally both useMemo callbacks and the rendering function itself should be side effect free. To me it made more sense to first let useMemo "resolve" the value and then set the global.

Perhaps the "right" way to do it would be to wrap it in a useEffect with `i18n` as a dep, but I didn't see a good reason to do that.

isaachinman: Sure, that makes sense, I was just concerned that unnecessary reassignment might be an issue when an app is handling thousands of concurrent reqs.

isaachinman:
In general, we should be able to cease passing `initialLocale` via props, [here](https://github.com/isaachinman/next-i18next/blob/master/src/serverSideTranslations.ts#L75), right?

skrivanos:
> In general, we should be able to cease passing `initialLocale` via props, [here](https://github.com/isaachinman/next-i18next/blob/master/src/serverSideTranslations.ts#L75), right?

Yes, I believe so. I can update the PR to remove it, just very short on time lately.

isaachinman:
Cool, no worries – I know the feeling.

DanPurdy:
Hey both, just to confirm that initialLocale is safe to remove from the props returned by serverSideTranslations. Ive changed it in my PR too so once we're good to merge this I'll rebase and pull apart mine.

skrivanos:
@isaachinman @DanPurdy finally had time to updated the MR now - I opted to exclude #1023 since this does not solve it completely.

src/appWithTranslation.tsx
```diff
@@ -22,12 +22,17 @@ export const appWithTranslation = (
...
+    const { locale } = props.router
 
-    if (props?.pageProps?._nextI18Next) {
-      let { userConfig } = props.pageProps._nextI18Next
-      const { initialI18nStore, initialLocale } = props.pageProps._nextI18Next
+    // Memoize the instance and only re-initialize when either:
+    // 1. The route changes (non-shallowly)
+    // 2. Router locale changes
+    const i18n: I18NextClient | null = useMemo(() => {
+      if (!locale || !_nextI18Next) return null
```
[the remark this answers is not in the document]
DanPurdy: The TS definition shows locale as optional but this seems only to be the case if you don't setup Internationalised routing at all. IF someone tries to use this plugin without that being setup they're probably going to have bigger things to worry about than this throwing i would guess 😄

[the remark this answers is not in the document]
skrivanos: @isaachinman if you want me to change something here please let me know. I believe what @DanPurdy says is true.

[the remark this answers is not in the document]
isaachinman: I think I would probably prefer to throw an exception if `locale` is missing.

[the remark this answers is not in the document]
skrivanos: @isaachinman I _finally_ got a chance to take a look at it and it shouldn't be possible to even get to the point of throwing the exception:

1. NextJS itself validates that `locales` is set to a non-empty array if `i18n` is included in `next.config.js`.
2. If `i18n` is **not** configured in `next.config.js`, `serverSideTranslations` will throw an error
3. If serverSideTranslations is **not** used on the page, the `!_nextI18Next` check will make the useMemo hook bail.

Therefore I'm not sure if it's actually worth it to throw the exception here? However, if you still prefer it I will fix it.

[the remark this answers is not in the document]
isaachinman: Based on point 2, I think we can leave it as-is.

isaachinman:
@skrivanos Can you rebase against latest? This PR is my top priority and would love to get it in ASAP.

skrivanos:
> @skrivanos Can you rebase against latest? This PR is my top priority and would love to get it in ASAP.

@isaachinman Done!

isaachinman (approved):

isaachinman:
@skrivanos @DanPurdy This PR has caused #1484. Any objections to leaving the memoisation, but removing the `router.locale` check?

skrivanos:
> @skrivanos @DanPurdy This PR has caused #1484. Any objections to leaving the memoisation, but removing the `router.locale` check?

@isaachinman Won't have time to dive into this again soon unfortunately since I'm a bit overcrowded with work at the moment. I can't quite remember why I changed it to use router.locale but there has to be a reason.

PR #1073, opened 2021-03-15
fix: create i18next clone to fix #1063

### LINKED ISSUE #1063: Memoization of server side instance causes pages to be rendered in the wrong language
## Describe the bug
Pages are only rendered in one language on the server when using getServerSideProps OR when using the dev server with getStaticProps.

Caused by:
 https://github.com/isaachinman/next-i18next/blob/c6fc2e3bdbee07b53cb4a0871cf1abcd372c7b26/src/createClient/node.ts#L9-L11

Added in commit:
https://github.com/isaachinman/next-i18next/commit/e7c5fadae448e005e88886f546c59324b055d2f0

8.1.1

## Steps to reproduce
Run example/simple, click switch locale and reload the page.

## Expected behaviour
To use the correct locale.

## Screenshots
Screenshot 2021-03-13 at 19 20 07

## Additional context
This doesn't cause issues in production when using `getStaticProps` because the translations are stored upon build, but it messes up pages using getServerSideProps (they will always render in whatever language the i18n instance was firstly created as).

adrai:
#1063

### DIFF
--- a/src/createClient/node.ts
+++ b/src/createClient/node.ts
@@ -8,6 +8,11 @@ let instance: I18n
 export default (config: InternalConfig): CreateClientReturn => {
   if (!instance) {
     instance = i18n.createInstance(config)
+  } else {
+    instance = instance.cloneInstance({
+      ...config,
+      initImmediate: false,
+    })
   }
   let initPromise: InitPromise
 
--- a/src/createClient/node.test.ts
+++ b/src/createClient/node.test.ts
@@ -7,8 +7,8 @@ const config = {
 } as any
 
 describe('createClientNode', () => {
+  const client = createClientNode(config)
   it('returns a node client', () => {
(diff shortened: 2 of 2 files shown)


src/createClient/node.ts
```diff
@@ -8,6 +8,11 @@ let instance: I18n
 export default (config: InternalConfig): CreateClientReturn => {
   if (!instance) {
     instance = i18n.createInstance(config)
+  } else {
+    instance = instance.cloneInstance({
+      ...config,
+      initImmediate: false,
```
isaachinman, line 14: Why `false`?

adrai: same as i18next-http-middleware: https://github.com/i18next/i18next-http-middleware/blob/master/lib/index.js#L39

Issue #1101, opened 2021-03-28
Memory leak caused by cloning the i18n instance


dcodus:
## Describe the bug
The #1073 pr introduced a significant memory leak when using `getServerSideProps`.

## Occurs in next-i18next version
[v8.1.2](https://github.com/isaachinman/next-i18next/releases/tag/v8.1.2)

## Steps to reproduce
1. using [simple example](https://github.com/isaachinman/next-i18next/tree/master/examples/simple) from repo.
2. use `getServerSideProps` instead of `getStaticProps` in [index.js](https://github.com/isaachinman/next-i18next/blob/master/examples/simple/pages/index.js#L42)
3. run `yarn build` on the example repo
4. start server in [debug mode](https://nextjs.org/docs/advanced-features/debugging#step-1-start-nextjs-in-debug-mode). In your `package.json`
```js
"start": "NODE_OPTIONS='--inspect' next start -p ${PORT:=3000}"
```
5. In chrome open `chrome://inspect` url then click on inspect
<img width="257" alt="Screen Shot 2021-03-28 at 2 28 15 PM" src="https://user-images.githubusercontent.com/12925816/112763310-ece06d80-8fd1-11eb-8497-777e9d8fd884.png">

6. Tab to the `Memory` tab.
7. Load `localhost:3000`. Take a snapshot (base).
8. Use [loadtest](https://www.npmjs.com/package/loadtest) or any other load testing library to hit localhost:3000 (2000 total requests). Send 1000 requests, stop the load, click collect garbage, take snapshot. Inspect the retained data.
9. Send another 1000 requests, stop the load test, click collect garbage, take a snapshot. Inspect the retained data. NOTE: it will take some time for the snapshot because of the tree created by the leak.
10. Observer the memory significantly increased and the GC never releases it.
11. Inspect the retained objects and you will see the i18n references.

## Expected behaviour
After stopping the loadtest and manually triggering GC (or waiting for it to auto trigger) memory should go back to baseline (plus any additional memory used post first request).
This is the behaviour in [v8.1.1](https://github.com/isaachinman/next-i18next/releases/tag/v8.1.1)

## Screenshots
As you can see the exact same test with the only difference being the version does not show a memory leak.
In 8.1.1 we go from 11.3 -> 12.3
In 8.1.2 we go from 11.3 -> 66.3

**8.1.2**
[...]

PR #1108, opened 2021-03-29
get rid of unecessary i18next instances

adrai:
i18next setup now is simpler and fixes https://github.com/isaachinman/next-i18next/issues/1101

### DIFF
--- a/src/createClient/node.test.ts
+++ b/src/createClient/node.test.ts
@@ -7,8 +7,8 @@ const config = {
 } as any
 
 describe('createClientNode', () => {
+  const client = createClientNode(config)
   it('returns a node client', () => {
-    const client = createClientNode(config)
     expect(typeof client.initPromise.then).toEqual('function')
     expect(typeof client.i18n.addResource).toEqual('function')
     expect(typeof (client.i18n as any).translator).toEqual('object')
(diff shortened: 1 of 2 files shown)


kachkaev:
If there are two concurrent SSR renders and only one i18next instance, won’t the languages end up mixed?

dcodus:
@adrai this is looking very promising

No memory leak after 5000 requests
Screen Shot 2021-03-29 at 3 22 04 AM

Server is stable
Screen Shot 2021-03-29 at 3 22 51 AM

SSR is behaving as expected as per your fix in 8.1.1

Screen Shot 2021-03-29 at 3 24 11 AM
Screen Shot 2021-03-29 at 3 23 49 AM

adrai:
> If there are two concurrent SSR renders and only one i18next instance, won’t the languages end up mixed?

That's why I wanted to clone... same bahaviour like i18next-http-middleware 🤷‍♂️
But I don't know how next.js works behind the scene...

adrai:
alternative: #1109

isaachinman:
Yes, if this was used in `getServerSideProps` and encountered two concurrent reqs, we'd have issues. A singleton unfortunately won't solve our problem.

PR #1110, opened 2021-03-29
fix: resolve memory leak, by cloning always from the main instance

### LINKED ISSUE #1101: Memory leak caused by cloning the i18n instance

adrai:
fixes #1101 by creating a clone always from the first instance of i18next.

### DIFF
--- a/src/createClient/node.ts
+++ b/src/createClient/node.ts
@@ -3,13 +3,15 @@ import i18nextFSBackend from 'i18next-fs-backend'
 
 import { InternalConfig, CreateClientReturn, InitPromise, I18n } from '../types'
 
-let instance: I18n
+let globalInstance: I18n
 
 export default (config: InternalConfig): CreateClientReturn => {
-  if (!instance) {
-    instance = i18n.createInstance(config)
+  let instance: I18n
+  if (!globalInstance) {
+    globalInstance = i18n.createInstance(config)
+    instance = globalInstance
   } else {
-    instance = instance.cloneInstance({
+    instance = globalInstance.cloneInstance({
       ...config,
       initImmediate: false,
     })
--- a/src/createClient/node.ts
+++ b/src/createClient/node.ts
@@ -7,7 +7,6 @@ let globalInstance: I18n
 
 export default (config: InternalConfig): CreateClientReturn => {
   let instance: I18n
-  
   if (!globalInstance) {
     globalInstance = i18n.createInstance(config)
     instance = globalInstance
--- a/src/createClient/node.ts
+++ b/src/createClient/node.ts
@@ -7,6 +7,7 @@ let globalInstance: I18n
 
 export default (config: InternalConfig): CreateClientReturn => {
   let instance: I18n
+  
   if (!globalInstance) {
     globalInstance = i18n.createInstance(config)
     instance = globalInstance


dcodus (approved):
Tested:
1. No memory leaks detected ✅ 
2. Stable servers under heavy load ✅ 
3. SSR works as expected for different locales ✅

dcodus:
@isaachinman When do you think we can get this out?

adrai:
@dcodus probably not in the same timezone 😉

src/createClient/node.ts
```diff
@@ -3,30 +3,32 @@ import i18nextFSBackend from 'i18next-fs-backend'
 
 import { InternalConfig, CreateClientReturn, InitPromise, I18n } from '../types'
 
-let instance: I18n
+let mainInstance: I18n
 
 export default (config: InternalConfig): CreateClientReturn => {
-  if (!instance) {
-    instance = i18n.createInstance(config)
+  let currentInstance: I18n
```
isaachinman, line 9: Can we maintain the naming of `instance` here, and call `mainInstance` `globalInstance` instead?

isaachinman:
Just a minor comment. Otherwise looks good to me.

src/createClient/node.ts
adrai: done

examples/simple/yarn.lock
```diff
@@ -2485,15 +2485,8 @@ neo-async@^2.5.0, neo-async@^2.6.1, neo-async@^2.6.2:
...
-  version "8.1.0"
-  dependencies:
-    "@types/hoist-non-react-statics" "^3.3.1"
-    "@types/i18next-fs-backend" "^1.0.0"
-    core-js "^3"
-    hoist-non-react-statics "^3.2.0"
-    i18next "^19.8.4"
-    i18next-fs-backend "^1.0.7"
-    react-i18next "^11.8.8"
+  version "0.0.0"
```
isaachinman, line 2488: Any idea what happened here? This looks unintended to me.

adrai: strange... just reverted the yarn file

isaachinman:
@adrai Can you remove the yarn file change completely?

adrai:
should be gone now

src/createClient/node.ts
```diff
@@ -3,13 +3,15 @@ import i18nextFSBackend from 'i18next-fs-backend'
 
 import { InternalConfig, CreateClientReturn, InitPromise, I18n } from '../types'
 
-let instance: I18n
+let globalInstance: I18n
 
 export default (config: InternalConfig): CreateClientReturn => {
-  if (!instance) {
-    instance = i18n.createInstance(config)
+  let instance: I18n
```
isaachinman, line 9:
```suggestion
  let instance: I18n
  
```

src/createClient/node.ts
```diff
@@ -3,13 +3,16 @@ import i18nextFSBackend from 'i18next-fs-backend'
...
+
```
isaachinman, line 10:
```suggestion

```

PR #1146, opened 2021-04-15
Update README.md

adrianbienias:
path.resolve() return absolute system path to the file
then that data is exposed as a part of `__NEXT_DATA__` to the client
showing the system path to the project could be a security risk


isaachinman:
The `path.resolve` call is required for many serverless platforms. If you don't want to expose internal implementation details to your client, simply do:

```tsx
localePath: typeof window === 'undefined' ? path.resolve() : '/locales'
```

PR #1147, opened 2021-04-15
docs: Update README with interpolation format example

### LINKED ISSUE #1145: Interpolation format is not used during build/server phase
## Describe the bug
We're experiencing a translation issue on our Next.JS boilerplate which uses next-i18next for translations.
The boilerplate: https://github.com/CrystallizeAPI/crystallize-nextjs-boilerplate

The interpolation format function does not get executed on the server, resulting in wrong translation output in the initial server response. The correct formatting is applied _after_ client hydration.

8.1.3

## Steps to reproduce
1) Go to the production website, https://furniture.superfast.shop/
2) Open devtools
3) Turn off Javascript
4) Refresh page to see the result from the build

## Expected behaviour
Prices should be in the format of e.g.: €60

### Before hydration
result-from-server

### After hydration
result-after-hydration

## Minimum reproduction

```
git clone git@github.com:hakonkrogh/i18next-simple.git
```
```
cd i18next-simple
```
```
npm install
```
```
yarn dev
```

Then
1) Open http://localhost:3000
2) Open devtools
3) Turn off Javascript
4) Refresh page to see the result from the build

hakonkrogh:
Fixes #1145

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,45 +179,16 @@ const MyApp = ({ Component, pageProps }) => <Component {...pageProps} />
 export default appWithTranslation(MyApp, nextI18NextConfig)
 ```
 
-#### Using i18next interpolation format
-
-If you want to take advantage of the native [Intl](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl) namespace to format things like [Numbers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) or [Dates](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat), you need to use the [interpolation.format function](https://www.i18next.com/translation-function/formatting) provided by `i18next` and include that in the next-i18next.config.
-
-Additionally you need to set `config.serializeConfig` to `false`. Here's a complete example:
-
-##### next-i18next.config.js
-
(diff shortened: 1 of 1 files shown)


isaachinman:
Hey @hakonkrogh – thanks for opening this PR! This isn't exactly what I had in mind though. We already have a [section on unserialisable configs](https://github.com/isaachinman/next-i18next#unserialisable-configs), and there are many reasons why users may need to follow this pattern, not just `Intl`.

All we need to do is add the `serverSideTranslations` config override snippet to the existing section. Thanks!

hakonkrogh:
> Hey @hakonkrogh – thanks for opening this PR! This isn't exactly what I had in mind though. We already have a [section on unserialisable configs](https://github.com/isaachinman/next-i18next#unserialisable-configs), and there are many reasons why users may need to follow this pattern, not just `Intl`.
> 
> All we need to do is add the `serverSideTranslations` config override snippet to the existing section. Thanks!

I see, I did not get that. I've updated the docs now after I read through the source of `getServerSideTranslations`.

I think though that it's quite an overhead to pass the config to each page like this. Can't this be solved by providing the config at the context in `appWithTranslation`?

isaachinman (approved):

isaachinman:
@hakonkrogh The `serverSideTranslations` function is called from within data fetching methods and has no access to the React component tree. This is just a consequence of how NextJs works.

PR #1155, opened 2021-04-22
feat: adjust translations path so that it can point to a specific path rath…

ryanbas21:
Hi,
  I wanted to start this and get some feedback. We have a scenario where we are using a monorepo tool called `Nx`. This moves the root `next` app into a subdirectory for example `apps/appName`.
  
  In this scenario we technically moved the `next.config.js` into `apps/appName` and this breaks translations because we can't find the `userConfig` without passing file paths to the many calls to `serverSideTranslations` our app makes.

 I added a `translationsRootDir` config option which is removed from the config before passing the config back to `next`. By doing this, one can leave the next config in root so that the config can be found by default, and we can adjust the namespace resolution to follow the `translationsRootDir` when defined.

One drawback I see though is that the entire "mono" would be following this one path, which may not be ideal.

I plan to update docs once this is a bit more ironed out and see if we can come to a solution.

I'm not a `next.js` expert by any means so I may be off in some understanding.

### DIFF
--- a/src/serverSideTranslations.test.ts
+++ b/src/serverSideTranslations.test.ts
@@ -20,7 +20,22 @@ describe('serverSideTranslations', () => {
       .rejects
       .toThrow('Initial locale argument was not passed into serverSideTranslations')
   })
-
+  it('finds namespaces when monoreporoot is defined', async () => {
+    (fs.readdirSync as jest.Mock).mockReturnValue(['one/two/three'])
+    const props = await serverSideTranslations('en-US', undefined,{
+      i18n: {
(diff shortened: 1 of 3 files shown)


isaachinman:
Not sure I'm in favour of this type of approach at all. If a monorepo tool is responsible for moving dirs around, it should also be responsible for pwd when it executes commands.

ryanbas21:
@isaachinman Totally understand this perspective, i'm not hard set on this approach either. We ended up doing something in line with what you said here anyways. If you think that is the more appropriate approach then i'll close this.

PR #1180, opened 2021-05-05
Update README.md to include global singleton usage

birdwell:
Add Global Singleton Example

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,10 +191,23 @@ export const getStaticProps = async ({ locale }) => ({
 })
 ```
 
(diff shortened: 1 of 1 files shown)


isaachinman:
Hi @birdwell – `browserLanguageDetection` is no longer a config option in `next-i18next`.

birdwell:
@isaachinman ahh, thanks for the update. Do we need to include that in our plugins list ourselves now to allow for that to turn on?

isaachinman:
NextJs itself takes care of language detection.

PR #1240, opened 2021-06-04
fix: resource and default namespace typings

JCQuintas:
# Description

Export the `Resources` and `DefaultNamespace` types from `react-i18next` to allow for proper typing of the `useTranslation` and `serverSideTranslations` functions as described on [react-i18next documentation](https://react.i18next.com/latest/typescript) 

I've added documentation, but I'm not sure that is the better place for it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,13 +174,8 @@ declare module 'next-i18next' {
     ns2: typeof ns2
   }
 
-  // The default for `next-i18next` is 'common', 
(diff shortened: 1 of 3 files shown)


JCQuintas:
Maybe we should wait for my other MR to land on `react-i18next` here: https://github.com/i18next/react-i18next/pull/1328 before we merge this to offer a more complete solution?

PR #1248, opened 2021-06-08
fix: Respect user-defined prefix/suffix when looking for default NS

### LINKED ISSUE #1090: LocaleStructure not compatibility single brackets for namespace
## Describe the bug
I faced localeStructure and interpolation problem
my json is for variables with single brackets. 
like this
`"test": "Why {appName}?",`
and i used in code 
`{t("test", { appName: process.env.NEXT_PUBLIC_APP_NAME })}`

in the config it was stated that

```
localeStructure: "{lng} / {ns}",
    interpolation: {
        prefix: "{",
        suffix: "}",
    },
```
now such an entry does not work, I looked in the code and found that in
`node_modules/next-i18next/dist/commonjs/config/createConfig.js`
you are doing a replacement
`var defaultLocaleStructure = localeStructure.replace ('{{lng}}', lng) .replace ('{{ns}}', defaultNS);`
If you change it to single brackets or add an entry, then everything will work.
8.1.2

## Steps to reproduce
like this. Use single brackets in common.json
`"test": "Why {appName}?",`
and i used in code 
`{t("test", { appName: process.env.NEXT_PUBLIC_APP_NAME })}`

## Expected behaviour
Expected result, that i can use single or more brackets for localeStructure and interpolation

MacOS 
Chrome 89.0.4389.90

alexmchardy:
Fixes #1090 by respecting any custom `interpolation.prefix` and `interpolation.suffix` while using `localeStructure` to see if the default namespace file exists.

Can be tested by changing `/examples/simple/next-i18next.config.js` to
```
module.exports = {
  i18n: {
    defaultLocale: 'en',
    locales: ['en', 'de'],
  },
  interpolation: {
    prefix: '%{',
    suffix: '}',
  },
  localeStructure: '%{lng}/%{ns}',
}
```

Before the change, this config would result in this error:
```
Error: Default namespace not found at public/locales/%{lng}/%{ns}.json
```

After the change, the error goes away and interpolations work as expected. (To verify that interpolations work, you need to change `/examples/simple/public/locales/en/common.json` to use the `%{` `}` prefix and suffix.)

### DIFF
--- a/src/config/createConfig.test.ts
+++ b/src/config/createConfig.test.ts
@@ -116,15 +116,31 @@ describe('createConfig', () => {
       })
     })
 
-    describe('when filesystem is missing defaultNS', () => {
-      it('throws an error', () => {
+    describe('defaultNS validation', () => {
+      it('when filesystem is missing defaultNS throws an error', () => {
         (fs.existsSync as jest.Mock).mockReturnValueOnce(false)
 
         const config = createConfig.bind(null, {
           lng: 'en',
-        } as any)
+        } as UserConfig)
 
         expect(config).toThrow('Default namespace not found at public/locales/en/common.json')
+      })
+
+      it('uses user provided prefix/suffix with localeStructure', () => {
+        (fs.existsSync as jest.Mock).mockReturnValueOnce(false)
+
+        const config = createConfig.bind(null, {
+          interpolation: {
+            prefix: '^^',
+            suffix: '$$',
+          },
+          lng: 'en',
+          localeStructure: '^^lng$$/^^ns$$',
+        } as UserConfig)
+
+        expect(config).toThrow('Default namespace not found at public/locales/en/common.json')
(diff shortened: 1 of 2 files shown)


devan-huapaya:
Can this be merged? @isaachinman <3

isaachinman:
Hey @alexmchardy @devan-huapaya – sorry for the delay.

This PR looks great, it just needs test coverage. Thanks!

devan-huapaya:
No worries @isaachinman  I'm just a random who was hoping to use the prefix setting!

I was able to get around the issue by wrapping useTranslation in it's own hook and adding 
`interpolation: { prefix: '[[', suffix: ']]' }` to all hook calls when options were passed in

alexmchardy:
@isaachinman added a test, hopefully looks good.

cc. @devan-huapaya

isaachinman (approved):

PR #1251, opened 2021-06-10
Update Documentation

samsisle:
# Issue

We've been getting cases where users are not able to build their Next.js page when using this package (working locally but not on Vercel).

It turns out people are using `getServerSideProps` for their page, which I believe is not supposed to work since the documentation also states `getInitialProps` does not work (both methods are almost identical).

When accessing the `getServerSideProps` page →

CleanShot 2021-06-10 at 02 50 42@2x

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ export async function getStaticProps({ locale }) {
 
 Note that `serverSideTranslations` must be imported from `next-i18next/serverSideTranslations` – this is a separate module that contains NodeJs-specific code.
 
-Also, note that `serverSideTranslations` is not compatible with `getInitialProps`, as it _only_ can execute in a server environment, whereas `getInitialProps` is called on the client side when navigating between pages.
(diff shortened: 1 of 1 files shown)


isaachinman:
The `getServerSideProps` data fetching method absolutely _is_ supported. What you are encountering is a drawback of Vercel's serverless architecture, which has been raised and answered countless times in issues on this repo. You will need to use `path.resolve` to instruct Vercel to keep your localisation files in the serverless output.

Feel free to search existing issues.

Good luck!

samsisle:
@isaachinman Thanks for the clarification. Would it be possible add documentation for `getServerSideProps` with Vercel?

isaachinman:
Hi @samsisle – this issue is not specific to `getServerSideProps`, it affects all data-fetching methods. Sure, we could add documentation about Vercel-specific requirements in the [notes](https://github.com/isaachinman/next-i18next#notes) section.

PR #1252, opened 2021-06-10
fix: client config ns assignment logic

### LINKED ISSUE #1250: ns option being overridden on the client to [defaultNS]
## Describe the bug

The `ns` option is being [overridden client side to `[defaultNS]`](https://github.com/isaachinman/next-i18next/blob/c83e7460ce7116932175ac0176e86384a17a776a/src/config/createConfig.ts#L143) within `createConfig` which is preventing `i18next-hmr` from referencing the user-provided `ns` option.

Here's the issue thread: https://github.com/felixmosh/i18next-hmr/issues/32#issuecomment-858366571

    next-i18next@8.4.0

## Steps to reproduce

```jsx
// pages/index.jsx

import { serverSideTranslations } from "next-i18next/serverSideTranslations"
import { useTranslation } from "next-i18next"

export const IndexPage = () => {
  const { i18n } = useTranslation()
  const ns = i18n?.options.ns ?? []

  console.log(ns) // always [defaultNS] on the client

  return <h1>Namespaces: <pre>{JSON.stringify(ns)}</pre></h1>
}

export default IndexPage

export const getStaticProps = async ({ locale }) => ({
  props: await serverSideTranslations(locale),
})
```

## Screenshot

Screenshot 2021-06-10 at 10 08 38

In the screenshot above, I am rendering `JSON.stringify(i18n.options.ns)`

Here's the code: https://github.com/wagerfield/i18next-hmr-hook/blob/main/src/components/page.jsx#L44

## Expected behaviour

The `i18n.options.ns` value should be the same as the provided option in `next-i18next.config.js` on the client and server.

- Device: MBP 2019 16"
- Browser: Chrome 91.0.4472.77 (Official Build) (x86_64)

## Additional context

As mentioned above, this bug is preventing `i18next-hmr` from being able to perform HMR on the client for all user provided namespaces. It only works for the `defaultNS` due to the line referenced above. Commenting out this line results in the `ns` being correct on the client and `i18next-hmr` works as expected:

    // combinedConfig.ns = [defaultNS]

I expect the fix for this is to wrap the assignment line like so:

```js
if (!Array.isArray(combinedConfig.ns)) {
  combinedConfig.ns = [defaultNS]
}
```

wagerfield:
Resolves #1250 

Since the `ns` option can be either a `string` or `string[]` ([docs ref](https://www.i18next.com/overview/configuration-options)) I am just checking for `combinedConfig.ns` to be defined before assigning `[defaultNs]`.

I've added a couple of tests as requested.

### DIFF
--- a/src/config/createConfig.ts
+++ b/src/config/createConfig.ts
@@ -140,7 +140,9 @@ export const createConfig = (userConfig: UserConfig): InternalConfig => {
       loadPath: `${clientLocalePath}/${localeStructure}.${localeExtension}`,
     }
 
-    combinedConfig.ns = [defaultNS]
+    if (!combinedConfig.ns) {
+      combinedConfig.ns = [defaultNS]
+    }
   }
 
   //
--- a/src/config/createConfig.ts
+++ b/src/config/createConfig.ts
@@ -140,7 +140,7 @@ export const createConfig = (userConfig: UserConfig): InternalConfig => {
       loadPath: `${clientLocalePath}/${localeStructure}.${localeExtension}`,
     }
 
-    if (!combinedConfig.ns) {
+    if (typeof combinedConfig.ns !== 'string' && !Array.isArray(combinedConfig.ns)) {
       combinedConfig.ns = [defaultNS]
     }
   }
--- a/src/config/createConfig.test.ts
+++ b/src/config/createConfig.test.ts
@@ -199,7 +199,12 @@ describe('createConfig', () => {
       expect(config.ns).toEqual(['core'])
     })
 
-    it('returns ns when provided', () => {
+    it('returns ns when provided as a string', () => {
+      const config = createConfig({ lng: 'en', ns: 'core' } as UserConfig)
+      expect(config.ns).toEqual('core')
+    })
+
+    it('returns ns when provided as an array', () => {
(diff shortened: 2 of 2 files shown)


src/config/createConfig.ts
```diff
@@ -140,7 +140,9 @@ export const createConfig = (userConfig: UserConfig): InternalConfig => {
       loadPath: `${clientLocalePath}/${localeStructure}.${localeExtension}`,
     }
 
-    combinedConfig.ns = [defaultNS]
+    if (!combinedConfig.ns) {
```
isaachinman, line 143: The `ns` option can either be a string or an array, so let's explicitly check:

```suggestion
    if (!typeof combinedConfig.ns === 'string' || Array.isArray(combinedConfig.ns)) {
```

isaachinman (changes requested):

src/config/createConfig.ts
wagerfield: Done 👍

wagerfield:
@isaachinman I don't really understand the failing cases from Code Climate?

The 2 blocks of code it's complaining about are in no way similar?! 😅

isaachinman (approved):

isaachinman:
Thanks @wagerfield!

wagerfield:
Anytime @isaachinman. When will this update be published to NPM do you think?

PR #1268, opened 2021-06-19
fix: serverSideTranslations loading all locales

### LINKED ISSUE #1267: Loads all languages after migrate to next.js v11 (webpack 5)
It used to work well before updating.
Namespaces works fine, but load all languages:
image

isaachinman:
Fixes #1267.

cc @Nikoms @ktml.

### DIFF
--- a/src/serverSideTranslations.test.ts
+++ b/src/serverSideTranslations.test.ts
@@ -33,20 +33,13 @@ describe('serverSideTranslations', () => {
           locales: ['en-US', 'fr-CA'],
         },
       } as UserConfig)
-      expect(fs.readdirSync).toHaveBeenCalledTimes(2)
+      expect(fs.readdirSync).toHaveBeenCalledTimes(1)
       expect(fs.readdirSync).toHaveBeenCalledWith(expect.stringMatching('/public/locales/en-US'))
-      expect(fs.readdirSync).toHaveBeenCalledWith(expect.stringMatching('/public/locales/fr-CA'))
       expect(props._nextI18Next.initialI18nStore)
         .toEqual({
           'en-US': {
             common: {},
             'namespace-of-en-US': {},
-            'namespace-of-fr-CA': {},
-          },
-          'fr-CA': {
-            common: {},
-            'namespace-of-en-US': {},
-            'namespace-of-fr-CA': {},
           },
         })
     })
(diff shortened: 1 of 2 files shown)


src/serverSideTranslations.test.ts
```diff
@@ -33,20 +33,13 @@ describe('serverSideTranslations', () => {
...
-      expect(fs.readdirSync).toHaveBeenCalledTimes(2)
+      expect(fs.readdirSync).toHaveBeenCalledTimes(1)
       expect(fs.readdirSync).toHaveBeenCalledWith(expect.stringMatching('/public/locales/en-US'))
-      expect(fs.readdirSync).toHaveBeenCalledWith(expect.stringMatching('/public/locales/fr-CA'))
       expect(props._nextI18Next.initialI18nStore)
         .toEqual({
           'en-US': {
             common: {},
             'namespace-of-en-US': {},
-            'namespace-of-fr-CA': {},
```
Nikoms, line 44: We need `namespace-of-fr-CA` in the namespaces of "en-US". Otherwise it's a regression I think (same for some other tests behind).

I'm not sure, but I had issues if `serverSideTranslations` didn't return all namespaces 😕

I think we still need to read all namespaces (`expect(fs.readdirSync).toHaveBeenCalledTimes(2)`), but not storing all locales

Nikoms: Or maybe not. You are right, sorry 😁

Nikoms (approved):

src/serverSideTranslations.test.ts
isaachinman: I think this handles it properly, no?

```tsx
namespacesRequired.forEach((ns) => {
  for (const locale in initialI18nStore) {
    initialI18nStore[locale][ns] = (
      (i18n.services.resourceStore.data[locale] || {})[ns] || {}
    )
  }
})
````

ktml (approved):
Thanks guys! I didn’t have enough about how NextJs works since I had just started using it.

I might see if I can fix `getFixedT` up properly.

src/serverSideTranslations.test.ts
Nikoms: Yes, sorry for the late answer 😊

PR #1302, opened 2021-07-06
fix: only import default config if no userConfig is passed

### LINKED ISSUE #1202: Not working with Next.js >= 10.2.x (probably because of webpack 5)
## Describe the bug
Seems with the update to Next.js 10.2, Next.js is by default using webpack 5: https://nextjs.org/blog/next-10-2#webpack-5
image

next-i18next: v8.5.1
Next.js: >= v10.2.x

## Steps to reproduce
Just update the next version in the simple example: https://github.com/isaachinman/next-i18next/blob/master/examples/simple/package.json#L12

## Error
I have no idea on how this works behind the scene, it may be Next.js itself needs to do something.
Currently an error like this is occurred:
`Error: Cannot find module '/Users/user/Projects/next-i18next/examples/simple/next-i18next.config.js'`
`MODULE_NOT_FOUND`
(happens on dev)

## Additional context
I'm not very familiar with Next.js and/or webpack 5, so if someone wants to invest a bit of time to investigate, this would be really good.

**Currently next-i18next only works with Next.js <= v10.1.x**

There is also an issue in the Next.js repo: https://github.com/vercel/next.js/issues/24729

May be also related: https://github.com/vercel/next.js/issues/24700

adrai:
Fixes #1202

### DIFF
--- a/src/serverSideTranslations.test.ts
+++ b/src/serverSideTranslations.test.ts
@@ -33,6 +33,7 @@ describe('serverSideTranslations', () => {
           locales: ['en-US', 'fr-CA'],
         },
       } as UserConfig)
+      expect(fs.existsSync).toHaveBeenCalledTimes(0)
       expect(fs.readdirSync).toHaveBeenCalledTimes(1)
       expect(fs.readdirSync).toHaveBeenCalledWith(expect.stringMatching('/public/locales/en-US'))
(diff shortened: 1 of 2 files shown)


isaachinman (approved):

isaachinman:
Would be great to add a test if you don't mind?

PR #1305, opened 2021-07-07
fix: set client side backend options only if there is no custom backend

adrai:
Using a different i18next backend having the same backend options like i18next-http-backend is currently causing the custom backend default options to be ignored.

With this fix, the addPath and loadPath options are only configured, if there is no custom backend.


isaachinman (commented):
Possible to add test coverage?

isaachinman (approved):

Issue #462, opened 2019-08-17
Can't find i18n folder when deploying via @zeit/Now


tamer-mohamed:
## Describe the bug
Tried to deploy `examples/simple` to now but failed to run.

## Occurs in next-i18next version
V1.0.0

## Steps to reproduce
- clone examples/simple
- deploy to now
-  502: BAD_GATEWAY

## Expected behaviour
it should run successfully on now platform

## Screenshots

```
ERROR	Uncaught Exception	{"errorType":"Error","errorMessage":"ENOENT: no such file or directory, scandir '/var/task/static/locales/en'","code":"ENOENT","stack":["Error: ENOENT: no such file or directory, scandir '/var/task/static/locales/en'","    at Object.readdirSync (fs.js:790:3)","    at getAllNamespaces (/var/task/page.js:14724:19)","    at _default (/var/task/page.js:14729:27)","    at new NextI18Next (/var/task/page.js:17983:46)","    at Object.k7Sn (/var/task/page.js:14791:18)","    at __webpack_require__ (/var/task/page.js:23:31)","    at Module.Y0NT (/var/task/page.js:9811:63)","    at __webpack_require__ (/var/task/page.js:23:31)","    at Module.f8ve (/var/task/page.js:11600:85)","    at __webpack_require__ (/var/task/page.js:23:31)"],"errno":-2,"syscall":"scandir","path":"/var/task/static/locales/en"}
END RequestId: ba58e4b2-1a15-46d6-9693-00b1206c6973
REPORT RequestId: ba58e4b2-1a15-46d6-9693-00b1206c6973	
```
<img width="981" alt="Screenshot 2019-08-17 at 19 18 04" src="https://user-images.githubusercontent.com/4436327/63215215-c1be4880-c123-11e9-90fc-6c7173b88b7f.png">

<img width="811" alt="Screenshot 2019-08-17 at 19 18 39" src="https://user-images.githubusercontent.com/4436327/63215222-db5f9000-c123-11e9-9fff-87af92e07cdf.png">

isaachinman:
Duplicate of #274.

DutchPrince:
I've got the same problem

isaachinman:
@DutchPrince I do not have knowledge about now the now platform abstracts filesystems. I would welcome any PRs. Alternatively, you can consume your localisation content via network requests with any number of `i18next` plugins.

DutchPrince:
[...]

PR #1347, opened 2021-08-04
fix: improved readme documentation to avoid #462 possibility

LemonFace0309:
Credit goes towards @musps in #462 for the fix. I encountered the same issue after deploying to Vercel and getting a 500 server error loading my ssr pages since my file locale paths couldn't be resolved.

This addition fixes the issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,10 +72,13 @@ This tells `next-i18next` what your `defaultLocale` and other locales are, so th
 #### `next-i18next.config.js`
 
 ```js
+const path = require('path')
+
 module.exports = {
   i18n: {
-    defaultLocale: 'en',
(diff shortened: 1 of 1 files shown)


isaachinman:
Hi @LemonFace0309 – this workaround is only applicable to specific serverless platforms. Actually, I am only aware of it being necessary on Vercel specifically. So at most, it could be added to the docs as a note, not the main use pattern.

obedm503:
It would be nice to have a note about vercel.

isaachinman:
I'd rather have a note in the documentation about _some_ serverless platforms, rather than increase boilerplate for all developers, regardless of their deployment of choice.

Issue #881, opened 2020-11-11
App reload require when adding new translation files


underfisk:
## Describe the bug
I find it strange that i need to restart the app in order for i18next to load the new files or new fields within the existing files, that is quite hard to be doing and I think this is not supposed to happen

## Occurs in next-i18next version
You must provide the npm module version, git tag, or even better the explict commit causing the issue.
Latest version

## Steps to reproduce
You must provide a reproducible repository, or copy/paste blocks of code which are sufficient to reproduce the bug.
Simply use the official documentation example, I even tested with that and happens the same, I need to close the app and open again in order for the files to be resolved

## Expected behaviour
Maybe not having to reload the app and next18n be able to load from fs while in development mode

## Screenshots
If applicable, add screenshots or a GIF to help explain your problem.

## OS (please complete the following information)
 - Device: [e.g. MBP 2018 13"]
 - Browser: [e.g. Chrome 70.0.3538.77]

isaachinman:
There have been several previous issues in regards to this, eg #796. This is easy to achieve with `nodemon`.

kachkaev:
I'd be interesting to consider this as part of #869 TBH. Imagine `<NextI18nextProvider>` polling some dev-only API endpoint and updating the translation data without page reload, when it changes. The API endpoint, in turn, just tracks for the file changes.

underfisk:
@kachkaev Indeed, but behind the scenes, it only needs to listen to the given fs path, and in case the file differs, it invokes the reloading and this could also be provided as a prop to the Provider in order to allow the usage of fs watcher
As you can imagine this can be "annoying" when you are adding translations and instead of waiting for the app hot-reload, you have to reload the whole app (and in large apps, this takes up some minutes to boot)

riywo:
This discussion seems insightful. https://github.com/vercel/next.js/discussions/11419

riywo:
I got a pretty simple solution. I just added a few tricks on my page component only for `next dev`: 1. Add `i18n.reloadResources()` to `getStaticProps`, 2. Add timer to the client to reload the page component periodically.

```ts
import { GetStaticProps } from "next";
import { i18n } from "next-i18next";
import { serverSideTranslations } from "next-i18next/serverSideTranslations";
import { useRouter } from "next/dist/client/router";
import { useEffect } from "react";
import Home from "../components/home";

export const getStaticProps: GetStaticProps = async ({ locale }) => {
  if (process.env.NODE_ENV === "development") {
[...]

PR #1359, opened 2021-08-16
Add `reloadOnPrerender` option to reload resources in serverSideTranslations so that developers don't have to restart their server when making changes to their translation JSON files

dcporter44:
I talked with @isaachinman at the bottom of [this thread (issue #881)](https://github.com/isaachinman/next-i18next/issues/881) regarding a solution to call `i18n.reloadResources()` inside of `serverSideTranslations`. Looks like there are several issues that have been created about this topic.

I've created a new config option called `reloadOnPrerender` that reloads the resources in `severSideTranslations`. I've updated the README as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,6 +213,14 @@ export const getStaticProps = async ({ locale }) => ({
 });
 ```
 
+#### Reloading Resources in Development
+
+Because resources are loaded once when the server is started, any changes made to your translation JSON files in development will not be loaded until the server is restarted.
+
+In production this does not tend to be an issue, but in development you may want to see updates to your translation JSON files without having to restart your development server each time. To do this, set the `reloadOnPrerender` config option to `true`.
+
+This option will reload your translations whenever `serverSideTranslations` is called (in `getStaticProps` or `getServerSideProps`). If you are using `serverSideTranslations` in `getServerSideProps`, it is recommended to disable `reloadOnPrerender` in production environments as to avoid reloading resources on each server call.
+
 #### Options
 
 | Key                 | Default value        |
@@ -221,6 +229,7 @@ export const getStaticProps = async ({ locale }) => ({
 | `localeExtension`   | `'json'`             |
 | `localePath`        | `'./public/locales'` |
 | `localeStructure`   | `'{{lng}}/{{ns}}'`   |
+| `reloadOnPrerender` | `false`              |
 | `serializeConfig`   | `true`               |
 | `strictMode`        | `true`               |
 | `use` (for plugins) | `[]`                 |
--- a/src/serverSideTranslations.ts
+++ b/src/serverSideTranslations.ts
@@ -4,6 +4,8 @@ import path from 'path'
 import { createConfig } from './config/createConfig'
 import createClient from './createClient'
 
+import { globalI18n } from './appWithTranslation'
+
 import { UserConfig, SSRConfig } from './types'
 import { FallbackLng } from 'i18next'
 
@@ -60,7 +62,12 @@ export const serverSideTranslations = async (
     localeExtension,
     localePath,
     fallbackLng,
+    reloadOnPrerender,
   } = config
+
+  if (reloadOnPrerender) {
+    await globalI18n?.reloadResources()
+  }
 
   const { i18n, initPromise } = createClient({
     ...config,
--- a/src/config/defaultConfig.ts
+++ b/src/config/defaultConfig.ts
@@ -27,6 +27,7 @@ export const defaultConfig = {
   react: {
     useSuspense: true,
   },
+  reloadOnPrerender: false,
   serializeConfig: true,
   strictMode: true,
   use: [],
--- a/src/serverSideTranslations.test.tsx
+++ b/src/serverSideTranslations.test.tsx
@@ -197,7 +197,7 @@ describe('serverSideTranslations', () => {
     renderDummyComponent()
 
     if (globalI18n) {
-      globalI18n.reloadResources = jest.fn();
+      globalI18n.reloadResources = jest.fn()
     }
 
     await serverSideTranslations('en-US', [], {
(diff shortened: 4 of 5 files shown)


src/serverSideTranslations.ts
```diff
@@ -60,8 +62,13 @@ export const serverSideTranslations = async (
     localeExtension,
     localePath,
     fallbackLng,
+    reloadOnPrerender,
   } = config
 
+  if (reloadOnPrerender) {
+    await globalI18n?.reloadResources()
+  }
```
isaachinman, line 70: Shouldn't this be called in a `useEffect`, at the very least?

isaachinman (commented):
I'm somewhat surprised that calling `reloadResources` on `globalI18n` works, but if so, this is quite an easy win.

We will need test coverage, though. Thanks!

src/serverSideTranslations.ts
dcporter44: We can't use hooks inside of `serverSideTranslations`, so I'm assuming you mean taking this out of here and putting it at the component level? I think that the above makes more sense given that the goal here is really just to reload the resources on the server while prerendering with `getStaticProps` or `getServerSideProps` specifically in development. What benefits to you see in putting this logic at the component level?

isaachinman: Does this current method work with NextJs' live reload, though?

dcporter44: No, this doesn't work with live reload. Live reload is only triggered by changes to files that are explicitly imported in your app. For live reload to work, you would have to have import statements for each json file in your `public/locales` folder. For example, `import enCommon from 'public/locales/en/common.json'`. Since all JSON files are imported dynamically by next-i18next, Next doesn't know to listen for them.

I think this is the same case when using `create-react-app` too. But this isn't a problem in simple React projects using react-i18next because the developer generally imports all their JSON files in their `i18n.js` file in order to load them into the `resources` property of the i18next config.

So since we're not going to be able to do live reload with these JSON files, it really doesn't make a lot of sense to use `useEffect` here. This PR was really just meant to prevent developers from having to restart their development server to see their changes. Hence the name `reloadOnPrerender`. It only reloads when "pre-rendering" with `getStaticProps` and `getServerSideProps`.

I agree though, it would be ideal if we could use `useEffect` and live reload. I don't see a simple way to do that though. Maybe there's a clever way to do it and I'm not clever enough 🙂

isaachinman: OK, I agree with all of what you've written. I'm not aware of any live reload hooks, but that'd likely need to be something that the NextJs team explicitly expose, for plugins to use.

Let's just get basic test coverage in place, and merge this!

dcporter44: Cool, I'll throw in some basic tests tonight if I get some time - if not, tomorrow night

tigressbailey:
Thanks for this. I was wondering when would it be launched?

src/serverSideTranslations.ts
dcporter44: @isaachinman Sorry been busy with work lately - I'll try to get this in this week. Do you have any recommendations as to what kind of tests could be written for this?

isaachinman: @dcporter44 We just need unit tests that assert that `reloadResources` is called and not called accordingly.

dcporter44:
@isaachinman Let me know if this test implementation seems correct. In order to test if `reloadResources` is called, I need to have access to the `globalI18n` object created by `appWithTranslation`. In order to do that I had to create a dummy component (like you did the `appWithTranslation.test.tsx`) so that `globalI18n` would be instantiated.

README.md
```diff
@@ -213,6 +213,14 @@ export const getStaticProps = async ({ locale }) => ({
 });
 ```
 
+#### Reloading Resources in Development
+
+Because resources are loaded once when the server is started, any changes made to your translation JSON files in development will not be loaded until the server is restarted.
+
+In production this does not tend to be an issue, but in development you may want to see updates to your translation JSON files without having to restart your development server each time. To do this, set the `reloadOnPrerender` config option to `true`.
+
+This option will reload your translations whenever `serverSideTranslations` is called (in `getStaticProps` or `getServerSideProps`). If you are using `serverSideTranslations` in `getServerSideProps`, it is recommended to disable `reloadOnPrerender` in production environments as to avoid reloading resources on each server call.
```
isaachinman, line 222: After reading your lovely docs, it makes me wonder – can we actually check `process.env.NODE_ENV` and disable this feature if it's production?

Or perhaps, alternatively, `reloadOnPrerender` could be an array of `process.env.NODE_ENV` values in which to reload?

isaachinman (commented):
Test coverage looks fine. Just had a passing thought about the API signature to avoid unintended regressions in prod envs.

README.md
minuukang: I think this option should be enable at development mode (like `process.env.NODE_ENV === 'development'`)

isaachinman: If @dcporter agrees with my suggestion, the default value of `reloadOnPrerender` could be `['development']`.

janklimo:
If my understanding is correct, this will not enable HMR of locale files. If that's the case, I think we shouldn't have dismissed #1370 as it provides better DX.

isaachinman:
@janklimo As I've stated numerous times, `next-i18next` will not support HMR, as users can easily set this up themselves via their i18next config.

This PR, however, is a one-liner and is easy to maintain.

janklimo:
@isaachinman I see, thank you for your response. I'm just getting started with the package so I'm not aware of the previous conversations. Would you have any pointers to a resource which explains how it can be easily done using i18next config? Perhaps we could add a readme section explaining it?

README.md
isaachinman: @dcporter44 Sure, let's leave it as-is for now, and we can improve on it in the future if appropriate.

dcporter44: Sounds good. I just updated my branch with master, so let me know if there’s anything else we need before you can merge.

isaachinman (approved):

MartinCura:
Hi! Is this complete functionality? Is it gonna be included in a release soon?

isaachinman:
The `reloadOnPrerender` option has been released in `next-i18next@8.7.0`.

MartinCura:
Thank you for the release! Saving the .json doesn't load the new strings after turning on the option but i suppose it's because modifying components and saving those files doesn't make `getServerSideProps` get called in my setup?

sunknudsen:
Is there a way to live update app when changes are made to translation files in `locales` directory?

CommanderRedYT:
> Is there a way to live update app when changes are made to translation files in `locales` directory?

It only works for me if the translations are JavaScript files that are exporting the translation object and using that in i18n.resources

sunknudsen:
Thanks for helping out @CommanderRedYT.

Would you happen to have some code I could look at?

sunknudsen:
@CommanderRedYT Thanks for helping out!

Adding following to `next-i18next.config.mjs` worked for me.

```mjs
import en from "./locales/en/common.json"
import fr from "./locales/fr/common.json"
```

PR #1384, opened 2021-08-30
add possibility to define next-i18next.config as json file

adrai:
This will help when using next-i18next (simple configurations only) with Next.js >= 10.2.x and prevents the issue #1202


isaachinman:
Can we use a `require` instead, to avoid this issue entirely?

adrai:
> Can we use a `require` instead, to avoid this issue entirely?

unfortunately no... it behaves the same as import

isaachinman:
Sorry, currently out of office – is there an issue on the NextJs repo to track this?

adrai:
> Sorry, currently out of office – is there an issue on the NextJs repo to track this?

no rush...

It's still the same issue as described in #1202 and based on Next.js it's by design, so no issue 🤷‍♂️ https://github.com/vercel/next.js/issues/25077

isaachinman:
If this is an intentional choice by the Vercel team, then I'd prefer to fix this in the "right" way. Can't we hardcore the config path to satisfy that requirement?

adrai:
> If this is an intentional choice by the Vercel team, then I'd prefer to fix this in the "right" way. Can't we hardcore the config path to satisfy that requirement?

If you know how, without any errors....

adrai:
@isaachinman any idea on how to hardcode the config so this could work?

isaachinman:
@adrai Sorry, just now back in office getting caught up. I suppose we'd have to drop the `path.resolve` call, and simply do:

```tsx
userConfig = await import('./next-i18next.config.js')
```

I think this _should_ be fine for most users, but may cause problems for people with atypical setups, including monorepos, etc.

> Can we use a require instead, to avoid this issue entirely?
> unfortunately no... it behaves the same as import

How can that be true? Are you saying that with webpack 5, _all_ NodeJs requires are broken if they use a variable or template literal?

adrai:
```javascript
userConfig = await import('./next-i18next.config.js')
```

Will not work, because relatively that file really does not exist. btw: also the build step is erroring:
```sh
src/serverSideTranslations.ts:51:33 - error TS2307: Cannot find module './next-i18next.config.js' or its corresponding type declarations.

51       userConfig = await import('./next-i18next.config.js')
```

---

Regarding the require... it behaves exactly the same as await import...
So these 2 variants are the same:
```javascript
userConfig = await import(path.resolve(`${DEFAULT_CONFIG_PATH}.js`))
userConfig = require(path.resolve(`${DEFAULT_CONFIG_PATH}.js`))
```

I tried everything... 😢

isaachinman:
I haven't had a chance to deep dive into webpack 5 yet, but I assume there must be a way of designating true NodeJs entry points via `package.json` or similar, so that webpack won't attempt these optimisations and won't hijack `require`.

isaachinman:
Apologies again, my time has been really limited lately.

Calling for help on this one – there must be other packages in the NextJs ecosystem that rely on config files.

@martpie @HaNdTriX @timer @ijjk Any advice?

isaachinman:
@adrai Well, failing any first-class solution, something like this should work:

```tsx
userConfig = eval(fs.readFileSync(path.resolve(DEFAULT_CONFIG_PATH), 'utf-8'))
```

adrai:
> @adrai Well, failing any first-class solution, something like this should work:
> 
> 
> 
> ```tsx
> 
> userConfig = eval(fs.readFileSync(path.resolve(DEFAULT_CONFIG_PATH), 'utf-8'))
> 
> ```

Unfortunately this would not work if the config includes some require statement... 😥
i.e. when using other i18next backends

isaachinman:
And how would that use case be supported by a JSON file?

adrai:
> And how would that use case be supported by a JSON file?

Not at all, that's why I was thinking of making it clear with a JSON file 👍

adrai:
Do you have some direct connection to the Next.js team to ask for their advice?

isaachinman:
@adrai Whatever solution we decide on should work regardless of a user's specific requirements.

> Do you have some direct connection to the Next.js team to ask for their advice?

Yes, I'll ping them via Slack.

isaachinman:
@adrai I was told it sounds related to this issue, which is actively being worked on: https://github.com/vercel/next.js/issues/24700

adrai:
> @adrai I was told it sounds related to this issue, which is actively being worked on: [vercel/next.js#24700](https://github.com/vercel/next.js/issues/24700)

I'm already subscribed to that issue ;-)
I've also tried 11.1.3-canary.33 with experimental nftTracing=true but no luck :-(

adrai:
fyi: tried also with 11.1.3-canary.41 and the new experimental property outputFileTracing 😢

PR #1410, opened 2021-09-07
refactor: Move hook out of condition

Arnarkari93:
- https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

### DIFF
--- a/src/appWithTranslation.tsx
+++ b/src/appWithTranslation.tsx
@@ -51,11 +51,11 @@ export const appWithTranslation = (
         lng: initialLocale,
         resources: initialI18nStore,
       }))
-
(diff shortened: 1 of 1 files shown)


isaachinman:
@Arnarkari93 You marked this as a bug fix via semantic titling. Did you actually encounter a bug?

Arnarkari93:
No, I did not encounter a bug. I have no idea how this semantic title thing works I just marked it as a fix as a guess. I can also mark it as a chore if you prefer that.

isaachinman (approved):

isaachinman:
Thanks @Arnarkari93!

Issue #1443, opened 2021-09-29
`appWithTranslations` HOC is not providing correct `I18nextProvider`


gabsprates:
## Describe the bug
While I navigate, the page don't shows the translations but the keys.

## Occurs in next-i18next version
`8.8.0`, and `next@10.0.1`

## Steps to reproduce
I have three pages and my setup is basically the same from this [example](https://github.com/isaachinman/next-i18next/tree/master/examples/simple):
- `_app.js` with `export default appWithTranslation(MyApp);`;
- each page with its own exported `getServerSideProps` as
    `async ({ locale }) => ({ props: { ...await serverSideTranslations(locale, ['common', 'PAGE_NAME']) } })`

## Expected behaviour
The pages should load the namespaces and show translations.

## OS
- OS: Ubuntu 20.04
- Browser: Chrome 92.0.4515.159 (Official Build) (64-bit)

## Additional context
I added the `I18nextProvider` to my tree and everything works:
```tsx
import { useTranslation } from 'next-i18next';
import { I18nextProvider } from 'react-i18next';

const MyApp = ({ Component, pageProps }) => {
  const { i18n } = useTranslation();
  
  return (
    <I18nextProvider i18n={i18n}>
      <App />
    </I18nextProvider>
  );
};
  
export default appWithTranslation(MyApp);
```

isaachinman:
Please provide a reproducible repository.

gesabehrends:
I have the same issue when updating to next-i18next@8.8.0, while everything works with next-i18next@8.7.0.

isaachinman:
@gesabehrends Please provide a reproducible repository.

angelovdev:
The app needs to be restarted to get the new keys to work. Have you tried that?

belgattitude:
@gabsprates few questions cause I've been affected (and downgraded to 8.7.0). 

- Do you import from react-i18next sometimes ? ie: `import { useTranslation } from 'react-i18next';` vs `import { useTranslation } from 'next-i18next` ?

- Can you try with those locked deps:

```json5
  "dependencies": {
    "i18next": "20.6.1", // Actually if i18next is at v21, I'm able to reproduce some things
    "next-i18next": "8.8.0",
    "react-i18next": "11.12.0",
  }
```

PS: Asking that because I could partially reproduce when i18next version is > 20 here https://github.com/belgattitude/nextjs-monorepo-example/pull/510. I'll probably need more time to have a full repro.

[...]

PR #1467, opened 2021-10-13
fix: move @types/i18next-fs-backend dep to dev-deps

belgattitude:
## Purpose

Avoid having "@types/i18next-fs-backend"  as a dep. 

**tldr;** if latest`@types/i18next-fs-backend"` (> 1.1.0), it will install i18next v21... we'll have both versions hoisted. I suspect packages managers (yarn/npm/pnpm), webpack (4/5), node resolutions (ssr) to give different results leading to bugs like https://github.com/isaachinman/next-i18next/issues/1443

## Possible explanation

Why the `@types/i18next-fs-backend": "^1.0.0"` is problematic here.

- 1.0.0 depends on i18next 19 (that will be installed)
- 1.1.0 (no deps)
- 1.1.2 depends on i18next 21 (that will be installed)

*PS: There's not yet official types for i18next-fs-backend, see https://github.com/i18next/i18next-fs-backend/issues/20*

I'm not sure about the move to devDeps, I could have locked it to `1.0.1` but I haven't seen a re-export of it. So it should be fine. (@isaachinman can you confirm ?, Otherwise we could lock the version or move it to optional peer-deps. If some projects relies on those types, they might have to explicitly add them, not technically a BC but who knows ?)

I've set version to `~1.0.0` in dev-deps... 

When next-18next supports i18next v21, it might need to be updated to `^1.1.0`.

## Possible fixes.

Will probably fix this https://github.com/isaachinman/next-i18next/issues/1443 and hopefully this https://github.com/i18next/react-i18next/issues/1379

I could not give a reproduction example, but here's the explanation here: https://github.com/belgattitude/nextjs-monorepo-example/pull/510 (I could reproduce it partially, but on prod I'm using a different version of yarn so I'm not sure, my workaround was to add i18next: ^20 explicitly).

## Extra

I realized that the example was loading i18next 19 too... Changed the lock: https://github.com/isaachinman/next-i18next/pull/1467/commits/b3c55979a3e27f61f9868c415b6cf3de87db44a1

### DIFF
--- a/examples/simple/yarn.lock
+++ b/examples/simple/yarn.lock
@@ -176,13 +176,6 @@
     "@types/react" "*"
     hoist-non-react-statics "^3.3.0"
 
-"@types/i18next-fs-backend@^1.0.0":
-  version "1.0.0"
-  resolved "https://registry.yarnpkg.com/@types/i18next-fs-backend/-/i18next-fs-backend-1.0.0.tgz#3fb0374f4d376375b7cc1d3729bca663d104fa64"
-  integrity sha512-PotQ0NE17NavxXCsdyq9qIKZQOB7+A5O/2nDdvfbfm6/IgvvC1YUO6hxK3nIHASu+QGjO1QV5J8PJw4OL12LMQ==
-  dependencies:
-    i18next "^19.7.0"
-
 "@types/json-schema@^7.0.5":
   version "7.0.6"
   resolved "https://registry.yarnpkg.com/@types/json-schema/-/json-schema-7.0.6.tgz#f4c7ec43e81b319a9815115031709f26987891f0"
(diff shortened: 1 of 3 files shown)


belgattitude:
I can confirm it would close this one too https://github.com/i18next/react-i18next/issues/1379.

In the meantime

```json5
  "resolutions": {
    "@types/i18next-fs-backend": "1.1.0", // this would be anough
    "i18next": "^20.6.1" // or to be safe
  },
```

isaachinman:
You'll need to `yarn install` again.

isaachinman:
I don't understand what you are suggesting. You just need to move the dependency from deps into dev deps, and do `yarn install && yarn run-example`.

PR #1522, opened 2021-11-15
fix: Pin i18next-fs-backend

### LINKED ISSUE #1518: Error: Cannot find module './path.cjs' while building on next-i18next v9.0.0
Error: Cannot find module './path.cjs' while building

image

next-i18next 9.0.0 & 8.10.0 & 8.9.0
yarn 1.22.10
next 11.1.2
node 14.18.0

## I try to fix it

I've rename "path.cjs" to "path.js" at readFile.js at i18next-fs-backend package directory in node_modules then it worked! 🤔

image

isaachinman:
Fixes #1518

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,5 +114,8 @@
   "peerDependencies": {
     "next": ">= 10.0.0",
     "react": ">= 16.8.0"
+  },
+  "resolutions": {
+    "i18next-fs-backend": ">=1.1.4"
   }
 }


patrick-mcdougle:
Is there a reason we did this with resolutions instead of just bumping it in the "dependencies" key?

isaachinman:
@patrick-mcdougle Yes, to fix #1518. `resolutions` is the easiest way to force a version range, without pinning in `dependencies`.

PR #1530, opened 2021-11-16
refactor: Set config.react.useSuspense to false

### LINKED ISSUE #1255: "ReactDOMServer does not yet support Suspense." - message on projects start on nextjs
## 🐛 Bug Report

On start nextjs shows error with next text: 
"ReactDOMServer does not yet support Suspense."
And after a two seconds it is renders the page with correct translation.

### Screen of error:
image

## To Reproduce

Test translation files serves with [http-server](https://www.npmjs.com/package/http-server)

There is a code example on code sandbox: 
[example](https://codesandbox.io/s/next-server-localization-gh6bj?file=/next-i18next.config.js:458-487)

the structure on a server is next: 

- \ ru 
   | - header.json
- \ ua
   | - header.json
- \ pl
   | - header.json

### translation file content (header.json)
```json
{
  "warenyky": "вареники"
}
```
   
### next-i18next.config.js
```js
const I18nextChainedBackend = require('i18next-chained-backend/dist/cjs/i18nextChainedBackend');
const I18NextHttpBackend = require('i18next-http-backend/cjs');

module.exports = {
  debug: true,
  i18n: {
    defaultLocale: 'ru',
    locales: ['ru', 'ua', 'pl'],
  },
  serializeConfig: false,
  use: [I18nextChainedBackend],
  backend: {
    backends: [
      I18NextHttpBackend,
    ],
    backendOptions: [
      {
        loadPath: 'https://9b67ef78f276.ngrok.io/{{lng}}/{{ns}}.json',
        crossDomain: true,
        requestOptions: {
          mode: 'no-cors',
          cache: 'default',
        },
      },
    ],
  },
};
```

### next.config.js
```js
const { i18n } = require('./next-i18next.config')

module.exports = {
  i18n,
};
```

### _app.tsx
```tsx
import type { AppProps, AppContext } from 'next/app';
import { appWithTranslation } from 'next-i18next'
import { reduxWrapper } from 'store';
import Head from 'next/head';
import nextI18NextConfig from '../next-i18next.config';

function MyApp({ Component, pageProps }: AppProps | any) {
  return (
    <>
      <Head>
        <title>some title</title>
      </Head>
        <Component {...pageProps} />
    </>
  );
}

MyApp.getInitialProps = async ({ Component, ctx }: AppContext) => {
  const pageProps = Component.getInitialProps
    ? await Component.getInitialProps(ctx)
    : {};
  return { pageProps: pageProps };
};

export default reduxWrapper.withRedux(appWithTranslation(MyApp, nextI18NextConfig));
```

### page.tsx
```tsx
import Head from 'next/head';
import React from 'react';
import { serverSideTranslations } from 'next-i18next/serverSideTranslations';
import { GetServerSideProps } from 'next';
import nextI18NextConfig from '../next-i18next.config.js';
import TargetComponent from '@components/TargetComponent';

const Page = (props) => {
  return (
    <>
      <Head>
        <title>page</title>
      </Head>
      <TargetComponent>content</TargetComponent>
    </>
  );
};

export const getServerSideProps: GetServerSideProps = async ({
  locale,
}: any) => {
  return {
    props: {
      ...(await serverSideTranslations(locale, ['header'], nextI18NextConfig)),
    },
  };
};

export default Page;
```

### TargetComponent.tsx
```tsx
import { useTranslation } from 'next-i18next';
import React from 'react';

export const TargetComponent = function (props) {
  const [t, i18n, ready] = useTranslation('example', { 
    useSuspense: false 
  });
  console.log('ready', ready, i18n,  i18n.getResourceBundle(i18n.language, 'header'));
  return (
    <div>
      {ready ? t('warenyky'): 'loading'}
    </div>
  );
};
```

## Expected behavior

The possibility to use backend stored translations with `next-i18next`, `i18next-http-backend` and `i18next-chained-backend` without "Suspense" error.

## Environment

- *runtime version*:  node v14, safari
- *next-i18next version*: 8.4.0
- *i18next-http-backend version*: 1.2.6
- *i18next-chained-backend version*: 3.0.2
- *next version*: 10.2.3
- *os*: Mac

isaachinman:
Fixes #1255

cc @adrai – following up on your comment: https://github.com/isaachinman/next-i18next/issues/1255#issuecomment-864523329

We are already awaiting the initialisation promise inside `serverSideTranslations` [here](https://github.com/isaachinman/next-i18next/blob/e97bfbf4bfb3d3c8660a53018e40cf6e0cfd6865/src/serverSideTranslations.ts#L77).

The reason you have to manually list your namespaces in your demo project [here](https://github.com/locize/next-i18next-locize/blob/main/next-i18next.config.js#L17) is because they are not present in your filesystem, which we normally [query for automatically](https://github.com/isaachinman/next-i18next/blob/e97bfbf4bfb3d3c8660a53018e40cf6e0cfd6865/src/config/createConfig.ts#L120).

Let me know if you think this looks like a safe change 👍

### DIFF
--- a/src/config/createConfig.test.ts
+++ b/src/config/createConfig.test.ts
@@ -45,6 +45,7 @@ describe('createConfig', () => {
         expect(config.preload).toEqual(['en'])
         expect(config.strictMode).toEqual(true)
         expect(config.use).toEqual([])
+        expect(config.react?.useSuspense).toEqual(false)
 
         expect(fs.existsSync).toHaveBeenCalledTimes(1)
         expect(fs.readdirSync).toHaveBeenCalledTimes(1)
@@ -197,6 +198,7 @@ describe('createConfig', () => {
       expect(config.preload).toBeUndefined()
       expect(config.strictMode).toEqual(true)
       expect(config.use).toEqual([])
+      expect(config.react?.useSuspense).toEqual(false)
     })
 
     it('deep merges backend', () => {
--- a/src/config/defaultConfig.ts
+++ b/src/config/defaultConfig.ts
@@ -25,7 +25,7 @@ export const defaultConfig = {
   localePath: LOCALE_PATH,
   localeStructure: LOCALE_STRUCTURE,
   react: {
-    useSuspense: true,
+    useSuspense: false,
   },
   reloadOnPrerender: false,
   serializeConfig: true,


adrai:
This is ok, just the default options anyway....
Maybe there should be a hint somewhere, when using a "remote" backend either the ready flag needs to be checked before calling the t() function or the namespaces needs to be listed in the ns option to be proloaded on init

isaachinman:
Yeah, good point. We could add it to [this section](https://github.com/isaachinman/next-i18next#notes). Although, is that statement exactly correct?

The namespaces will need to be listed either way. If you await the ready flag without passing namespaces, the instance will be "ready" but without any content, right?

adrai:
Yes, that section should be ok.

the i18next instance is always ready after initialization, but if the namespaces where not listed in the ns option, they are not loaded on init, but on request (i.e. useTranslation) and this means they are loaded asynchronous, that's why the ready flag has to be checked...

isaachinman:
But, the ready flag cannot be checked arbitrarily during React render (when the `useTranslation` hook is invoked)... Unless I am misunderstanding?

adrai:
> But, the ready flag cannot be checked arbitrarily during React render (when the `useTranslation` hook is invoked)... Unless I am misunderstanding?

What do you exactly mean by this? Here an example: https://github.com/i18next/react-i18next/blob/2bd69a2cc638acfafb37dc0c56db6add4e9dac85/test/typescript/examples.test.tsx#L59

isaachinman:
Ah, I see what you mean. Thanks for explaining. Would you have a moment to write this up for the `next-i18next` docs?

adrai:
> Ah, I see what you mean. Thanks for explaining. Would you have a moment to write this up for the `next-i18next` docs?

I'm not that good in english, but I will try in the next hours...

PR #1531, opened 2021-11-16
docs: note for asynchronous backends

adrai:
like requested in #1530

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,12 +260,28 @@ To migrate from previous versions to the version 8, check out the [v8-migration 
 
 ## Notes
 
+### Docker
+
 For Docker deployment, note that if you use the `Dockerfile` from [Next.js docs](https://nextjs.org/docs/deployment#docker-image) do not forget to copy `next.config.js` and `next-i18next.config.js` into the Docker image.
 
 ```
 COPY --from=builder /app/next.config.js ./next.config.js
 COPY --from=builder /app/next-i18next.config.js ./next-i18next.config.js
 ```
+
+### Asynchronous i18next backends
+
+If you choose to use an i18next backend different to the in-built [i18next-fs-backend](https://github.com/i18next/i18next-fs-backend), you need to make sure the translation resources are loaded before you call the `t` function.
+Since [Suspense is not yet supported on server side](https://github.com/isaachinman/next-i18next/issues/1255), this can be solved in 2 different ways:
+
+**1) preload the namespaces:**
+
+Set the `ns` option, like in [this example](https://github.com/locize/next-i18next-locize/blob/main/next-i18next.config.js#L17). This way the translation resources will all be loaded on initialization.
+
+**2) check the ready flag:**
+
+You need to handle the not ready state by eg. render a loading component as long `ready === false` or `props.tReady === false`. Not doing so will result in rendering your translations before they loaded, which will cause save missing be called although translations exists (just yet not loaded).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -260,13 +260,29 @@ To migrate from previous versions to the version 8, check out the [v8-migration
...
 For Docker deployment, note that if you use the `Dockerfile` from [Next.js docs](https://nextjs.org/docs/deployment#docker-image) do not forget to copy `next.config.js` and `next-i18next.config.js` into the Docker image.
 
 ```
 COPY --from=builder /app/next.config.js ./next.config.js
 COPY --from=builder /app/next-i18next.config.js ./next-i18next.config.js
 ```
 
+### Asynchronous i18next backends
+
+If you choose to use an i18next backend different to the in-built [i18next-fs-backend](https://github.com/i18next/i18next-fs-backend), you need to make sure the translation resources are loaded before you call the `t` function.
```
isaachinman, line 274:
```suggestion
If you choose to use an i18next backend different to the built-in [i18next-fs-backend](https://github.com/i18next/i18next-fs-backend), you will need to ensure the translation resources are loaded before you call the `t` function.
```

README.md
```diff
@@ -260,13 +260,29 @@ To migrate from previous versions to the version 8, check out the [v8-migration
...
+Since [Suspense is not yet supported on server side](https://github.com/isaachinman/next-i18next/issues/1255), this can be solved in 2 different ways:
```
isaachinman, line 275:
```suggestion
Since [React suspense is not yet supported for SSR](https://github.com/isaachinman/next-i18next/issues/1255), this can be solved in 2 different ways:
```

README.md
```diff
@@ -260,13 +260,29 @@ To migrate from previous versions to the version 8, check out the [v8-migration
...
+
+**1) preload the namespaces:**
```
isaachinman, line 277:
```suggestion
**1) Preload the namespaces:**
```

README.md
```diff
@@ -260,13 +260,29 @@ To migrate from previous versions to the version 8, check out the [v8-migration
...
+
+Set the `ns` option, like in [this example](https://github.com/locize/next-i18next-locize/blob/main/next-i18next.config.js#L17). This way the translation resources will all be loaded on initialization.
```
isaachinman, line 279:
```suggestion
Set the `ns` option, like in [this example](https://github.com/locize/next-i18next-locize/blob/main/next-i18next.config.js#L17). Doing this will ensure all translation resources are loaded on initialization.
```

README.md
```diff
@@ -260,13 +260,29 @@ To migrate from previous versions to the version 8, check out the [v8-migration
...
+
+**2) check the ready flag:**
```
isaachinman, line 281:
```suggestion
**2) Check the ready flag:**
```

README.md
```diff
@@ -260,13 +260,29 @@ To migrate from previous versions to the version 8, check out the [v8-migration
...
+
+You need to handle the not ready state by eg. render a loading component as long `ready === false` or `props.tReady === false`. Not doing so will result in rendering your translations before they loaded, which will cause save missing be called although translations exists (just yet not loaded).
```
isaachinman, line 283:
```suggestion
If you cannot or do not want to provide the `ns` array, calls to the `t` function will cause namespaces to be loaded on the fly. This means you'll need to handle the "not ready" state by checking `ready === true` or `props.tReady === true`. Not doing so will result in rendering your translations before they loaded, which will cause "save missing" be called despite the translations actually existing (just yet not loaded).
```

isaachinman (changes requested):
Thanks, this is really helpful!

isaachinman (approved):

PR #1548, opened 2021-11-24
refactor: prevent unexpected null locale

wulimao49:
currently if the app wrapped with `appWithTranslation` rerendes with no change on `_nextI18Next`, the `locale` will be `null` and causing the whole react tree unmount.

i intend to put `locale` into `useMemo` to keep the value as expected.


isaachinman:
Hi @wulimao49 – do we have an issue open for this? Can you provide a reproducible example?

PR #1593, opened 2021-12-22
Allow function type for localePath

### LINKED ISSUE #1500: localeStructure without sub-directory keeps searching for subdirectory
## localeStructure without sub-directory keeps searching for subdirectory
When changing the configuration to use a flat directory structure like this, so I can use files like `common.en.json` or `something-else.fr.json` there is an error thrown, expecting the language directory to be there.

configuration:
```
const path = require('path')

module.exports = {
  i18n: {
    defaultLocale: 'en',
    locales: ['en', 'de', 'fr'],
  },
  localeStructure: '{{ns}}.{{lng}}',
  localePath: path.resolve('./public/translations'),
}
```

Error:
```
Error: ENOENT: no such file or directory, scandir '/Users/con/work/project/app/public/translations/de'
at Object.readdirSync (fs.js:1047:3)
at getLocaleNamespaces (/Users/con/work/project/app/node_modules/next-i18next/dist/commonjs/config/createConfig.js:180:23)
at /Users/con/work/project/app/node_modules/next-i18next/dist/commonjs/config/createConfig.js:188:20
at Array.map (<anonymous>)
  at getNamespaces (/Users/con/work/project/app/node_modules/next-i18next/dist/commonjs/config/createConfig.js:185:44)
  at createConfig (/Users/con/work/project/app/node_modules/next-i18next/dist/commonjs/config/createConfig.js:228:29)
  at _callee$ (/Users/con/work/project/app/node_modules/next-i18next/dist/commonjs/serverSideTranslations.js:199:53)
  at tryCatch (/Users/con/work/project/app/node_modules/regenerator-runtime/runtime.js:63:40)
  at Generator.invoke [as _invoke] (/Users/con/work/project/app/node_modules/regenerator-runtime/runtime.js:294:22)
  at Generator.next (/Users/con/work/project/app/node_modules/regenerator-runtime/runtime.js:119:21) {
    errno: -2,
    syscall: 'scandir',
    path: '/Users/con/work/project/app/public/translations/de'
  }
```

8.8.0

## Steps to reproduce
change the configuration to the one above

## Expected behaviour
The loading of the files works as expected.

## Additional context
I did some debugging in the `createConfig.js` from node-modules and found out, that the `combinedConfig` variable is filled correctly and these variables also look correct:

```        
var defaultLocaleStructure = localeStructure.replace("".concat(prefix, "lng").concat(suffix), lng).replace("".concat(prefix, "ns").concat(suffix), defaultNS);
var defaultFile = "/".concat(defaultLocaleStructure, ".").concat(localeExtension);
var defaultNSPath = path.join(localePath, defaultFile);
var defaultNSExists = fs.existsSync(defaultNSPath);

console.log(defaultLocaleStructure);
console.log(defaultFile);
console.log(defaultNSPath);
console.log(defaultNSExists);
```

outputs:

```
common.de
/common.de.json
/Users/con/work/project/app/public/translations/common.de.json
true
```

So I think the problem relies in some un-configured namespace configuration below where path is concatinated without respecting the combined config.

```
var namespacesByLocale = locales.map(function (locale) {
return getLocaleNamespaces(path.resolve(process.cwd(), "".concat(serverLocalePath, "/").concat(locale)));
});
```

lucaslcode:
Fixes #1500

> `localePath` as a function is of the form `(locale: string, namespace: string, env: 'client' | 'server', missing: boolean) => string` returning the entire path including filename and extension. When `missing` is true, return the path for the `addPath` option of `i18next-fs-backend`, when false, return the path for the `loadPath` option. [More info at the `i18next-fs-backend` repo.](https://github.com/i18next/i18next-fs-backend/tree/master#backend-options)

See comments

Tests added ~~Wasn't sure how to add tests for the additional functionality - any guidance appreciated (maybe I should add tests for the error cases and that the function is correctly in the final config by calling it?)~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,16 +223,18 @@ This option will reload your translations whenever `serverSideTranslations` is c
 
 #### Options
 
-| Key                 | Default value        |
-| ------------------- | -------------------- |
-| `defaultNS`         | `'common'`           |
-| `localeExtension`   | `'json'`             |
-| `localePath`        | `'./public/locales'` |
-| `localeStructure`   | `'{{lng}}/{{ns}}'`   |
-| `reloadOnPrerender` | `false`              |
-| `serializeConfig`   | `true`               |
-| `strictMode`        | `true`               |
-| `use` (for plugins) | `[]`                 |
+| Key                 | Default value        | Note                                   |
+| ------------------- | -------------------- | -------------------------------------- |
+| `defaultNS`         | `'common'`           |                                        |
+| `localePath`        | `'./public/locales'` | Can be a function, see note below.     |
+| `localeExtension`   | `'json'`             | Ignored if `localePath` is a function. |
+| `localeStructure`   | `'{{lng}}/{{ns}}'`   | Ignored if `localePath` is a function. |
+| `reloadOnPrerender` | `false`              |                                        |
+| `serializeConfig`   | `true`               |                                        |
+| `strictMode`        | `true`               |                                        |
+| `use` (for plugins) | `[]`                 |                                        |
+
+`localePath` as a function is of the form `(locale: string, namespace: string, env: 'client' | 'server', missing: boolean) => string` returning the entire path including filename and extension. When `missing` is true, return the path for the `addPath` option of `i18next-fs-backend`, when false, return the path for the `loadPath` option. [More info at the `i18next-fs-backend` repo.](https://github.com/i18next/i18next-fs-backend/tree/master#backend-options)
 
 All other [i18next options](https://www.i18next.com/overview/configuration-options) can be passed in as well.
 
--- a/README.md
+++ b/README.md
@@ -74,8 +74,8 @@ This tells `next-i18next` what your `defaultLocale` and other locales are, so th
 ```js
 module.exports = {
   i18n: {
-    defaultLocale: 'en',
-    locales: ['en', 'de'],
+    defaultLocale: "en",
+    locales: ["en", "de"],
   },
 };
 ```
@@ -85,7 +85,7 @@ Now, create or modify your `next.config.js` file, by passing the `i18n` object i
 #### [`next.config.js`](https://nextjs.org/docs/api-reference/next.config.js/introduction)
 
 ```js
-const { i18n } = require('./next-i18next.config');
+const { i18n } = require("./next-i18next.config");
 
 module.exports = {
   i18n,
@@ -99,7 +99,7 @@ There are three functions that `next-i18next` exports, which you will need to us
 This is a HOC which wraps your [`_app`](https://nextjs.org/docs/advanced-features/custom-app):
 
 ```tsx
-import { appWithTranslation } from 'next-i18next';
+import { appWithTranslation } from "next-i18next";
 
 const MyApp = ({ Component, pageProps }) => <Component {...pageProps} />;
 
@@ -113,12 +113,12 @@ The `appWithTranslation` HOC is primarily responsible for adding a [`I18nextProv
 This is an async function that you need to include on your page-level components, via either [`getStaticProps`](https://nextjs.org/docs/basic-features/data-fetching#getstaticprops-static-generation) or [`getServerSideProps`](https://nextjs.org/docs/basic-features/data-fetching#getserversideprops-server-side-rendering) (depending on your use case):
 
 ```tsx
-import { serverSideTranslations } from 'next-i18next/serverSideTranslations';
+import { serverSideTranslations } from "next-i18next/serverSideTranslations";
 
 export async function getStaticProps({ locale }) {
   return {
     props: {
-      ...(await serverSideTranslations(locale, ['common', 'footer'])),
+      ...(await serverSideTranslations(locale, ["common", "footer"])),
       // Will be passed to the page component as props
     },
   };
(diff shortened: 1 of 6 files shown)


src/config/createConfig.ts
```diff
@@ -124,22 +136,29 @@ export const createConfig = (userConfig: UserConfig): InternalConfig => {
...
 
-    let clientLocalePath = localePath
-
     //
     // Remove public prefix from client site config
     //
-    if (localePath.match(/^\.?\/public\//)) {
-      clientLocalePath = localePath.replace(/^\.?\/public/, '')
-    }
+    const clientLocalePath = typeof localePath === 'string' && localePath.match(/^\.?\/public\//)
```
lucaslcode, line 142: Has to be const to satisfy typescript, since line 157 uses a closure.

lucaslcode: Additionally, it seems to me this stripping of `/public` on the client isn't necessary, since nothing is even read on the client - I'm not totally sure though.

isaachinman:
Thanks @lucaslcode! Just wanted to give a heads up that I am travelling now, and probably won't be able to review this until sometime in January. Rest assured I will work with you to get this merged as soon as possible.

lucaslcode:
Thanks for the update, no worries!

src/config/createConfig.ts
```diff
@@ -54,7 +54,7 @@ export const createConfig = (userConfig: UserConfig): InternalConfig => {
       // Validate defaultNS
       // https://github.com/isaachinman/next-i18next/issues/358
       //
-      if (typeof defaultNS === 'string' && typeof lng !== 'undefined') {
+      if (typeof serverLocalePath === 'string' && typeof defaultNS === 'string' && typeof lng !== 'undefined') {
```
isaachinman, line 57: If `typeof serverLocalePath === 'function'`, doesn't that make validation of `defaultNS` even easier? I think we should still validate in both cases.

src/config/createConfig.ts
```diff
@@ -69,15 +69,27 @@ export const createConfig = (userConfig: UserConfig): InternalConfig => {
...
       //
-      combinedConfig.backend = {
-        addPath: path.resolve(process.cwd(), `${serverLocalePath}/${localeStructure}.missing.${localeExtension}`),
-        loadPath: path.resolve(process.cwd(), `${serverLocalePath}/${localeStructure}.${localeExtension}`),
+      if (typeof serverLocalePath === 'string') {
+        combinedConfig.backend = {
+          addPath: path.resolve(process.cwd(), `${serverLocalePath}/${localeStructure}.missing.${localeExtension}`),
+          loadPath: path.resolve(process.cwd(), `${serverLocalePath}/${localeStructure}.${localeExtension}`),
+        }
+      } else {
```
isaachinman, line 77: Just to be safe:

```suggestion
      } else if (typeof serverLocalePath === 'function') {
```

src/config/createConfig.ts
```diff
@@ -69,15 +69,27 @@ export const createConfig = (userConfig: UserConfig): InternalConfig => {
...
+            serverLocalePath(locale, namespace, 'server', false),
+        }
       }
 
       //
       // Set server side preload (namespaces)
       //
       if (!combinedConfig.ns && typeof lng !== 'undefined') {
+        if (typeof serverLocalePath === 'function')
+          throw new Error('Must provide all namespaces in ns option if using a function as localePath')
```
isaachinman, line 91:
```suggestion
        if (typeof serverLocalePath === 'function') {
          throw new Error('Must provide all namespaces in ns option if using a function as localePath')
        }
```

src/config/createConfig.ts
isaachinman: Yeah, we can probably remove the stripping of `/public`. It's a remnant of when this package used to use `i18next-http-backend` by default, and used `getInitialProps`, necessitating client side loading of namespaces.

src/serverSideTranslations.ts
```diff
@@ -85,6 +85,9 @@ export const serverSideTranslations = async (
   })
 
   if (!Array.isArray(namespacesRequired)) {
+    if (typeof localePath === 'function')
+      throw new Error('Must provide namespacesRequired to serverSideTranslations when using a function as localePath')
```
isaachinman, line 89:
```suggestion
    if (typeof localePath === 'function') {
      throw new Error('Must provide namespacesRequired to serverSideTranslations when using a function as localePath')
    }
```

isaachinman (commented):
Thanks for your patience – finally had a chance to get to this.

Just a few comments to discuss a bit, but overall I am extremely impressed with both the quality of the code, and your ability to understand the wider repo.

Massive thanks for this contribution!

lucaslcode:
@isaachinman thanks for the review (and the compliment 😄). I've addressed the feedback, including removing the stripping of `/public` and therefore no longer including `env: 'server' | 'client'` in the callback parameters, since there isn't a use-case for it as far as I can tell.

README.md
```diff
@@ -74,8 +74,8 @@ This tells `next-i18next` what your `defaultLocale` and other locales are, so th
 ```js
 module.exports = {
   i18n: {
-    defaultLocale: 'en',
-    locales: ['en', 'de'],
+    defaultLocale: "en",
```
isaachinman, line 77: Can you omit your autoformatting on this file?

isaachinman (approved):

isaachinman:
Looks great @lucaslcode, thanks for all your help!

Weirdly this PR is not running tests in Circle – I think it's related to a CircleCI OAuth issue a couple weeks ago. Can you try rebasing against latest `master`? If that doesn't work, we may need to recreate the PR entirely.

lucaslcode:
I'll try a new PR

PR #1609, opened 2022-01-04
feat: Expose Translation react-i18next component

myrjola:
These changes export the [react-i18next Translation component](https://react.i18next.com/latest/translation-render-prop) from the next-i18next module. I found this component the best match for introducing translations in the [Page.getLayout]( https://nextjs.org/docs/basic-features/layouts#per-page-layouts) function. The useTranslation hook is [not allowed because the getLayout function is not top-level](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level).

I also added test coverage for the Translation component by using it on the second page to render the header. It's tested by `yarn test:e2e`.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -1,4 +1,4 @@
-export { I18nContext, Trans, useTranslation, withTranslation } from 'react-i18next'
+export { I18nContext, Trans, Translation, useTranslation, withTranslation } from 'react-i18next'
 
 export { appWithTranslation, globalI18n as i18n } from './appWithTranslation'
 
--- a/src/types.ts
+++ b/src/types.ts
@@ -7,6 +7,7 @@ import {
   WithTranslation as ReactI18nextWithTranslation,
   Resources,
   DefaultNamespace,
+  Translation,
 } from 'react-i18next'
 import { InitOptions, i18n as I18NextClient, TFunction as I18NextTFunction } from 'i18next'
 import { appWithTranslation, i18n } from './'
@@ -63,6 +64,7 @@ export {
   appWithTranslation,
   useTranslation,
   Trans,
+  Translation,
   withTranslation,
   Resources,
   DefaultNamespace,


examples/simple/pages/second-page.js
```diff
@@ -13,7 +13,9 @@ const SecondPage = () => {
   return (
     <>
       <main>
-        <Header heading={t('h1')} title={t('title')} />
+        <Translation ns='second-page'>
+          {(translate) => <Header heading={translate('h1')} title={translate('title')}/> }
+        </Translation>
```
kachkaev, line 18: Re-exporting the component won’t harm but I’m not sure if `<Translation>` should become a part of the example.

Render props and HOCs are are deprecated pattern and can be substituted with hooks in modern react apps. Those who don’t understand this distinction well may look at the example and end up with a more complex syntax than they otherwise could.

myrjola: Thanks for bringing the deprecation to my attention! Where is it documented? I could not see any deprecation notice on https://react.i18next.com/latest/translation-render-prop.

I wonder what would be a better approach if I would need translations in the [Next.js getLayout pattern](https://nextjs.org/docs/basic-features/layouts#per-page-layouts).

Specifically, I'd like to translate the page title. Since getLayout is not a React component but a function, I can't use the useTranslation hook. Instead of using `<Translation>`, would it make sense to pass the t function to the getLayout function as a parameter?

```jsx
export default function MyApp({ Component, pageProps }) {
  const { t } = useTranslation()

  // Use the layout defined at the page level, if available
  const getLayout = Component.getLayout || ((page) => page)

  return getLayout(<Component {...pageProps} />, t)
}
```

```jsx
Page.getLayout = function getLayout(page, t) {
  return (
    <Layout>
      <Head>
        <title>{t('pageTitle')}</title>
      </Head>
      <NestedLayout>{page}</NestedLayout>
    </Layout>
  )
}
```

kachkaev: I don’t know if HOCs and render props were deprecated ‘officially’ anywhere. They achieve the same results as hooks do, and are used much less these days.

In your `getLayout` pattern you can get access to `useTranslation()` as long as you call it inside some component:

```tsx
function MyCoolLayout() {
  const t = useTranslation()
 
  return (
    <Layout>
      <Head>
        <title>{t('pageTitle')}</title>
      </Head>
      <NestedLayout>{page}</NestedLayout>
    </Layout>
  )
}

Page.getLayout = function getLayout(page, t) {
  return (
    <MyCoolLayout>{page}</MyCoolLayout>
  )
}
```

Not sure if `<Head><title>{t('pageTitle')}</title></Head>` belongs to the layout, but that’s outside this PR scope.

myrjola: I removed the changes to second-page.js to not set a bad example. Regarding test coverage, I think we don't have coverage for the <Trans> component either, but we can probably rely on react-i18next to work as it should.

> In your getLayout pattern you can get access to useTranslation() as long as you call it inside some component:

Yes! Separating <Head> to its own component is cleaner. Thank you!

isaachinman (approved):
Looks good, we just need CI to run. Circle should be building forked PRs, so not sure why the status checks haven't come back. You can try pushing an empty commit.

Thanks @myrjola!

myrjola:
Thank you for a great package @isaachinman! I pushed a new commit, but it seems the codeclimate and node-multi-build checks are not running. It seems that this other PR also has the same issue https://github.com/isaachinman/next-i18next/pull/1593

Is there something wrong with the Circle CI integration?

isaachinman:
Not sure @myrjola, but it appears something is indeed wrong for forked PRs. I've regenerated the CircleCI ssh key – do you mind trying again? Thanks!

isaachinman (changes requested):
Apologies for not catching this before, I was travelling during the holidays. [This file](https://github.com/isaachinman/next-i18next/blob/master/src/types.ts) will need to be updated for TS users. Unfortunately we're still maintaining types manually.

myrjola:
Added the Translation export to the types.ts file.

isaachinman (approved):

isaachinman:
Thanks @myrjola!

PR #1631, opened 2022-01-17
fix: appWithTranslation useMemo bug (null locale)

### LINKED ISSUE #1628: Next/Link causing app to unmount
## Describe the bug
For some reason if you use a next/link component and click the same link twice, the entire app will unmount and mount itself, causing some parts of my application to behave weirdly. Upon further inspection, clicking a link your already on will set the locale key on App to null (which is probably what's causing the issue). It seems that this only happens in production (im using vercel). 

**next-i18next.config.js**
`module.exports = {
  i18n: {
    defaultLocale: "en-US",
    locales: ["en-US", "sv-SE"],
  },
};
`

10.1.0

## Steps to reproduce
I have created a minimal setup [repo](https://github.com/JonahLund/next-i18next_link_bug) that's deployed to [vercel](https://next-i18next-link-bug.vercel.app/)

## Expected behaviour
The app should not unmount if you click on a next/link component twice.

## Screenshots
Everything fine
image
Until you click on the same link again
image

JonahLund:
This PR will fix the Next/Link app unmount.

I placed the locale in a state instead of a variable, which fixes locale getting set to null.

The state will only change if the `initialLocale` has changed and is not equal to the current state and is not null.

Fixes #1628

### DIFF
--- a/src/appWithTranslation.tsx
+++ b/src/appWithTranslation.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo } from 'react'
+import React, { useMemo, useState } from 'react'
 import hoistNonReactStatics from 'hoist-non-react-statics'
 import { I18nextProvider } from 'react-i18next'
 import type { AppProps as NextJsAppProps } from 'next/app'
@@ -21,10 +21,9 @@ export const appWithTranslation = <Props extends AppProps = AppProps>(
   WrappedComponent: React.ComponentType<Props>,
   configOverride: UserConfig | null = null,
 ) => {
-  let locale: string | null = null
-
   const AppWithTranslation = (props: Props) => {
     const { _nextI18Next } = props.pageProps as SSRConfig
+    const [locale, setLocale] = useState<string | undefined>(undefined)
 
     // Memoize the instance and only re-initialize when either:
     // 1. The route changes (non-shallowly)
@@ -35,7 +34,7 @@ export const appWithTranslation = <Props extends AppProps = AppProps>(
       let { userConfig } = _nextI18Next
       const { initialI18nStore, initialLocale } = _nextI18Next
 
-      locale = initialLocale
+      setLocale(initialLocale)
 
       if (userConfig === null && configOverride === null) {
         throw new Error('appWithTranslation was called without a next-i18next config')
@@ -52,9 +51,9 @@ export const appWithTranslation = <Props extends AppProps = AppProps>(
       const instance = createClient({
         ...createConfig({
           ...userConfig,
-          lng: locale,
+          lng: locale || initialLocale,
         }),
-        lng: locale,
+        lng: locale || initialLocale,
         resources: initialI18nStore,
       }).i18n
 
--- a/src/appWithTranslation.tsx
+++ b/src/appWithTranslation.tsx
@@ -23,7 +23,7 @@ export const appWithTranslation = <Props extends AppProps = AppProps>(
 ) => {
   const AppWithTranslation = (props: Props) => {
     const { _nextI18Next } = props.pageProps as SSRConfig
-    const [locale, setLocale] = useState<string | undefined>(_nextI18Next ? _nextI18Next.initialLocale : undefined) // eslint-disable-line
+    const [locale, setLocale] = useState<string>(_nextI18Next?.initialLocale)
 
     // Memoize the instance and only re-initialize when either:
     // 1. The route changes (non-shallowly)
--- a/src/appWithTranslation.tsx
+++ b/src/appWithTranslation.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useState } from 'react'
+import React, { useMemo } from 'react'
 import hoistNonReactStatics from 'hoist-non-react-statics'
 import { I18nextProvider } from 'react-i18next'
 import type { AppProps as NextJsAppProps } from 'next/app'
@@ -23,7 +23,7 @@ export const appWithTranslation = <Props extends AppProps = AppProps>(
 ) => {
   const AppWithTranslation = (props: Props) => {
     const { _nextI18Next } = props.pageProps as SSRConfig
-    const [locale, setLocale] = useState<string>(_nextI18Next?.initialLocale)
+    const locale: string | null = _nextI18Next?.initialLocale ?? null
 
     // Memoize the instance and only re-initialize when either:
     // 1. The route changes (non-shallowly)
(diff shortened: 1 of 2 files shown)


src/appWithTranslation.tsx
```diff
@@ -23,7 +23,7 @@ export const appWithTranslation = <Props extends AppProps = AppProps>(
 ) => {
   const AppWithTranslation = (props: Props) => {
     const { _nextI18Next } = props.pageProps as SSRConfig
-    let locale = null
+    const [locale, setLocale] = useState<string>(_nextI18Next?.initialLocale)
```
isaachinman, line 26:
```suggestion
    const [locale, setLocale] = useState<string>(_nextI18Next?.initialLocale ?? null)
```

isaachinman (commented):
Thanks! Do you mind writing some test coverage?

JonahLund:
I think i came up with a better and simpler solution that doesnt require any state hooks

JonahLund:
Good thing those tests were there

JonahLund:
> Thanks! Do you mind writing some test coverage?

I assume that the test coverage isn't needed now that it passed?

JonahLund:
It seems that there were already a PR addressing this issue #1548

wulimao49:
there is a workaround
`memo(appWithTranslation(MyApp))`

isaachinman:
@JonahLund I didn't get a response from @wulimao49 on #1548, so wasn't sure how to proceed. In any case, I think the solution here is simpler, so let's proceed with this PR.

Can we just add basic tests that pass `_nextI18Next?.initialLocale` as some expected value, and assert that `locale` gets set to it?

Specifically it would be great if we could actually write a test that reproduces the buggy `null` behaviour we are fixing, to ensure we do not have any regressions in the future.

JonahLund:
> @JonahLund I didn't get a response from @wulimao49 on #1548, so wasn't sure how to proceed. In any case, I think the solution here is simpler, so let's proceed with this PR.
> 
> Can we just add basic tests that pass `_nextI18Next?.initialLocale` as some expected value, and assert that `locale` gets set to it?
> 
> Specifically it would be great if we could actually write a test that reproduces the buggy `null` behaviour we are fixing, to ensure we do not have any regressions in the future.

Any idea of how to do that? I tried writing a test that should fail in the non fixed version but it doesn't seem to do that.

isaachinman:
All it should take is re-rendering `appWithTranslation` with the same `locale` a few times, right?

JonahLund:
> All it should take is re-rendering `appWithTranslation` with the same `locale` a few times, right?

Yeah exactly

isaachinman:
Only thing that comes to mind is that this `useMemo` bug only seems to occur in production mode, so perhaps the default jest env won't work.

JonahLund:
> Only thing that comes to mind is that this `useMemo` bug only seems to occur in production mode, so perhaps the default jest env won't work.

Yeah maybe, i tried this but the key stays 'de', so i dont know if im using the `.mock.calls` wrong or if the `rerender` function doesn't work the way i think it works, im definitely not any professional at jest testing 😂
image

isaachinman:
According to [the docs](https://testing-library.com/docs/example-update-props/), the `rerender` function should work. Do you mind having another go, or sharing what WIP code you have?

JonahLund:
> According to [the docs](https://testing-library.com/docs/example-update-props/), the `rerender` function should work. Do you mind having another go, or sharing what WIP code you have?

Yeah so basically we just need to access the key prop to verify that it's not null. And im not entirely sure how to do that via the react testing library

isaachinman:
You can add a `data-locale` (or similar) attribute if it helps for testing purposes 👍

JonahLund:
Could also just throw an error if locale is null, and then just make sure it doesnt throw an error in the test.

JonahLund:
Well if _nextI18Next isnt null and locale is*

isaachinman:
If we wanted to throw errors, we would need to bump the major version. Let's just stick with a bug fix for now.

JonahLund:
> Hey @JonahLund – looks like we've got some failing tests.
> 
> This issue is affecting a fair few users, so it'd be great to get it merged as soon as we can.
> 
> Please let me know if you've still got time for this, otherwise I don't mind taking over. Thanks!

Hello, I wasn't able to successfully reproduce the bug in the testing coverage,  so that's why I haven't made any new commits.

Yeah the only thing left is rerendering the component and making sure the `key` prop isn't null. Which is what i was unable to do. So if you are able to reproduce it then that would be great :)

JonahLund:
The bug fix works as it should, just not the latest testing coverage i committed.

JonahLund:
Ok I will give it another try with the tests.

JonahLund:
I tried with a different react testing library, was easier to get all properties from the component, but still can't seem to reproduce the `null` behaviour. This is what i came up with

When running this test on the bugged version, it passes, which it shouldn't.

`it('assures locale key is set to the right value', () => {
    const component = renderer.create(<DummyApp
      {...(createProps('de'))}
    />)

    component.update(<DummyApp
      {...(createProps('en'))}
    />)

    component.update(<DummyApp
      {...(createProps('en'))}
    />)

    const getKeyProp = ((root: any) => root._fiber.child.child.key)

    expect(getKeyProp(component.root)).toBe('en')
  })`

Maybe this is something that has to be e2e tested in cypress, im not sure.

isaachinman:
@JonahLund Thanks again for spending so much time on this. Let me step in and help, as it's Sunday and I have a few minutes. It's absolutely possible to introspect `children.key`, and I believe that's the right approach here. The only thing you were missing is that the `createProps` function returns a new object every time, so memoisation was actually not taking place.

It's a little cumbersome, but we can "trick" memoisation into taking effect by mutating the same object in place:

```tsx
it('assures locale key is set to the right value', () => {
  const props = createProps('de')

  const { rerender } = render(<DummyApp
    {...props}
  />)

  props.router.locale = 'en'
  props.pageProps._nextI18Next.initialLocale = 'en'

  rerender(<DummyApp
    {...props}
  />)
  rerender(<DummyApp
    {...props}
  />)
  rerender(<DummyApp
    {...props}
  />)

  props.router.locale = 'de'
  props.pageProps._nextI18Next.initialLocale = 'de'

  rerender(<DummyApp
    {...(createProps('de'))}
  />)

  const [[first], [second], [third], [fourth], [fifth]] =
    (I18nextProvider as jest.Mock).mock.calls

  expect(first.children.key).toEqual('de')
  expect(second.children.key).toEqual('en')
  expect(third.children.key).toEqual('en')
  expect(fourth.children.key).toEqual('en')
  expect(fifth.children.key).toEqual('de')
})
```

Crucially, note that if you revert the actual source code changes, this test will fail.

If you can add this test coverage, I will merge immediately. Thanks!

JonahLund:
Awesome! :)

isaachinman (approved):

isaachinman:
Thank you @JonahLund!

PR #1681, opened 2022-02-10
docs: Clarify config for custom interpolation suffix/prefix

joostdecock:
It took me a while to figure out how to get a custom interpolation prefix/suffix to work.

I think it's a little unintuitive that changing an option that relates to how the text inside the translation files is structured has a knock-on effect to how files are loaded. That is not criticism, I realize it's because of how the various packages interact, but it's still a bit of a head-scratcher.

After doing some digging, I came across #1090 that clarifies the issue, but I feel it's worth putting this in the README.

This PR does that. It's not much, but I hope you find it useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,6 +254,27 @@ const Component = () => {
 }
 ```
 
+#### Custom interpolation prefix/suffix
+
+By default, i18next uses `{{` as prefix and `}}` as suffix for [interpolation](https://www.i18next.com/translation-function/interpolation).
+If you want/need to override these interpolation settings, you **must** also specify an alternative `localeStructure` setting that matches your custom prefix and suffix.
+
+For example, if you want to use `{` and `}` the config would look like this:
+
+```js
+{
+  i18n: {
+    defaultLocale: 'en',
+    locales: ['en', 'nl'],
+  },
+  interpolation: {
+    prefix: '{',
+    suffix: '}',
+  },
+  localeStructure: '{lng}/{ns}',
+}
+```
+
 ## Migration to v8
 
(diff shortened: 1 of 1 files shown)


isaachinman:
Hey @joostdecock – #1248 was supposed to fix this issue, and `localeStructure` should respect the interpolation options. Are you saying you've noticed a regression in that regard?

joostdecock:
It works, but only if you (also) add the `localeStructure` setting.
If my reading of #1248 is correct, that is also what is supposed to happen (as in, the example clearly sets the `localeStructure` setting.


This example lifted from #1248 will work as-is:
```js
module.exports = {
  i18n: {
    defaultLocale: 'en',
    locales: ['en', 'de'],
  },
  interpolation: {
    prefix: '%{',
    suffix: '}',
  },
  localeStructure: '%{lng}/%{ns}',
}
```

But removing the `localeStructure` setting will make this not work:
```js
module.exports = {
  i18n: {
    defaultLocale: 'en',
    locales: ['en', 'de'],
  },
  interpolation: {
    prefix: '%{',
    suffix: '}',
  }
}
```


So it's easy enough to get it to work **if** you know that you **must** set `localeStructure` to match your custom prefix/suffix.
I found that confusing at first since the prefix/suffix setting relates to how placeholders for interpolation are escaped inside your data. So it's not really obvious that this has a knock-on effect on locating the files themselves.

So I felt some clarification would be best.
I hope I'm managing to explain myself :joy:

isaachinman:
Yep, makes sense, and thanks for a detailed explanation. In the end, these are just raw, user-config strings, and it wouldn't be possible to detect and swap custom prefix/suffix values.

Happy to merge this, thanks very much for your contribution!

isaachinman (approved):

PR #1744, opened 2022-03-16
Feature: allow function returning promise to be used in the user config

IdkMan2Usertive:
Thanks to this improvement we can use function returning Promise (or not) in `next-i18next.config.js`.
This is a must-have feature for our monorepo project where we don't know early which locales will be used in the build so I have to use async/await stuff to determine the locales.

### DIFF
--- a/src/serverSideTranslations.ts
+++ b/src/serverSideTranslations.ts
@@ -43,7 +43,9 @@ export const serverSideTranslations = async (
     throw new Error('Initial locale argument was not passed into serverSideTranslations')
   }
 
-  let userConfig = configOverride
+  let userConfig: UserConfig
+    | (() => UserConfig | Promise<UserConfig>)
+    | null = configOverride
 
   if (!userConfig && fs.existsSync(path.resolve(DEFAULT_CONFIG_PATH))) {
     userConfig = await import(path.resolve(DEFAULT_CONFIG_PATH))
@@ -51,6 +53,10 @@ export const serverSideTranslations = async (
 
   if (userConfig === null) {
     throw new Error('next-i18next was unable to find a user config')
+  }
+
+  if (typeof userConfig === 'function') {
+    userConfig = await userConfig()
   }
 
   const config = createConfig({


IdkMan2Usertive:
I have a large monorepo which is producing websites, so if I run command like `yarn generate-website dummy-website`  my scripts are traversing the codebase to look for the `dummy-website`'s metadata. As each website has its own languages (and even default language) its metadata defines which languages are used in the process of building this website. This is why I can't define i18n configuration on postinstall event, because this information is determined on the fly.

Quite strange, but I hope that you understand that's what business needs 😉.

IdkMan2Usertive:
If it helps - this is how my config looks like right now:
```
const {determineActiveProduct, getProductInfo} = require('@dev7/altar2-tools');

module.exports = async () => {
  /**
   * @type {import('@dev7/altar2-tools').RegisteredProduct}
   */
  const registeredProduct = await determineActiveProduct();

  /**
   * @type {import('@dev7/altar2-tools').ProductInfo}
   */
  const productInfo = await getProductInfo(registeredProduct.name);

  return {
    i18n: {
      locales: productInfo.locales,
      defaultLocale: productInfo.defaultLocale,
    },
    localePath: './src/locales',
    localeStructure: '{{lng}}/{{ns}}',
    reloadOnPrerender: process.env.NODE_ENV !== 'production',
    defaultNS: 'common',
    serializeConfig: false,
    strictMode: true,
  };
};
```

isaachinman:
Sorry, I still do not understand why you cannot perform that step as prebuild.`await getProductInfo(registeredProduct.name).locales` would work just as fine in a script run 1 second before your build.

IdkMan2Usertive:
> Sorry, I still do not understand why you cannot perform that step as prebuild. `await getProductInfo(registeredProduct.name).locales` would work just as fine in a script run 1 second before your build.

But how would I pass locales array from `prerebuild` script to `next-i18next.config.js` file? Using environment variables?

IdkMan2Usertive:
Okay, everything was working great until I have defined `localePath` option as a function. 

Previously I was awaiting the config generation function in the getStaticProps and then I was passing the newly created config to the front-end using props. From now config is unserializable and I cannot do it this way.

So I have found another solution which IMO is close to what you have proposed:
1. I have marked config as unserializable `serializeConfig: false` option.

screenshot1

2. I have created a command that is fired just before the build step (or dev server startup). This command is generating all needed information about website, running required asynchronous stuff and saves it in the `.active-product.json` file, next to the `next-i18next.config.js` file.

screenshot2

3. And I am simply importing the website's information (`.active-product.json`) synchronously using `require`.

screenshot3

Quite surprising - it works both - on the back-end & front-end.

IdkMan2Usertive:
So this PR is no longer relevant in my case, feel free to close or merge, whatever.

PR #1747, opened 2022-03-19
fix: Update readme to include localePath for usage with getServerSideProps

egecavusoglu:
Hey, I lost quite a bit of time trying to figure out why my deployment on Vercel was not working when using getServerSideProps and by looking into a lot of issues I was able to figure it out. I thought it would be helpful to include this config update in the docs, so here is my PR. Thanks!


isaachinman:
This info is already in the documentation [here](https://github.com/isaachinman/next-i18next#vercel-and-netlify).

PR #1912, opened 2022-07-17
fix: Pass namespaces to the client, closes #1839

### LINKED ISSUE #1839: Namespaces option is not synced on the client.
## Describe the bug
`options.ns` is not propagated to the client properly, it contains some namespace on the server, but the instance on the client doesn't

11.0.0

## Steps to reproduce
1. Clone [simple example](https://github.com/isaachinman/next-i18next/tree/master/examples/simple)
2. Change the `dev` script to `NODE_OPTIONS='--inspect' next`, which will allow you to debug next server
3. Add break point at the end of `serverSideTranslations`, check `i18n.options.ns`, it should contain all user required namespaces.
4. Add break point of the client, check `i18n.options.ns` it will contain only `common`.

## Expected behaviour
`i18n.options.ns` on the client should reflect the same value as in the server.

### Server side
image

### Client side
image

 - Device: MBP 2019
 - Browser:  Chrome 101

felixmosh:
#### Checklist

- [x] tests are included

### DIFF
--- a/src/appWithTranslation.tsx
+++ b/src/appWithTranslation.tsx
@@ -25,6 +25,7 @@ export const appWithTranslation = <Props extends AppProps = AppProps>(
     const { _nextI18Next } = props.pageProps as SSRConfig
     let locale: string | null =
       _nextI18Next?.initialLocale ?? props?.router?.locale
+    const ns = _nextI18Next?.ns
 
     // Memoize the instance and only re-initialize when either:
     // 1. The route changes (non-shallowly)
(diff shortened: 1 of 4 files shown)


adrai:
Thank you @felixmosh.
Do you think it is ok, to release this with a patch version?
Or do you think this could influence existing setups?

felixmosh:
Yeah, It is a patch to a bug :]

adrai:
Thank you for your contribution.
It's included in v11.2.1

PR #1916, opened 2022-07-22
Fix getFixedT function always returning values in current language

Jyrno42:
I was running into an issue where `getFixedT` always returned values in the current language not the requested language. I found  this issue - https://github.com/i18next/next-i18next/issues/1184 about it. The fix for it was merged in and released and seemed to work. However a bit later the fix got reverted as people did not expect all locales to be loaded by the library. See the following issue - https://github.com/i18next/next-i18next/issues/1267.

This PR proposes a potential new argument to the `serverSideTranslations` function (could also be a config override field really). This argument can be used to provide extra locales to load for the specific page and allows one to opt into loading more or all locales. As a result `getFixedT` will then work for the locales that are loaded for the page.

> If the general direction/idea is accepted I will clean this PR up and will add tests/documentation updates too.

#### Checklist

- [x] tests are included

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,7 +159,32 @@ To do that, you can pass an array of required namespaces for each page into `ser
 
 Note: `useTranslation` provides namespaces to the component that you use it in. However, `serverSideTranslations` provides the total available namespaces to the entire React tree and belongs on the page level. Both are required.
 
-### 5. Advanced configuration
+### 5. Declaring locale dependencies
+
+By default, `next-i18next` will send _only the active locale_ down to the client on each request. This helps reduce the size of the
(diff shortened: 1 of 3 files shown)


adrai:
sounds interesting... can you also adapt the readme and add a test?

Jyrno42:
Sure I'll do that.

Do you think the extra argument is fine or should I move it into `configOverride`?

adrai:
configOverride is the passed in next-i18next config... so I think the extra argument should be ok.

adrai:
Thank you for your contribution.
Included in v11.3.0

Issue #1412, opened 2021-09-07
i18next properties dont seem to do anything


MauriceNino:
## Describe the bug
I want to support sub-locales. So for example, if someone calls my website with `Accept-Language: de-DE`, then `de` should be loaded.

For this, there seems to be the property `nonExplicitSupportedLngs` in i18next. I tried adding it to the config, but it is being ignored.

```js
// next-i18next.config.js
const isProd = process.env.NODE_ENV === 'production';

module.exports = {
  i18n: {
    defaultLocale: 'en',
    locales: ['en', 'de']
  },
  fallbackLng: 'en',
  supportedLngs: ['en', 'de'],
  nonExplicitSupportedLngs: true,
  load: 'all',
  reloadOnPrerender: !isProd
};
```

## Occurs in next-i18next version
v8.7.0

## Steps to reproduce
This commit on my repository has the "bug": https://github.com/MauriceNino/portfolio/commit/792997f0b7df0de41e592fb58ac672d2b523afc9

You can also send a GET request against the live version:

```http
# Does not work
GET mauz.io HTTP/1.1
Accept-Language: de-DE

###

# Works
GET mauz.io HTTP/1.1
Accept-Language: de
```

isaachinman:
All handling of locales is done by NextJs directly. If you want `de-DE` to fallback to `de`, you will need to investigate how to make that happen via NextJs itself. Good luck!

MauriceNino:
What are the i18next properties for, then? Do I even pass them at the correct position?  @isaachinman

isaachinman:
The i18next properties all work as normal, with the caveat that NextJs dictates the current locale, as it handles internationalised routing. Hope that makes sense.

MauriceNino:
Makes sense. Can I somehow implement a custom language detection function? Seems like it was possible with older versions of this library (#208).

If not, are there any plans on supporting that again?

Thanks a lot for taking your time for me!

isaachinman:
> Can I somehow implement a custom language detection function?

Not at the moment, to my knowledge.

> Seems like it was possible with older versions of this library

Yes, that's correct. Previously, `next-i18next` handled all aspects of localisation, including detection and routing. The Vercel team's ultimate goal is to support internationalisation first class, completely, but they decided to start with just the detection + routing aspect.

Therefore, any detection or routing concerns should be forwarded to the NextJs repo.

MauriceNino:
Understood, thanks a lot!

[...]

PR #1930, opened 2022-08-21
Add support for nonExplicitSupportedLngs

chrilis:
This PR adds support for `nonExplicitSupportedLngs` and is related to #1412. This features adds the possibility to support multiple variants of a language (`de-AT`, `de-CH` and `de-DE`) with a single JSON file (e.g. `public/locales/de/common.js`).

During the work on this PR I noticed that the the `default` value in `fallbackLng` should always be loaded. At least this is how I interprete this [testcase](https://github.com/i18next/next-i18next/pull/1930/files#diff-c87087f0d6c0b23943183ea270e1b1926cdf23e8bc7b7ae013762816561db596L152). Although this is related to #1927, the fix is part of this PR, because `nonExplicitSupportedLngs` depends on the same logic. Regarding the test-case above: I added `en-US` to supported `locales` and removed `en` from the `initialI18nStore`, because I think this should be the expected behavior. Is this correct @adrai?

Currently the combination `fallbackLng` with a function and `nonExplicitSupportedLngs: true` is not supported. Mainly because I didn't know how this could be implemented 😅 But I guess that the use cases for this combination should be very rare.

I'm looking forward to your feedback @adrai, thank you! 

#### Checklist

- [x] tests are included

#### Checklist (for documentation change)

- [x] only relevant documentation part is changed (make a diff before you submit the PR)
- [x] motivation/reason is provided

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,6 +184,30 @@ As a result the translations for both `no` and `en` locales will always be loade
 
 > Note: The extra argument should be added to all pages that use `getFixedT` function.
 
+#### Fallback locales
+By default, `next-i18next` will add the `defaultLocale` as fallback. To change this, you can set [`fallbackLng`](https://www.i18next.com/principles/fallback). All values supported by `i18next` (`string`, `array`, `object` and `function`) are supported by `next-i18next` too.
+
+Additionally `nonExplicitSupportedLngs` can be set to `true` to support all variants of a language, without the need to provide JSON files for each of them. Notice that all variants still must be included in `locales` to enable routing within `next.js`.
+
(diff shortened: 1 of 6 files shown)


adrai:
lgtm

adrai:
This looks really good to me.
Do you think this could be released with a minor version, or should we go with a major?

chrilis:
I'm not sure. The changes could lead to more languages being loaded, which would increase the page size. I tend to think of that as a breaking change - what do you think?

adrai:
just released v12.0.0... thank you very much for your great contribution @chrilis.

PR #1943, opened 2022-09-04
Fix fallbackLng if namespaces are undefined

### LINKED ISSUE #1941: Error if folder doesn't exist when using `fallbackLng`
## 🐛 Bug Report

- We run multiple domains, some of which are using the same `defaultLocale`
- Next doesn't support specifying the same `defaultLocale` for multiple domains
  - https://github.com/vercel/next.js/discussions/37991
  - https://stackoverflow.com/questions/73410976/next-js-i18n-2-domains-with-the-save-default-locale 
- To get past this, we specify unique variants of languages e.g. English (`en-DE`, `en-EN`) and rely on `fallbackLng`  to serve the correct language using the configuration below:

**`next-i18next.config.js`**

```JS
module.exports = {
  i18n: {
    defaultLocale: "en",
    locales: ["en", "en-DE", "en-EN"],
  },
  fallbackLng: {
    default: ["en"],
    "en-DE": ["en"],
  },
  debug: process.env.NODE_ENV === 'development',
  reloadOnPrerender: process.env.NODE_ENV === 'development',
};

```

**`next.config.js`**

```JS
const { i18n } = require("./next-i18next.config");

/** @type {import('next').NextConfig} */
const nextConfig = {
  reactStrictMode: true,
  swcMinify: true,
  i18n: {
    ...i18n,
    localeDetection: false,
    domains: [
      {
        domain: "next-i18next.com",
        defaultLocale: "en-EN",
        locales: ["en-EN"],
      },
      {
        domain: "de.next-i18next.com",
        defaultLocale: "en-DE",
        locales: ["en-DE"],
      },
    ],
  },
};

module.exports = nextConfig;

```

The above works, but only if the folder for that fallback language exists, if the folder doesn't exist we run into this error:

oops

It seems that we're not checking whether the locale folder exists before we attempt to read from it:

https://github.com/i18next/next-i18next/blob/master/src/serverSideTranslations.ts#L80-L88

## To Reproduce

- Checkout this repo featuring the issue: https://github.com/aaimio/next-i18next-bug
- Set up two domains in your `/etc/hosts`
  - `127.0.0.1 de.next-i18next.com`
  - `127.0.0.1 next-i18next.com`
- `npm install`
- `npm run dev`
- In your browser, visit `de.next-i18next.com` you will get the same error

## Expected behavior

We should first check whether the folder exists before reading from it.

- If default locale is specified, and its folder **doesn't exist**, fallback to value in `fallbackLng`, use that language's JSON file
- If default locale is specified, and its folder **exists**, use that language's JSON file

## Your Environment

- *Runtime version*: v18.4.0
- *next-i18next version*: 12.0.0
- *OS*: Mac

chrilis:
Fixes #1941

I did not understand why the `existsSync` mock returns `false` on the first call for all test cases in `serversideTranslations.tests.tsx`. Maybe it is related to #1302 and therefore obsolete? I have changed it to always return `true`, but since this change effects all tests, I'm not sure if this is the intended behaviour.

Additionally, it might make sense to "enable" server-side environment for all tests with something like
```
 beforeAll(() => {
      Object.assign(process, { browser: false })
      delete (global as any).window
    })
```
as is already the case for `createConfig.tests.tsx`. Currently `createConfig` is running in "client.side mode", which may not be the desired behavior for testing `getServerSideTranslations`? If this sounds reasonable to you @adrai, I can add changes to this PR or open another one - whichever you prefer.

#### Checklist

- [x] tests are included

### DIFF
--- a/src/serverSideTranslations.test.tsx
+++ b/src/serverSideTranslations.test.tsx
@@ -41,8 +41,7 @@ const renderDummyComponent = () =>
 
 describe('serverSideTranslations', () => {
   beforeEach(() => {
-    (fs.existsSync as jest.Mock).mockReturnValueOnce(false);
-    (fs.existsSync as jest.Mock).mockReturnValueOnce(true);
+    (fs.existsSync as jest.Mock).mockReturnValue(true);
     (fs.readdirSync as jest.Mock).mockReturnValue([])
   })
   afterEach(jest.resetAllMocks)
@@ -65,7 +64,7 @@ describe('serverSideTranslations', () => {
           locales: ['en-US', 'fr-CA'],
         },
       } as UserConfig)
-      expect(fs.existsSync).toHaveBeenCalledTimes(0)
+      expect(fs.existsSync).toHaveBeenCalledTimes(1)
       expect(fs.readdirSync).toHaveBeenCalledTimes(1)
       expect(fs.readdirSync).toHaveBeenCalledWith(expect.stringMatching('/public/locales/en-US'))
       expect(props._nextI18Next.initialI18nStore)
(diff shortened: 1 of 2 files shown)


adrai:
yes, sounds reasonable... just do it here no problem @chrilis

adrai:
Thank you @chrilis, it's included in v12.0.1

PR #1951, opened 2022-09-15
Move i18next and react-i18next to peerDependencies

rikkit:
This PR moves `i18next` and `react-i18next` to peerDependencies, so that consuming Yarn PnP projects aren't force to have multiple instances of `i18next` if they themselves depend on `react-i18next`.

After this PR, projects that use `next-i18next` will have to ensure they have `react-i18next` and `i18next` in their own dependencies: 

```sh
npm i react-i18next i18next
yarn add react-i18next i18next
```

See discussion in https://github.com/i18next/next-i18next/issues/1917

#### Checklist

- [x] only relevant code is changed (make a diff before you submit the PR)
- [x] run tests `npm run test`
- [ ] ~tests are included~
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/i18next/.github/blob/master/CONTRIBUTING.md)

#### Checklist (for documentation change)

- [x] only relevant documentation part is changed (make a diff before you submit the PR)
- [x] motivation/reason is provided
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/i18next/.github/blob/master/CONTRIBUTING.md)

PR #1966, opened 2022-10-19
Support i18next 22, react-i18next 12 and move both to peer-deps

belgattitude:
Move `react-i18next` and `i18next` to peerDependencies and update them to latest

## Why ?

- If we want to support 2 major versions of i18next, that's the way.
- react-i18next / next-i18next are sensitive to duplicates (two instances), peer deps will allow the consuming app/package to decide and will help deduplication by the package manager.

> **Note**
> Not all problems about instances are related to deduplication... next-i18next and react-i18next are prone to dual packaging hazards. I'll plan to provide a fix for that in a later iteration (on both packages). But moving to peer-deps will help as a first step. Cause I feel it's difficult to reason when fixing issues (ie: https://github.com/i18next/next-i18next/issues/1917)

## What's new

- Yarn PNP should work (need to confirm)

## Version

- Major (v13) to clearly indicates the change in install
- No breaking changes 

Makes sense to merge along https://github.com/i18next/next-i18next/pull/1974 and possibly 

## Upgrade

See https://github.com/i18next/next-i18next/pull/1966/files#diff-f411a5009131db5a6c7242b49ca78488f2d274ef9e5e5f4de9262a42108a5386

## Notes

Recent versions of packages managers finally aligned on peer-deps and does not install them by default.

- NPM 8: https://github.com/npm/rfcs/blob/main/implemented/0030-no-install-optional-peer-deps.md

In certain situations (auto-install-peers in pnpm or npm v7), some people might forget to add the peer-deps as documented in v13. That might lead to *invalid* bug reports if the auto-installed version is not within the expectations or created legit duplicates (semver). In those possible cases, ensure both react-i18next and i18next are installed in the consuming app. 

-----------

PS: I couldn't run `npm run test` without doing `npm run install:examples` first.

#### Checklist

- [ ] tests are included


adrai:
Let's wait for the react-i18next update first

belgattitude:
@adrai react-i18next use peerDependencies with a `>=19.0.0` *range*.  

https://github.com/i18next/react-i18next/blob/a0eed42303bda4863a832cbd8b280d04448f2690/package.json#L92

So technically speaking there's nothing on their side. That said I've just forced the ^22.0.0 on this repo to test it out:

https://github.com/belgattitude/nextjs-monorepo-example/pull/2802

So far it passes e2e and previews.

belgattitude:
That said I feel it would make sense to add i18next as a peerDependency at some point:

```json
  "peerDependencies": {
    "i18next": "^21.9.1 || ^22.0.0"
  }
```

+ Update install instructions will need to be updated, ie: `yarn install next-i18next i18next` + maybe major to clarify the change

adrai:
> That said I feel it would make sense to add i18next as a peerDependency at some point:
> 
> 
> 
> ```json
> 
>   "peerDependencies": {
> 
>     "i18next": "^21.9.1 || ^22.0.0"
> 
>   }
> 
> ```
> 
> 
> 
> Plus update install instructions, ie: `yarn install next-i18next i18next` + maybe major to clarify the change
> 
> 
> 
> 

There's an ongoing discussion about this: https://github.com/i18next/next-i18next/pull/1951

adrai:
@bryantobing12 @yuuk @martinjlowm @bojackhorseman0309 @rikkit @kachkaev @capellini @matthewwolfe @isaachinman can you please have a look at this PR?
Based on the discussion here: https://github.com/i18next/next-i18next/issues/1917#issuecomment-1258137642
My main concern is this PR conflicts with the current drop-in wrapper approach and probably will not solve all mentioned issues.

package.json
```diff
@@ -117,10 +117,12 @@
     "eslint-plugin-typescript-sort-keys": "^2.1.0",
     "gh-release": "6.0.4",
     "husky": "^3.0.0",
+    "i18next": "^22.0.1",
     "jest": "^26.6.3",
     "next": "^12.2.5",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
+    "react-i18next": "^11.18.6",
```
adrai, line 125: as soon as this lands, I'll create a new react-i18next major version: https://github.com/i18next/react-i18next/pull/1501

belgattitude: Please wait, I have some things to ask and probably would like to know about the policy regarding dual packaging.

belgattitude: I'll update the issue description during the day

belgattitude:
@adrai I've looked and provided a fix for https://github.com/i18next/next-i18next/issues/1917#issuecomment-1285401978. Tested with yarn and pnpm. With it I'm somehow letting the bundler choose the right package (not totally correct but I can't explain it more). That will circumvent [dual package hazards](https://github.com/i18next/next-i18next/issues/1917#issuecomment-1285401978). But at a later point the best is to rework the packaging/exports of both next-i18next and react-i18next. It might come with a major as well.

I'll take some time tomorrow to think if it's worthy to make both changes together.

belgattitude:
In reply to 
> Based on the discussion here: https://github.com/i18next/next-i18next/issues/1917#issuecomment-1258137642
My main concern is this PR conflicts with the current drop-in wrapper approach 

What we could say for sure

- Only peer-deps will allow ranges (useful for i18next 21 && 22). 
- We loose drop-in approach (install 3 packages - I mean react, emotion... did it as well for reducing the amount of issues)

> and probably will not solve all mentioned issues.

Not cause if you're going into detail, some of them comes from the dual packaging (try to play by removing the cjs or esm export)... But at least we can distinguish problems related to duplicates.

package.json
belgattitude: Great I'll update the range

adrai:
> * We loose drop-in approach

no problem for me, just want to be sure this is generally accepted...
see this discussion here: https://github.com/i18next/next-i18next/issues/136

belgattitude:
> no problem for me, just want to be sure this is generally accepted...

Yes, true. My feeling: not everything is perfect. Some time ago I had the impression that most of the time, it ended up by a package manager discussion. Now they aligned. I've added a note in the description.

rikkit:
Happy to see the peer deps change looks like it's going ahead, I'll close my PR #1951 since this branch has the same changes.

@belgattitude I'd be happy to test this branch using Yarn 3 PnP if you need, let me know.

belgattitude:
@adrai I've rebased to latest changes and added the new range to react-i18next v11+v12

Unfortunately I think there's a little error due to https://github.com/i18next/react-i18next/issues/1562

package.json
```diff
@@ -142,13 +144,13 @@
     "@types/hoist-non-react-statics": "^3.3.1",
     "core-js": "^3",
     "hoist-non-react-statics": "^3.3.2",
-    "i18next": "^21.9.1",
-    "i18next-fs-backend": "^1.1.5",
-    "react-i18next": "^11.18.4"
+    "i18next-fs-backend": "^1.1.5"
   },
   "peerDependencies": {
+    "i18next": "^21.9.1 || ^22.0.1",
```
adrai, line 150: Maybe only support >=22 and >=12?

That way we know next-i18next will not support the older versions anymore...

belgattitude: Yes that's an idea too. 

As a maintainer I would drop 21 and 12 (but keep a ^22.0.1 rather than >=22, in the past there were few bc in minors, mostly typescript related. Easier to reason when dropping minors is possible). 

So I'll push some changes in this direction later on. I keep it in mind.

belgattitude: Lets see also what will be done about https://github.com/i18next/react-i18next/issues/1562#issuecomment-1286164854.

belgattitude:
@adrai First of all thanks for your quick merges. It's super nice.

I'll retake on this tomorrow, but to let you know

- I removed the export of Resources https://github.com/i18next/next-i18next/pull/1966/commits/2f64ff557956cb33ea866ceecd46e19f90e6d0b3. 
- As things have changed between majors, I would be happy to have this https://github.com/i18next/next-i18next/pull/1966/commits/2f64ff557956cb33ea866ceecd46e19f90e6d0b3 on master if you don't see any issue. I will help me on this one

package.json
adrai: fyi: 22.0.2 has just been published with an important type fix

examples/simple/types.d/i18next.d.ts.future
```diff
@@ -0,0 +1,25 @@
+/**
```
belgattitude, line 1: @adrai testing upstream i18next 22 and types changes.

- https://github.com/i18next/react-i18next-gitbook/pull/140/files
- https://www.i18next.com/overview/typescript

I tried to move the `declare module 'react-i18next'` to `declare module 'i18next'`, but actually it does not work (it's ignored). The current  `declare module 'react-i18next'` still works perfect.

I guess if we're going only to v22... we might have to change few things (outside of scope of this P/R)

adrai: maybe @pedrodurek can advice?

package.json
belgattitude: I've updated the PR to latest. And removed the ranges for now. Still things to do

examples/simple/types.d/i18next.d.ts.future
adrai: @rosskevin are you more familiar with next.js?

rosskevin: I don't use i18next and haven't been following specific changes closely.  It's probably best for @pedrodurek to advise on the aspects of the custom types changes here.

belgattitude: Agree. 

@adrai as it seems to work without the change.  I propose to remove this file and finalize this pr. Can be done in a subsequent pr.

Wdyt ?

adrai: You mean using the new i18next and react-i18next version but without the i18next.d.ts file?
If @pedrodurek can confirm this, yes.
I don't like to create another next-i18next major afterwards if this is not ok.

belgattitude: Agree.  just to bring some findings... I think it's safe though. 

Existing types augmentations works and  handled from react-i18next not next-i18next. 

Totally fine to wait for me. Thx

pedrodurek: Hey @belgattitude
You should change the import at the scope of the file, otherwise I won't re-export the other types from `i18next`.
`import 'i18next'` instead of `import 'react-i18next'`

adrai:
@belgattitude fyi: https://github.com/i18next/next-i18next/issues/1991

adrai:
fyi: https://github.com/i18next/next-i18next/issues/1982#issuecomment-1291704469

adrai:
fyi2: there might still be an issue if using an i18next backend to load the resources in combination with TS: https://github.com/i18next/i18next/pull/1853#issuecomment-1291566329

examples/simple/types.d/i18next.d.ts.future
belgattitude: @pedrodurek Thanks, I did the change but still no luck. 

I feel there's some more changes to do. I propose this:

1. Extract this part to another P/R (see https://github.com/i18next/next-i18next/pull/1994), so we can work in isolation. 
2. Make this PR mergable (for now it will work based on react-i18next.d.ts) - Done.

You'll find explanations, issues where we can reproduce https://github.com/i18next/next-i18next/pull/1994.

belgattitude:
@adrai thanks for the links, I won't have time to read them properly. But on my *limited* time, I would propose this to ease the process.

1. Review this PR in order to merge it Otherwise I have to deal with too much context shift.
2. Merge it (so I can make it easier to work on both the i18next types augmentation in there https://github.com/i18next/next-i18next/pull/1994. and the new exports in https://github.com/i18next/next-i18next/pull/1973)


This PR scope is:

- Move to peer
- Update to latest

adrai:
Since there was not much feedback regarding the peer deps, I think we should merge it... (but not yet release it)
Can you update the i18nest-fs-backend dependency before merging?

adrai:
> Since there was not much feedback regarding the peer deps, I think we should merge it... (but not yet release it) Can you update the i18next-fs-backend dependency before merging? This would solve also this issue (#1982), right?

updating your branch with master and updating i18next-fs-backend to v2.0.0 in your branch should already be ok

belgattitude:
Great ! v2.0.0 updated.

examples/simple/types.d/i18next.d.ts.future
pedrodurek: I'll finish addressing some other issues that were reported and will look into it right after

PR #1969, opened 2022-10-20
QA improve checks

belgattitude:
I'd like to rework the (dual) packaging if possible. That is a continuation of https://github.com/i18next/next-i18next/pull/1966#issuecomment-1285449857.

But before of that, I would like to ensure I have a correct baseline. So I introduced few things

- a github action
- a size-limit script (with webpack to follow possible tree-shakability regressions)
- an ecmascript check (to prevent regressions in the build)

The size-limit might obsolete in bundlesize, I might remove later on.

Locally

```bash
npm install
npm run build
npm run check-size
npm run check-dist
```

Want to analyze the differences in bundles (esm, cjs):

```bash
npm run analyze-size
```

Will open the 

image

We can already see that cjs does not tree-shake. Big opportunity cause by default nextjs will use it rather than esm.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,48 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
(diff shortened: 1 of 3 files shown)


belgattitude:
this is a draft, I'll focus on this first:

image

belgattitude:
Ah thanks, you're faster than light :smile:

adrai:
fyi: github actions are failing

PR #1970, opened 2022-10-20
Chore: commit lock files

belgattitude:
I'd like to commit lock files.

## Why ?

Reproducible + will ease ci setup

## Downsides

For maintainers when a PR will be provided that touches deps/dev-deps they need to update the lock.

BTW: IMHO the champion of reproducibility is yarn 3+, if interested I might add support for it. There's nice things that this repo would benefit of (pnpm is not yet there)


adrai:
I would prefer to stay with npm for now.

adrai:
btw: tried without cache => https://github.com/i18next/next-i18next/actions/runs/3290992796/jobs/5424575353

belgattitude:
Won't work yet... I'll fix tomorrow.

But @adrai, have you considered to move to yarn 3+ ? would make it easier in my experience.

adrai:
the setup was yarn based in the past, but decided to stay return to standard npm

adrai:
fyi: looks better: https://github.com/i18next/next-i18next/actions/runs/3291104786/jobs/5424827916

adrai:
fyi: only node v18 seems to have problems with the e2e: https://github.com/i18next/next-i18next/actions/runs/3291170906/jobs/5424980668

belgattitude:
I would not add 18.x it now. The idea for me is to get to the issue of packaging.

I saw you've closed the PR.

IMHO the lock file should be there at some point. Is it something you're not wanting to put ? and why ?

adrai:
In module/library projects I prefer not to commit the lockfile(s)... (I've removed v18 for the ci now)

PR #1972, opened 2022-10-20
chore(size-limit): check also for next-app dist folder

belgattitude:
I know we already have bundlesize, but I'd like to expand the size-checks to the example app with size-limit. I'd like to have fine grain control on side-effects of publish improvements.

Same method used as in https://github.com/i18next/next-i18next/pull/1969

Running 

```bash
npm run check-size
```

Will give

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,9 @@ jobs:
       - name: Build next-i18next
         run: npm run build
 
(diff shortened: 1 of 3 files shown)


belgattitude:
A a later point we can instruct the CI to publish a comment about sizes. But not right now

belgattitude:
@adrai Don't work late :)

PR #1973, opened 2022-10-20
Initial support for es module exports 

belgattitude:
> Do not merge yet. I'd like to add more ci tests. Dual packaging is hard when going into details. 

Rework publishing iteration 1.

- [x] Add es module exports. 
- [x] Keep dual packaging (and packaging hazards that comes along)

 ## Why ?

It paves the way to:

- Avoid duplicates instances of react-i18next in some situations (at least in the long run and esm only)
- Prepare a shift to esm only. Warning: recent nextjs will work, not older ones. Jest is almost there, but there's probably a lot of old installs. So I'll try to keep a conservative approach
- Explicitly declare subpaths

## Version

Minor version. Should not break anything. 

## Risks

Dual packaging is always tricky. There's risks to add the exports field. Out of my mind (not exhaustive): 

- Typescript (< 4.9) still have issues when dealing with types and subpaths exports (like serverSideProps). Fixed by typeConfig
- Older webpack consider the non-standard 'import' field (rather than modules), so I kept it there with comment

## Next iterations

I'll detail soon. But I see a lot of potential for either size-reduction, either normalization.  Other packages i18next, react-i18next will have to be updated too to  bring full potential.

## Size history

Before:

image

After:

No size regression


babel.config.cjs
```diff
@@ -1,4 +1,10 @@
-{
+const browsersList = [
```
belgattitude, line 1: I prefer to not expose yet the browsersList into package.json. In next iteration I'll probably move this.

package.json
```diff
@@ -20,7 +21,49 @@
...
+    "./serverSideTranslations": {
+      "types": "./dist/types/serverSideTranslations.d.ts",
+      "module": "./dist/esm/serverSideTranslations.js",
+      "import": "./dist/esm/serverSideTranslations.js",
+      "require": "./dist/commonjs/serverSideTranslations.js",
+      "default": "./dist/esm/serverSideTranslations.js"
+    },
+    "./package.json": "./package.json"
+  },
+  "typesVersions": {
```
belgattitude, line 57: Credits to @andrewbranch, see https://github.com/andrewbranch/example-subpath-exports-ts-compat#strategy-support-matrix

belgattitude:
@adrai I've tested on few IDE and seems to work fine (at least with a recent typescript version)

It's a first step, I tried to not optimize the bundle yet and keep what did exists. On the user side it does not bring improvements (yet) but there's a slight possibility of issues.

2 options:

- we merge it (so it will be included in v13) and I'll provide size-reduction on a 13.1. That way not everything is sent at the same time.
- we keep here for a while.

adrai:
I would wait for the feedback regarding the peer dep change first...
and then since it's a major we can release all together with v13...

belgattitude:
Redone in https://github.com/i18next/next-i18next/pull/1998

PR #1975, opened 2022-10-20
Refactor simple example to typescript

belgattitude:
While working on esm publication: https://github.com/i18next/next-i18next/pull/1973, I'd like to ensure this works with esmExternals and compute potential size differences.

To do so: 

- I moved the simple example to typescript (cause next move their examples too, more popular nowadays)
- Added a way to inject experimental.esmExternals through env.
- Added a step in CI for it

I took the existing example as base. It think the config is a bit outdated. I'll comment in the file change


examples/simple/pages/index.tsx
```diff
@@ -77,10 +82,10 @@ const Homepage = () => {
   )
 }
 
-// export const getServerSideProps = async ({ locale }) => ({
-export const getStaticProps = async ({ locale }) => ({
+// or getServerSideProps: GetServerSideProps<Props> = async ({ locale })
+export const getStaticProps: GetStaticProps<Props> = async ({ locale }) => ({
   props: {
-    ...await serverSideTranslations(locale, ['common', 'footer']),
+    ...await serverSideTranslations(locale ?? 'en', ['common', 'footer']),
```
adrai, line 88: Is this really necessary? Shouldn't it fallback automatically?

adrai:
image

examples/simple/pages/index.tsx
adrai: If this is not the case, we may move [this check](https://github.com/i18next/next-i18next/blob/master/src/serverSideTranslations.ts#L20) down [here](https://github.com/i18next/next-i18next/blob/master/src/serverSideTranslations.ts#L33) and fallback to `defaultLocale` if undefined.

examples/simple/next.config.js
```diff
@@ -1,5 +1,21 @@
-const { i18n } = require('./next-i18next.config')
+// @ts-check
+const { i18n } = require('./next-i18next.config.js')
 
-module.exports = {
+// You can remove the following 2 lines when integrating our example.
+const { loadCustomBuildParams } = require('./next-utils.config')
```
belgattitude, line 5: @adrai This is purely for be able to have a ci step that test the next version under esmExternals.

Drawback: newcomers might find the example convoluted.

adrai: As soon as that is not experimental anymore, we might remove that again?

examples/simple/pages/index.tsx
belgattitude: locale is always there (except maybe for nextjs prior to when locale was added)

I don't really want to the throw Error in there. I feel it's enough with the ??.

Second for the sake of the *simple* example I don't really want to require/import the next-i18next.config.js to get the fallback. At least not yet, cause it wasn't working in older version without the localePath trick. If I remember well.

That said a more advanced example should be great. One that typechecks the keys and handle properly localePath in all scenarios (getStatic, getServer, app and csr pages + monorepo), but for that the example will look a bit like https://github.com/bryantobing12/layout-reproduce-i18next/pull/1. Which is not simple.

So for now, I'm just focusing on publishing. Doing so I saw a lot more to do in the codebase and those little might not be necessary anymore once I've cleaned'up. (not sure I can but let's hope)

belgattitude:
If there's no blocking comment and CI passes, I feel this should merged. 

On my side it looks good

examples/simple/pages/index.tsx
adrai: I mean in the library code, not in the example code... check the links in my previous comment.

examples/simple/next.config.js
belgattitude: Yes totally. The nice thing is that I've added ts-checks so in case nextjs is updated, the outdated config will be visisble on build (through a warning). Will be easier to maintain.

PS: I could have gone in mjs, but I've reverted cause I feel most people are not yet used to

.github/workflows/ci.yml
```diff
@@ -43,3 +43,7 @@ jobs:
       - name: E2E tests
         run: npm run test:e2e
 
+      - name: Build example with experimental.esmExternals
```
belgattitude, line 46: That is the part that I want to merge into the peer-deps change and new publishing. It will help me cover edge cases

examples/simple/pages/index.tsx
belgattitude: Ah yes. To be sure we're talking about the same

Here nextjs type the locale as string | undefined, and I want it as string. 

Even if we would accept an undefined as first parameter to serverSideTranslations and do the job there, I feel would worsen the signature (I mean having *optional* as first param is not living well, ie `myFunction(undefined, undefined, {param: 1}`)

I don't feel neither than nextjs is at fault here. It's good that the developer handles the fallback. I mean in how next-i18next works right now. And mysefl I tend to create my own function (that will deal with more)

adrai: ok

PR #1985, opened 2022-10-21
fix(example): import config in getServerSideProps (vercel)

belgattitude:
In case the previous one won't work. This is the missing part


belgattitude:
As it works: can be closed

https://next.i18next.com/second-page

PR #1987, opened 2022-10-24
fix: Fix types for appWithTranslation

### LINKED ISSUE #1944: Next.js 12.3.0 AppProps generic breaks appWithTranslation type
Next.js 12.3.0 [#38867](https://github.com/vercel/next.js/pull/38867) introduces the use of a generic for AppProps type, that is incompatible with `appWithTranslation` type.

## To Reproduce

The following example causes a TypeScript error (TS2345):`Argument of type '({ Component, pageProps }: AppProps<{ session: Session;}>) => JSX.Element' is not assignable to parameter of type 'ComponentType<AppProps>'`.

```ts
const App = ({Component, pageProps}: AppProps<{session: Session}>) => (
  <SessionProvider session={pageProps.session}>
    <ThemeProvider>
      <Component {...pageProps} />
    </ThemeProvider>
  </SessionProvider>
)

export default appWithTranslation(App)
```

oddsund:
This PR updates the appWithTranslation generic type to fit what is sent in by the consumers of the library.
It also updates _nextI18Next in SSRConfig to be nullable, which it may be in the case of not calling serverSideTranslations in the page.
The inverse, which is how it is today, would mean that appWithTranslation **always** would get a _nextI18next prop, which is not guaranteed, as indicated by the various null checks and errors thrown.

This fixes #1944 

#### Checklist

- [x] tests are included
- [X] commit message and code follows the [Developer's Certification of Origin](https://github.com/i18next/.github/blob/master/CONTRIBUTING.md)

### DIFF
--- a/src/appWithTranslation.server.test.tsx
+++ b/src/appWithTranslation.server.test.tsx
@@ -8,6 +8,7 @@ import { I18nextProvider } from 'react-i18next'
 import { renderToString } from 'react-dom/server'
 
 import { appWithTranslation } from './appWithTranslation'
+import { AppProps } from 'next/app'
 
 jest.mock('fs', () => ({
   existsSync: jest.fn(),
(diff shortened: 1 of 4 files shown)


adrai:
will be published with the next major -> #1966

PR #1998, opened 2022-10-26
Support es modules

belgattitude:
> Do not merge yet. I'd like to add more ci tests. Dual packaging is hard when going into details. 

Rework publishing iteration 1.

- [x] Add es module exports. 
- [x] Keep dual packaging (and packaging hazards that comes along)
- [x] Decrease size-limits to account for new peer-deps (treeshakes)

 ## Why ?

It paves the way to:

- Avoid duplicates instances of react-i18next in some situations (at least in the long run and esm only)
- Prepare a shift to esm only. Warning: recent nextjs will work, not older ones. Jest is almost there, but there's probably a lot of old installs. So I'll try to keep a conservative approach
- Explicitly declare subpaths

## Version

Minor version. Should not break anything. 

## Risks

Dual packaging is always tricky. There's risks to add the exports field. Out of my mind (not exhaustive): 

- Typescript (< 4.9) still have issues when dealing with types and subpaths exports (like serverSideProps). Fixed by typeConfig
- Older webpack consider the non-standard 'import' field (rather than modules), so I kept it there with comment

## Next iterations

I'll detail soon. But I see a lot of potential for either size-reduction, either normalization.  Other packages i18next, react-i18next will have to be updated too to  bring full potential.

## Size history

Before:

```
Route (pages)                              Size     First Load JS
┌ ● /                                      1.08 kB        97.9 kB
├   /_app                                  0 B            94.6 kB
├ ○ /[40](https://github.com/i18next/next-i18next/actions/runs/3330057611/jobs/5508051889#step:8:41)4                                   181 B          94.8 kB
└ λ /second-page                           799 B          97.7 kB
+ First Load JS shared by all              94.6 kB
  ├ chunks/framework-3b5a00d5d7e8d93b.js   [45](https://github.com/i18next/next-i18next/actions/runs/3330057611/jobs/5508051889#step:8:46).4 kB
  ├ chunks/main-2b5cf5964c9408ce.js        26.5 kB
  ├ chunks/pages/_app-6e90689c67c1aa69.js  22 kB
  └ chunks/webpack-8fa1640cc84ba8fe.js     750 B
```

After:

```
Route (pages)                              Size     First Load JS
┌ ● /                                      1.08 kB        97.9 kB
├   /_app                                  0 B            94.6 kB
├ ○ /[40](https://github.com/i18next/next-i18next/actions/runs/3330057611/jobs/5508051889#step:8:41)4                                   181 B          94.8 kB
└ λ /second-page                           799 B          97.7 kB
[...]

PR #1997, opened 2022-10-26
feat: support i18next v22 type augmentation

belgattitude:
Change example to support the new i18next v22 types augmentation. 

To update look at https://github.com/i18next/next-i18next/pull/1997/files#diff-53f0261e2e3a0e94c11e361fb2729a6d2995921dca99da95aa77209f66226156

PR #2003, opened 2022-11-06
Examples should use git install (do not merge)

belgattitude:
When working on #1998 and #1997, I feel it's useful to be able to test from the git build rather than installing the published version (possibly old)

IMHO, we should move out from npm to yarn 3+ or pnpm 7+ and rework the directory structure. Why ? Npm and workspaces aren't yet there when we have to deal with monorepos (ie: workspace protocol rather than files, versioning, isolation...).

### DIFF
--- a/examples/simple/components/Footer.tsx
+++ b/examples/simple/components/Footer.tsx
@@ -3,21 +3,17 @@ import { useTranslation, Trans } from 'next-i18next'
 import type { FC } from 'react'
 
 export const Footer: FC = () => {
-
   const { t } = useTranslation('footer')
+  const { version = 'git' } = pkg as unknown as { version: string }
 
   return (
     <footer>
-      <p>
-        {t('description')}
-      </p>
-      <p>
-        next-i18next v
-        {pkg.version}
-      </p>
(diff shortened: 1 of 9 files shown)


belgattitude:
I'm investigating this

```
../../dist/commonjs/serverSideTranslations.js
Critical dependency: the request of a dependency is an expression

Import trace for requested module:
../../dist/commonjs/serverSideTranslations.js
../../serverSideTranslations.js
```

and

```
Error occurred prerendering page "/en". Read more: https://nextjs.org/docs/messages/prerender-error
Error: Cannot find module '/home/sebastien/github/next-i18next/examples/simple/next-i18next.config.js'

```

belgattitude:
@adrai  I won't probably put more work on it in its current form. But it was interesting. The reason why I investigated this:

- I'd like to ensure that we can catch earlier bugs in publishing and give more confidence when reworking the repo

belgattitude:
@adrai If you postpone the next release (v13)... There's actually something wrong, I'll get it sorted out. Reproduced in https://github.com/belgattitude/nextjs-monorepo-example/pull/2864 (inline with what cypress gives in this P/R).

belgattitude:
Closed in favour of https://github.com/i18next/next-i18next/pull/2012

PR #2004, opened 2022-11-07
chore(scripts): simplify npm scripts

belgattitude:
Just a simplification of scripts.

PS: Reason, I'm working in another repo to debug why current master does not work. When using npm-run-all, I don't have to think about yarn/npm. The current used will be picked up


adrai:
but this needs npm-run-all to be installed now, right?

adrai:
lgtm

kachkaev:
I’m using `npm-run-all` in my projects too and prefer calling `npm-run-all` command instead of `run-p` and `run-s` ([example](https://github.com/kachkaev/njt/blob/2edb428c8472d4b996eb8675dd8a2c6a30333ec9/package.json#L11-L16)). The latter are a bit confusing because someone may search for `run-p` and `run-s` packages in dependencies.

belgattitude:
Thanks @kachkaev I'll provide a PR to fix that. Make sense

PR #2007, opened 2022-11-07
Ci: iteration 1 to test the repor build rather than published version

belgattitude:
Not confident about the ci. Examples are installing the published version, this defeats them.

The code I've put is temporary !. A better solution will be handled by https://github.com/i18next/next-i18next/pull/2003 but as the repo is a little new to me I'd like to go step by step


belgattitude:
Mmmh. I'm supposed to have a test-e2e failure here. I'll try to see why that isn't happening.

PR #2008, opened 2022-11-07
chore: remove bundlesize in favour of size-limit

belgattitude:
I'd like to remove bundlesize. Why ?

- We introduced size-limit (and npm run check-size in the https://github.com/i18next/next-i18next/pull/1969 and subsequent PR's). I feel it's an easier tool.
- Bundlesize is relying on itorb (node-gyp), this makes install slower as it has to be compiled for no so much benefits (minus 15 seconds in install). 
- I also separate check-size from test-e2e. (I feel it's clearer IMO, but open any changes here)

```
npm why iltorb
iltorb@2.4.5 extraneous
node_modules/iltorb
  iltorb@"^2.4.3" from brotli-size@0.1.0
  node_modules/brotli-size
    brotli-size@"0.1.0" from bundlesize@0.18.1
    node_modules/bundlesize

```

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "example:ssg": "npm run build && cp -r dist examples/ssg/node_modules/next-i18next && cd examples/ssg && npm run dev",
     "example:ssg:prod": "npm run build:example:ssg && cd examples/ssg && npm run start",
     "cypress": "cypress run --config-file cypress/cypress.json",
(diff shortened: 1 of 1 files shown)


belgattitude:
@adrai let me know your opinion on this.

adrai:
> @adrai let me know your opinion on this.

ok for me... you're the expert here ;-)

PR #2012, opened 2022-11-08
E2E test should use the local version

belgattitude:
Before this PR our e2e tests and example install were using the last published version on npm.

This defeats the CI confidence.

As not is perfect, I haven't been able to simply use the npm `link:../..` protocol to create a symlink to the root (where the dist folder resides). The main reason is that the there's issues:

**The first:** 

Code like this:

https://github.com/i18next/next-i18next/blob/0c796cbc6dfb439582d6fe448dfb7b72d15f15d7/src/serverSideTranslations.ts#L26-L29

Will be transpiled by babel (cjs) to this

```js
 return Promise.resolve("".concat(_path["default"].resolve(DEFAULT_CONFIG_PATH))).then(function (s) {
              return _interopRequireWildcard(require(s));
            });
```

And it won't work in a webpack context: require will fail. (I've tested a lot of options, like webpackIgnores...). Changing this is a significant work.

My take on this:

@adrai FYI: We shouldn't play with that and always require the consuming code to provide the config (especially next 13 turbopack, etc). For example appWithTranslation and serverSideTranslations must receive the config file path. Otherwise it won't work (in monorepo situations for example). There's options to do so (like creating a new method for ssr and make the second parameter of appWithTranslation required). 

In order to move forward, I'll let this for a later time.

PS: This P/R introduce more complexity for the examples. But I feel for now it's a good trade-off. I don't think people will use blindly our package.json as is (with the file:../...). The gain for us is quality.


belgattitude:
@adrai before merging I would like your opinion on this. 

TLDR: we make examples a bit dirty, we gain on confidence on CI.

package.json
```diff
@@ -59,6 +59,8 @@
     "example:ssg:prod": "npm run build:example:ssg && cd examples/ssg && npm run start",
     "cypress": "cypress run --config-file cypress/cypress.json",
     "test": "npm-run-all -s check-types clean build build:example:simple && NODE_ENV=test jest --maxWorkers=1 --silent",
+    "move-build-to-examples?": "Hack: see # @link https://github.com/i18next/next-i18next/pull/2012",
+    "move-build-to-examples": "rimraf ./examples/*/node_modules/next-i18next && cpy './package.json' './*.js' './*.ts' './dist' ./examples/simple/node_modules/next-i18next && cpy './package.json' './*.js' './*.ts' './dist' ./examples/ssg/node_modules/next-i18next",
```
belgattitude, line 63: Dirty part :face_exhaling: till better options emerges.

adrai:
> @adrai before merging I would like your opinion on this.
> 
> TLDR: we make examples a bit dirty, we gain confidence on CI.

I think i can live with that.

PR #2013, opened 2022-11-08
refactor: explicit config loading in examples

belgattitude:
A proposal in the light of potential issues regarding how we load the next-i18next.config.js.

1. Currently it's not possible to use the examples "as is" in a monorepo structure, they will lead to errors like

```
Error: Cannot find module '/home/sebastien/github/next-i18next/examples/simple/next-i18next.config.js'
```

See https://github.com/i18next/next-i18next/pull/2012 for reasons why and https://github.com/i18next/next-i18next/issues/1917#issuecomment-1285401978

2. You will need to pass in an i18next instance by using initReactI18next 

```
You will need to pass in an i18next instance by using initReactI18next 
```

See: https://github.com/i18next/next-i18next/issues/1917#issuecomment-1320439358

I feel the examples should cover this by default. 

I would even go further later on and take the opportunity of upcoming release (v13) to include some breaking changes. 

1. **appWithTranslation**: make the second parameter (the config override) mandatory.  It seems okay cause people upgrading will have to do it only on one file (_app.tsx).  
2. **serverSideTranslations**: This is more risky as there will be multiple occurences of it in user routes. So BC more complex to deal with. A suggestion would be to create a new function (ie: `getServerTranslations(config, locale, namespaces?)`). So people affected could progressively adapt (ie: when they move to monorepo...)

In my experience, I used to not rely directly on the serverSideTranslations and wrap it in something similar to this:

```typescript
/**
 * Retrieve translations on server-side, wraps next-i18next.serverSideTranslations
 * to allow further customizations.
 */
import type { SSRConfig, UserConfig } from 'next-i18next';
import { serverSideTranslations } from 'next-i18next/serverSideTranslations';
import nextI18nextConfig from '../../../next-i18next.config';

export const getServerSideTranslations = async (
  locale: string,
  namespacesRequired?: string[] | Readonly<string[]> | undefined,
  configOverride?: UserConfig | null
): Promise<SSRConfig> => {
  const override = configOverride ?? nextI18nextConfig;

  // Slice needed here cause serverSlideTranslations does not accept Readonly type
  return serverSideTranslations(locale, namespacesRequired?.slice(), override);
};
```

https://github.com/belgattitude/nextjs-monorepo-example/blob/3f385aeb3b392671cce91114416ed32312d6ea6c/apps/nextjs-app/src/lib/i18n/getServerSideTranslations.ts#L6-L18

Don't know yet if it would make sense to:

- either document/recommend to wrap like the last example
- either add the 2 BC (or paths) suggested above

I'll take some time to think about this. Not sure myself


adrai:
At the beginning the simple example was really there to have the minimum amount of code to use next-i18next...
Alternatively, a new example with more advanced setup could be created?

belgattitude:
Make sense.

belgattitude:
I'll plan to rework the examples. Moving to a monorepo would make sense. I'll come with a proposal soon

In the meantime I've added a section to help users facing the issue

- https://github.com/i18next/next-i18next/pull/2018/files#diff-2a0bc40e955e3ada6192d2991fb316c2e356e510e2cdab1f6f3a14db0e41bf2b

belgattitude:
@adrai I feel the requirement of passing `next-i18n.config.js` will have to be done at some point.

Without passing the config I fear we actually makes it complex. 

Example

- https://github.com/i18next/next-i18next/issues/1917#issuecomment-1320439358
(and many others)

I'll keep this PR open for investigation after v13.0.0 and update the TROUBLESHOOT once I'm sure

PR #2018, opened 2022-11-10
docs: prepare v13

belgattitude:
Prepare v13, last checks and release notes

- [x] Node 18 (in #2017 )
- [x] Test a canary on (wip in https://github.com/belgattitude/nextjs-monorepo-example/pull/2864) - All good


belgattitude:
Might be blocked in https://github.com/belgattitude/nextjs-monorepo-example/pull/2864

image

Will look this week-end or Monday

TROUBLESHOOT.md
```diff
@@ -1,5 +1,50 @@
 ## Troubleshoot
 
+### Can't find next-i18next.config.js
```
belgattitude, line 3: @adrai that's a proposal 

The goal is to help users:

- monorepos and pnpm users (pnpm creates symlinks)

I'm not totally for this solution. I would prefer to make the config required, see also https://github.com/i18next/next-i18next/pull/2013. But that would be another BC.

adrai:
@pedrodurek what is your opinion?

TROUBLESHOOT.md
belgattitude: See also https://github.com/i18next/next-i18next/issues/1917#issuecomment-1285401978

belgattitude:
I'll merge as it's only doc and some peer-deps update.

But interested about @pedrodurek opinion as well. I'll see if I can advance on other topics.

Thanks a lot @adrai

PR #2020, opened 2022-11-14
chore(examples): lock nextjs version

belgattitude:
Due to a regression in swcMinify in nextjs 13.0.3, our (e2e) size limit checks fail. See https://github.com/vercel/next.js/issues/42813

The issue have since been fixed but not released.

As we don't use lock for examples, it's difficult to avoid having that issue to pop up from time to time (making the ci fail for no *real* reason). I locked the versions of nextjs (till we have better solution). good tradeoff

### DIFF
--- a/examples/simple/package.json
+++ b/examples/simple/package.json
@@ -13,7 +13,7 @@
   },
   "dependencies": {
     "i18next": "^22.0.4",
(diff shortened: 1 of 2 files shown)


belgattitude:
I'll keep on my list a better solution. Needs a little more time

PR #2057, opened 2022-12-12
fix: align with nextjs i18n config types

belgattitude:
Add missging properties to to i18n user config. That aligns with latest nextjs types.

Will fix: https://github.com/i18next/next-i18next/issues/2049

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -15,9 +15,20 @@ import {
 } from 'i18next'
 import { appWithTranslation, i18n } from './'
 
+/**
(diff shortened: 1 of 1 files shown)


belgattitude:
@adrai IMHO safe to merge (and I'll do) but regarding release 13.0.1 ....  I feel it's good to focus on this https://github.com/i18next/i18next/issues/1881 because we'll have bug reports about it.

My time is very short, but I'll try to check in a couple of days. Have a great day

adrai:
lgtm if this fixes the issue

belgattitude:
I hope so :smile:

PR #2058, opened 2022-12-12
Examples updated to latest i18next

belgattitude:
A bug regarding key typings haven introduced in i18next v22.4.0. It's now fixed in latest `^22.4.3` (upstream dep)

- See https://github.com/i18next/i18next/issues/1859#issuecomment-1346177482
- See https://github.com/i18next/i18next/issues/1881

If you are experiencing errors such as `TS2746: This JSX tag's 'children' prop expects a single child of type 'ReactI18NextChild | Iterable ', but multiple children were provided.`...

..simply ensure i18next is at version `^22.4.3` in your package.json and run install.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,13 @@
 ### Fix
 
 - Fix missing `i18n.localeDetection` in UserConfig, see [#2057](https://github.com/i18next/next-i18next/pull/2057).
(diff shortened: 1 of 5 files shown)


adrai:
v13.0.1 is published...

btw: the ssg example seems not to build anymore... but I did not check why... here it works: https://github.com/i18next/next-language-detector/tree/main/examples/basic

PR #2084, opened 2023-02-02
feat(server-side): custom default config path

intpp:
To solve issues like this one: https://github.com/nrwl/nx/discussions/4983, and to not create hacky functions I suggest the following change.

It doesn't break anything but it adds the ability to set the default config path.

It will allow you to simply use the package in NX monorepo with the changes from the `dynamic path` example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -329,6 +329,34 @@ For example, if you want to use `{` and `}` the config would look like this:
 }
 ```
 
+#### Custom `next-i18next.config.js` path
+
+If you want to change the default config path, you can set the environment variable `I18NEXT_DEFAULT_CONFIG_PATH`.
+
+For example, inside the `.env` file you can set a static path:
+```
+I18NEXT_DEFAULT_CONFIG_PATH=/path/to/project/apps/my-app/next-i18next.config.js
+```
+
+Or you can use a trick for dynamic path and set the following inside `next.config.js`:
+
+```js
+process.env.I18NEXT_DEFAULT_CONFIG_PATH = `${__dirname}/next-i18next.config.js`;
+
+// ... Some other imports
+
+const { i18n } = require('./next-i18next.config');
+
+// ... Some other code
+
+module.exports = {
+  i18n,
+  ...
+};
+```
+
+This means that the i18n configuration file will be in the same directory as `next.config.js` and it doesn't matter where your current working directory is. This helps for example for `nx` when you have monorepo and start your application from project root but the application is in `apps/{appName}`.
+
 ## Notes
 
 ### Vercel and Netlify
--- a/src/serverSideTranslations.ts
+++ b/src/serverSideTranslations.ts
@@ -9,7 +9,7 @@ import { globalI18n } from './appWithTranslation'
 import { UserConfig, SSRConfig } from './types'
 import { getFallbackForLng, unique } from './utils'
 
-const DEFAULT_CONFIG_PATH = './next-i18next.config.js'
+const { I18NEXT_DEFAULT_CONFIG_PATH: DEFAULT_CONFIG_PATH = './next-i18next.config.js' } = process.env
 
 export const serverSideTranslations = async (
   initialLocale: string,


adrai:
cool...

adrai:
included in v13.1.0

quanglam2807:
This change broke production on Vercel but I don't know why:

```
2023-02-02T19:25:02.237Z	2c59565a-3cdb-45fb-a5ce-772f5d58ebcb	ERROR	Error: next-i18next was unable to find a user config
    at _callee$ (/var/task/node_modules/next-i18next/dist/commonjs/serverSideTranslations.js:93:19)
    at tryCatch (/var/task/node_modules/@babel/runtime/helpers/regeneratorRuntime.js:44:17)
    at Generator.<anonymous> (/var/task/node_modules/@babel/runtime/helpers/regeneratorRuntime.js:125:22)
    at Generator.next (/var/task/node_modules/@babel/runtime/helpers/regeneratorRuntime.js:69:21)
    at asyncGeneratorStep (/var/task/node_modules/@babel/runtime/helpers/asyncToGenerator.js:3:24)
    at _next (/var/task/node_modules/@babel/runtime/helpers/asyncToGenerator.js:22:9)
    at /var/task/node_modules/@babel/runtime/helpers/asyncToGenerator.js:27:7
    at new Promise (<anonymous>)
    at /var/task/node_modules/@babel/runtime/helpers/asyncToGenerator.js:19:12
    at serverSideTranslations (/var/task/node_modules/next-i18next/dist/commonjs/serverSideTranslations.js:157:17) {
  page: '/apps/popular/page/[page]'
}
RequestId: 2c59565a-3cdb-45fb-a5ce-772f5d58ebcb Error: Runtime exited with error: exit status 1
Runtime.ExitError

```

adrai:
just tested if our simple example deploys on Vercel... and seems to work without any issues: https://github.com/i18next/next-i18next/tree/master/examples/simple

quanglam2807:
Thanks, @adrai. I did more testing. It seems like it only affects dynamic routes. You can re-produce the issue with this repo: https://github.com/quanglam2807/next-i18next-simple-test/blob/main/pages/pages/%5Bpage%5D.tsx

adrai:
@quanglam2807 this is really strange... to me this looks like to be a Vercel issue... do you want to try to open an issue/discussion here? https://github.com/vercel/community/discussions

adrai:
@quanglam2807 It seems like, as soon as the serverSideTranlsations file accesses process.env the error occurs: https://github.com/i18next/next-i18next/commit/c6ccf412043e7778509a53d1386388de7eded6e9#diff-78da8fd247ac10bc4179e3dad343177bb6bf49923dcc4bd6cab86babfae025bfL27
I've now "reverted" the changes of this PR and released v13.1.4 (which basically is like v13.0.3)

@intpp it would be nice if you could investigate why this happens with the Vercel team/support

quanglam2807:
Thank you, @adrai. I also suspected the same thing but oddly enough, calling `process.env.I18NEXT_DEFAULT_CONFIG_PATH` directly in `getStaticProps` worked.

intpp:
I'll check

intpp:
@adrai Hi. I did some research on the problem. It causes by the library https://github.com/vercel/nft, which is used by the Next.js' `build` command.

https://nextjs.org/docs/advanced-features/output-file-tracing
```
During next build, Next.js will use https://github.com/vercel/nft to statically analyze import, require, and fs usage to determine all files that a page might load.
```

When we use `process.env` in the variables as a value (not an expression), it can't calculate the possible path for the default value.

It works if we set `I18NEXT_DEFAULT_CONFIG_PATH` later. 

```typescript
let DEFAULT_CONFIG_PATH = './next-i18next.config.js'

// Added
if (process.env.I18NEXT_DEFAULT_CONFIG_PATH) {
  DEFAULT_CONFIG_PATH = process.env.I18NEXT_DEFAULT_CONFIG_PATH
}

export const serverSideTranslations = async (
  initialLocale: string,
  namespacesRequired: string[] | undefined = undefined,
  configOverride: UserConfig | null = null,
  extraLocales: string[] | false = false
): Promise<SSRConfig> => {
  if (typeof initialLocale !== 'string') {
    throw new Error(
      'Initial locale argument was not passed into serverSideTranslations'
    )
  }

  let userConfig = configOverride

  const configPath = path.resolve(DEFAULT_CONFIG_PATH)

  if (
    !userConfig &&
    fs.existsSync(configPath)
  ) {
    userConfig = await import(configPath)
  }

  if (userConfig === null) {
    throw new Error(`next-i18next was unable to find a user config at ${configPath}`)
  }
```

It works for a single application and a monorepo if `next-i18next.config.js` is in the same directory as `next.config.js`. But you must copy it manually to your build directory when it doesn't.

I have published the test version https://www.npmjs.com/package/intpp-next-i18next?activeTab=explore with the change and tested it on Vercel - It works (https://github.com/intpp/i18next-demo).

I'll create a new PR with the change.

PR #2089, opened 2023-02-07
fix(server-side): use environment `I18NEXT_DEFAULT_CONFIG_PATH` to ch…

intpp:
Changes according to https://github.com/i18next/next-i18next/pull/2084#issuecomment-1420511358.


adrai:
thank you...
included in v13.1.5

PR #2097, opened 2023-02-24
Update serverSideTranslation args type

mayone-du:
Thanks for developing a great library!

The `useTranslation` function completes the namespace defined in `CustomTypeOptions` on `i18next.d.ts`, but the `serverSideTranslations` function does not, so I changed the type.

(If the change causes any inconvenience, close this PR.)

#### Checklist

- [x] tests are included

### DIFF
--- a/src/serverSideTranslations.ts
+++ b/src/serverSideTranslations.ts
@@ -8,6 +8,7 @@ import { globalI18n } from './appWithTranslation'
 
 import { UserConfig, SSRConfig } from './types'
 import { getFallbackForLng, unique } from './utils'
+import { Namespace } from 'i18next'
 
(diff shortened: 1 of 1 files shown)


adrai:
Why is that change necessary?
Is there an issue?

mayone-du:
The reason is that the argument passed when executing `getServerSideTranslations` is not type safe.
This is not in the issue, but I created a PR because I was unhappy with it while using it.

adrai:
I'm not a TS expert, but I don't think `Namespace[number]` is correct, @pedrodurek @belgattitude?

mayone-du:
Since the original Namespace type definition is T | T[], Namespace[number] narrows it down to T.

(T contains namespace if it is defined, or a string if not.)

adrai:
included in v13.2.0

PR #2104, opened 2023-03-02
fix serverSideTranslations args

mayone-du:
#2097 change was incorrect. Sorry about that.

At this time (v13.2.0), we ended up with the same type information as before we merged #2097, so we fixed that.

#### Checklist

- [x] tests are included


adrai:
v13.2.1

Issue #2095, opened 2023-02-22
Switching between languages breaks and triggers infinite network calls when using useSuspense: true, and client-side-only translations


danielmarcano:
## 🐛 Bug Report

Switching between languages by using Next.js built-in internationalized routing is breaking and triggering infinite Network calls if:

- We set `{ react: { useSuspense: true } }` inside the `next-i18next.config.js` file's configuration object
- We load our translations on the client-side by following [this guide](https://github.com/i18next/i18next-http-backend/tree/master/example/next#client-side-loading-of-translations-via-http), and set: `backendOptions: [{ expirationTime: 0 }]` (in development, to prevent caching translations if we are manipulating them while developing the app) inside the `next-i18next.config.js` file

## To Reproduce

A [codesandbox](https://codesandbox.io/p/github/danielmarcano/next-i18next-use-suspense-error/main?workspaceId=9297ae8a-8175-4748-8b30-6d300126bb95&file=%2Fsrc%2Fpages%2Findex.tsx&selection=%5B%7B%22endColumn%22%3A1%2C%22endLineNumber%22%3A2%2C%22startColumn%22%3A1%2C%22startLineNumber%22%3A1%7D%5D) that reproduces the error

In short, we can reproduce this bug by:

- Adding client-side-only translations by following [this guide](https://github.com/i18next/i18next-http-backend/tree/master/example/next#client-side-loading-of-translations-via-http) and adding all necessary packages
- Adding `{ react: { useSuspense: true } }` and `backendOptions: [{ expirationTime: 0 }]` to the `next-i18next.config.js` object

```js
// next-i18next.config.js

const HttpBackend = require("i18next-http-backend/cjs");
const ChainedBackend = require("i18next-chained-backend").default;
const LocalStorageBackend = require("i18next-localstorage-backend").default;

const ONE_HOUR_IN_MILISECONDS = 60 * 60 * 1000;

module.exports = {
  backend: {
  backendOptions: [{ expirationTime: process.env.NODE_ENV === 'production' ? ONE_HOUR_IN_MILISECONDS : 0 }],
    backends:
      typeof window !== "undefined" ? [LocalStorageBackend, HttpBackend] : [],
  },
  serializeConfig: false,
  use: typeof window !== "undefined" ? [ChainedBackend] : [],
  i18n: {
    locales: ["en", "es"],
    defaultLocale: "en",
  },
  defaultNs: "common",
  react: {
    useSuspense: true,
  },
};

```

[...]

PR #2226, opened 2023-10-31
refactor: reuse existing instance

wjaykim:
This PR changes how next-i18next loads resources on page or locale change.

Instead of creating a new instance on page or locale change, the change uses [`addResourceBundle`](https://www.i18next.com/overview/api#addresourcebundle) method to load newly changed resources if instance was already created.

I made this change because I found an edge-case:
1. Enter page with `getStaticProps` and `serverSideTranslations` configured.
2. Move to page without `serverSideTranslations`, instead load translations in client-side with i18next-http-backend;
Also I used `suspense` option to `true`
3. Component with `useTranslation` hook got suspended, and it suspends forever because i18next instance is created repeatedly, thus not initialized forever.

After applying this patch, the infinite suspending issue was removed.

#### Checklist

- [x] tests are included

### DIFF
--- a/src/appWithTranslation.client.test.tsx
+++ b/src/appWithTranslation.client.test.tsx
@@ -277,18 +277,18 @@ describe('appWithTranslation', () => {
     expect(args[0].i18n.language).toBe('de')
   })
 
-  it('does not re-call createClient on re-renders unless locale or props have changed', () => {
+  it('does not re-call createClient on re-renders', () => {
     const { rerender } = renderComponent()
     expect(createClient).toHaveBeenCalledTimes(1)
     rerender(<DummyApp {...defaultRenderProps} />)
     expect(createClient).toHaveBeenCalledTimes(1)
     const newProps = createProps()
     rerender(<DummyApp {...newProps} />)
-    expect(createClient).toHaveBeenCalledTimes(2)
+    expect(createClient).toHaveBeenCalledTimes(1)
     newProps.pageProps._nextI18Next.initialLocale = 'de'
     newProps.router.locale = 'de'
     rerender(<DummyApp {...newProps} />)
-    expect(createClient).toHaveBeenCalledTimes(3)
+    expect(createClient).toHaveBeenCalledTimes(1)
   })
 
   it('assures locale key is set to the right value', () => {
--- a/src/appWithTranslation.tsx
+++ b/src/appWithTranslation.tsx
@@ -67,6 +67,7 @@ export const appWithTranslation = <Props extends NextJsAppProps>(
 
       let instance = instanceRef.current
       if (instance) {
+        instance.changeLanguage(locale)
         if (resources) {
           for (const locale of Object.keys(resources)) {
             for (const ns of Object.keys(resources[locale])) {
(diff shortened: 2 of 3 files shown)


adrai:
Can you please provide a minimal example repo I can use to verify the issue first?

wjaykim:
While creating reproduction, I found that when using suspense in layout component(located in _app, higher than page level), the issue is always reproduced.

Please see https://github.com/wjaykim/next-i18next-infinite-suspense for reproduction, and also I added `patch` branch which you can see patched version of `next-i18next` works well compared to `main` branch.

Also I think maybe #2095 is caused by this.

wjaykim:
I accidentally closed PR.

adrai:
The question I have is why do you want to prevent it from being rendered on the server?
As soon as you don't use your `SSRSafeSuspense` component it works without problems...

wjaykim:
> The question I have is why do you want to prevent it from being rendered on the server?
I don't want to add `getStaticProps` each page to render header.

> As soon as you don't use your SSRSafeSuspense component it works without problems...
What do you mean? It doesn't works, actually..
The reason I wrapped with SSRSafeSuspense is without it, nextjs will cause hydration mismatch related error.

adrai:
But why do you set useSuspense to true?
next-i18next was designed in a way that no suspense should be used.

adrai:
btw: using your "fixed" version in this example you can see the language change does not work properly anymore: https://github.com/i18next/next-i18next/tree/master/examples/simple

wjaykim:
> next-i18next was designed in a way that no suspense should be used.

Didn't know that at first, but if this change can make next-i18next be compatible with Suspense, why we shouldn't apply this fix?

> btw: using your "fixed" version in this example you can see the language change does not work properly anymore

I didn't know that, so updated to fix the issue.

adrai:
if that will not hurt the existing behaviour for the existing users, we can merge it (maybe also behind a new option), but right now it seems it still gets issues with the "normal" usage of next-i18next:
Run the examples, change language, navigate to other pages and check the console logs:
image

wjaykim:
> if that will not hurt the existing behaviour for the existing users

Yes that's right. I want you to check if my change doesn't change any existing behavior, and I'm willing to make this PR perfect so it can be merged because suspense is really fascinating feature to me.

To find out this PR has no problem, what's the best way to prove it? Running official examples of this repo?

wjaykim:
It became quite a change to avoid state update in render phase.

If instance was created previously, useMemo will return additional data containing `locale` and `resources`.
Then `changeLanguage` and `addResourceBundle` method will be called to update locale and resources in `useLayoutEffect`.

I also added comments in the code.

Please take a look! 🙃

wjaykim:
Right.. I'll reach you after find more good way and when all tests passed (including e2e)

wjaykim:
Latest code passes all tests including e2e in local. Can you take a look?

adrai:
looks good

adrai:
I hope this does not influence some other edge case usage....
I suggest, we merge that and create a new major version.
If there is any issue from the community that we can't address, we may add an option to be able to switch back to the old instance creation.
What do you think?

wjaykim:
Yes, if we concern more about stability, we can make the code works same as before for default, and add option to reuse instance optionally.

adrai:
but only if issues arise

adrai:
Thank you very much for your contribution.
v15.0.0 was just released.

PR #2270, opened 2024-04-10
Only overwrite `ns` config if it provided

NamPNQ:
Now we require including `serverSideTranslation` to every page that needed to translate

With this change, we can ignore that behavior by using all namespace available

#### Checklist

- [x] tests are included

#### Checklist (for documentation change)

- [ ] only relevant documentation part is changed (make a diff before you submit the PR)
- [ ] motivation/reason is provided
- [ ] commit message and code follows the [Developer's Certification of Origin](https://github.com/i18next/.github/blob/master/CONTRIBUTING.md)


adrai:
The tests are failing...
and I still don't get why `...(ns && { ns })` should change anything...
btw: I get this error with your example:
image

NamPNQ:
@adrai 
To enhance comprehension and clarity, consider the following rephrased version:

Replacing `{ns}` with `{...(ns&&{ns}}` was necessary due to server-side behavior, as evident in the code at https://github.com/i18next/next-i18next/blob/master/src/config/createConfig.ts#L254. When `ns` isn't provided, the server loads all namespaces. However, for auto static, `ns` isn't defined in the configuration, resulting in it being set to undefined. Consequently, no translations occur.

Currently, I'm not having any a solution for handling this mismatch. Since rendering occurs on the server side, while the client side lacks namespace information for preloading, translations won't occur, leading to a mismatch.

For test fail, because I'm missing add build:example for that

adrai:
The mismatch is exactly because setversideTranslations is not called... so this will never work.

The mismatch needs to be resolved else I can't merge this PR.

NamPNQ:
I update to use config `ns` and `initImmediate: false` to fix mismatch, the mismatch only happens in cache first time (the translation not fetched) or the cache expired

src/config/defaultConfig.ts
```diff
@@ -13,7 +13,7 @@ export const defaultConfig = {
     locales: LOCALES,
   },
   get initImmediate(): boolean {
-    return process.browser && typeof window !== 'undefined'
```
adrai, line 16: why is this necessary?

NamPNQ: process.browser is deprecated, and return undefined instead of boolean, so the result of this function is always undefined

adrai:
it's included in v15.3.0

PR #2290, opened 2024-07-29
chore(examples): add test-vitest example

ImBIOS:
#### Checklist

- [x] tests are included


adrai:
What is your motivation for having this inside this repository?
I need the examples that are here in the examples directory to smoke test next-i18next... So I would prefer not to have more examples here.

PR #2310, opened 2024-12-16
Move @types/hoist-non-react-statics to devDependencies"

Erik-Gro:
This pull request moves the "@types/hoist-non-react-statics" package from the dependencies section to the devDependencies section in the package.json file. This change is appropriate because type declaration files are typically used during development and compilation but are not required at runtime.

The modification ensures that the type definitions for "hoist-non-react-statics" are available during development and TypeScript compilation, but won't be included in the production build of projects using this package. This change helps to reduce the size of the production bundle and follows best practices for managing dependencies in TypeScript projects.
To implement this change, I used the following npm command:

npm install @types/hoist-non-react-statics --save-dev

This command installs the package as a dev dependency and automatically updates the package.json file, moving it from the "dependencies" section to the "devDependencies" section.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,7 +24,6 @@
       "license": "MIT",
       "dependencies": {
         "@babel/runtime": "^7.23.2",
-        "@types/hoist-non-react-statics": "^3.3.4",
         "core-js": "^3",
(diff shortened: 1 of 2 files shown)


adrai:
thank you, it's included in v15.4.1

PR #2316, opened 2025-02-06
types: re-add `@types/hoist-non-react-statics` to dependencies

### LINKED ISSUE #2315: new appWithTranslation's return type not backwards compatible
## 🐛 Bug Report

In my NextJs(pages) app I currently have "i18next": "23.14.0", "next-i18next": "15.3.1", "react-i18next": "15.0.1", and everything works fine.
Doing deps update of i18next to "i18next": "^24.2.2", "next-i18next": "^15.4.1", "react-i18next": "^15.4.0", and getting a type error in my _app.tsx:
Image
After spending some time and looking at the code changes in next-i18next I didn't see any obvious changes that would lead to this.
By rolling back the problem went away.

## To Reproduce

A [minimal reproducible example](https://github.com/MadaShindeInai/i18issuerepro).
1. clone repo and write `pnpm i`
2. open pages/_app.tsx and see and error

## Expected behavior

Ensure that the return type of appWithTranslation is backwards compatible or provide instructions on how to migrate correctly to mitigate the TypeScript inconsistency.

## Your Environment

- *runtime version*: node v22.13.0
- *i18next version*: "i18next": "^24.2.2", "next-i18next": "^15.4.1", "react-i18next": "^15.4.0"
- *os*: Mac
- *any other relevant information* "next": "^15.0.1"(same with latest), "react": "^18.3.1" (same with 19+), "@trpc/*": "^11.0.0-rc.730" (same with v10 stable), "@types/eslint": "^8.56.10", (same with 9+), "typescript": "^5.5.3"(same with 5.7.3)

marcalexiei:
Fixes #2315

Reverts #2310

---

From #2310:

> type declaration files are typically used during development and compilation but are not required at runtime.

This is generally true, however `next-i18next` actually relies on types from `hoist-non-react-statics` which are exposed to the user:

https://github.com/i18next/next-i18next/blob/fe1b79a596f269bca9410d00e340e2dcfb633db0/src/appWithTranslation.tsx#L2

https://github.com/i18next/next-i18next/blob/fe1b79a596f269bca9410d00e340e2dcfb633db0/src/appWithTranslation.tsx#L129

If types are not included when installing the library, users will experience types errors, 
that might be hidden if `skipLibCheck` typescript compiler option is set to `true` (like in #2315).

---

If you don't want that `@types` package is always installed you should use `peerDependencies` alongside [`peerDependenciesMeta`](https://docs.npmjs.com/cli/v10/configuring-npm/package-json/?v=true#peerdependenciesmeta).
However, this will require user to manually install `@types` package,
so a mention to this should be added in all relevant places.

#### Checklist

- [ ] tests are included

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,6 +24,7 @@
       "license": "MIT",
       "dependencies": {
         "@babel/runtime": "^7.23.2",
+        "@types/hoist-non-react-statics": "^3.3.6",
         "core-js": "^3",
         "hoist-non-react-statics": "^3.3.2",
         "i18next-fs-backend": "^2.6.0"
@@ -40,7 +41,6 @@
         "@size-limit/webpack": "^10.0.2",
         "@size-limit/webpack-why": "^10.0.2",
         "@testing-library/react": "^14.0.0",
-        "@types/hoist-non-react-statics": "^3.3.6",
         "@types/jest": "^29.5.7",
         "@types/node": "^20.8.10",
         "@types/react": "^18.2.33",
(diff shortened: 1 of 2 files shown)


adrai:
@Erik-Gro fyi

adrai:
thank you, it's included in v15.4.2

PR #2343, opened 2026-06-20
chore: use public repository metadata

jiapengwang-code:
## Summary
- replace the SSH-only repository URL with a public HTTPS Git URL
- keep npm repository metadata usable without GitHub SSH credentials

## Verification
- ESLint (`eslint src`)
- TypeScript checks for both Pages Router and App Router configurations
- Jest: 169 tests passed; 18 existing Windows path-separator assertions failed because they expect `/` instead of `\`
- `pnpm pack --dry-run --json`
- direct package metadata assertion
- `git diff --check`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "next-i18next",
   "version": "16.0.7",
-  "repository": "git@github.com:i18next/next-i18next.git",
+  "repository": "git+https://github.com/i18next/next-i18next.git",
   "author": "i18next",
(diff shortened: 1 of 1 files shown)


adrai:
Thanks, but we've applied this directly in our own commit to match the `repository` object form used across the rest of the i18next org (i18next, react-i18next). Closing this one. Note: next-i18next doesn't participate in third-party bounty programs, so please don't file bounty-driven PRs against it.

